PR #1, opened 2023-09-18
Update Cargo.toml remove www

szabgab:
The web site redirects anyway, so better to link to the final destiniation


HaoboGu:
Thanks for the contribution!

szabgab:
You are welcome. I am building the [rust digger](https://rust-digger.code-maven.com/) in order to catch such small (and some bigger) things that can be improved.

PR #10, opened 2024-03-15
Improve readability

lonesometraveler:
This pull request proposes enhancements to the `register_keycode` and `unregister_keycode` functions to improve the readability and clarity.

By using the `iter().position()`, we can directly express the intent of finding an empty slot or a previously registered key, resulting in clearer and more concise code.


HaoboGu:
thanks!

PR #13, opened 2024-03-17
Remove redundant assignment

lonesometraveler:
This pull request removes a redundant assignment of `key_state.pressed` to `true` in the `process_key_action_tap` function. The assignment is unnecessary as the condition already checks if the key state is pressed.


HaoboGu:
thanks :D

PR #14, opened 2024-03-17
Refactor `scan` and `debounce` functions

lonesometraveler:
This pull request refactors the `scan` and `debounce` functions for improved clarity and separation of concerns.

In the original implementation, the `debounce` function handled both debounce logic and updating key states. This mixing of responsibilities could potentially complicate code comprehension and maintenance.

In the refactored code, the `debounce` module exclusively focuses on determining if a state change has occurred, without directly modifying key states. Subsequently, the `matrix` module takes action based on the returned value, thereby clarifying the respective responsibilities of each component and enhancing code understandability.

Additionally, the `debounce` function is renamed to `has_changed` to more accurately reflect its purpose of detecting changes in state, such as the passing of a debounce period.

By passing immutable references instead of mutable ones, the refactor ensures that the debounce logic can operate on the key state without needing mutable access. This improves code readability and reduces the risk of unintended side effects.

### DIFF
--- a/rmk/src/debounce.rs
+++ b/rmk/src/debounce.rs
@@ -48,7 +48,7 @@ impl<const INPUT_PIN_NUM: usize, const OUTPUT_PIN_NUM: usize>
     }
 
     /// Per-key debounce, same with zmk's debounce algorithm
-    pub(crate) fn has_changed(
+    pub(crate) fn detect_change_with_debounce(
         &mut self,
         in_idx: usize,
         out_idx: usize,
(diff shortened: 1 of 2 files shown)


rmk/src/debounce.rs
```diff
@@ -48,37 +48,34 @@ impl<const INPUT_PIN_NUM: usize, const OUTPUT_PIN_NUM: usize>
     }
 
     /// Per-key debounce, same with zmk's debounce algorithm
-    pub(crate) fn debounce(
+    pub(crate) fn has_changed(
```
HaoboGu: `self.debouncer.has_changed` is a little bit confusing to me, it feels like this function would return whether the *debouncer's state* has changed, not the *key state*. Changing it to `self.debouncer.detect_change_with_debounce` or something similar would be better imo.

lonesometraveler: `detect_change_with_debounce` then.

HaoboGu:
thank you very much!

PR #18, opened 2024-03-18
Refactor debouncer module

lonesometraveler:
This pull request refactors the `debounce` module.

### Changes:

- Adjusted the visibility of the counter struct.
- Renamed `DebounceState` to `DebounceCounter` to reflect its purpose better.

The rationale behind the name change is to introduce a new enum like this next:

```rust
pub(crate) enum DebounceState {
    Debounced,
    InProgress,
    HeldHigh,
    HeldLow,
    Ignored,
}
```

The `detect_change_with_debounce` function can return this enum instead of a bool. This adjustment facilitates the reinsertion of held-down keys.

### DIFF
--- a/rmk/src/debounce.rs
+++ b/rmk/src/debounce.rs
@@ -28,15 +28,6 @@ impl DebounceCounter {
     }
 }
 
-/// Debounce state
-pub(crate) enum DebounceState {
-    Debounced,
-    InProgress,
-    HeldHigh,
-    HeldLow,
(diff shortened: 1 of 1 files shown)


HaoboGu (approved):
thanks!

HaoboGu:
Maybe an enum like "KeyState" would be needed. Both reinsertion and tap/hold feature require more accurate monitoring of key states.

PR #34, opened 2024-04-11
Use `bitfield-struct` to replace`packed_struct`, making it possible to compile for esp32s3

HaoboGu:
Use `bitfield-struct` to replace`packed_struct`, making it possible to compile for esp32s3


rmk/src/light.rs
```diff
@@ -178,17 +180,19 @@ impl<P: OutputPin> LightService<P> {
...
-                        debug!("Read keyboard state: {:?}", indicator);
-                        // Ignore the error, which is `Infallible` in most cases
-                        self.set_leds(indicator).map_err(|_| ())
-                    }
-                    Err(_) => {
-                        error!("packing error: {:b}", self.led_indicator_data[0]);
-                        Err(())
-                    }
-                }
+                let indicator = LedIndicator::from_bits(self.led_indicator_data[0]);
```
HaoboGu, line 183: Remove those code

PR #58, opened 2024-08-24
Update installation method of probe-rs

anderso:
The cargo command no longer works as of 0.24: https://github.com/probe-rs/probe-rs/releases/tag/v0.24.0 This changes it to the installation method featured on the probe.rs site.


HaoboGu:
thanks!

PR #60, opened 2024-08-31
BLE split keyboard

HaoboGu:


yo-boy:
Hello, first I want to say thanks for writing this great project, I'm very interested in seeing it grow

I just wanted to comment here about something, for wireless split keyboards being able to use either side as the master is fairly important, because the master side usually uses a lot more battery meaning it will die first (and need to charge) before the other side, for this reason a lot of wireless split keyboards choose the master side depending on higher battery voltage each time they power up

(also, having the ability to plug in either side with USB and have it work automatically is cool)

once again thanks for your work, I'll try my hand at helping with code once I have hardware that supports it and I get further in the rust embedded book

HaoboGu:
> I just wanted to comment here about something, for wireless split keyboards being able to use either side as the master is fairly important, because the master side usually uses a lot more battery meaning it will die first (and need to charge) before the other side, for this reason a lot of wireless split keyboards choose the master side depending on higher battery voltage each time they power up

Thanks a lot for the reminder! 

I did consider the auto-master-switching feature, but I found it is quite difficult since RMK implements the split keyboard in a single-master + multiple-slave way. That means it's possible to make a multi-split keyboard, but brings the complexity of choosing the master. I haven't got a good idea about how to implement it, do you have any reference, or code example for how to implement it? I'd like to add this awesome feature.

yo-boy:
I actually misremembered, they just use random numbers here https://github.com/vE5li/butterware 

But this should be helpful since it's also in rust and embassy, the relevant function is in the split determine.rs file

I also thought zmk already had this implemented but apparently not, maybe I read the to do list long ago and assumed it would be done four years later, lol

Having it over more than two devices will be quite a bit more complex, you could have a specific config for like if it's just two halves then you can use it, and if it's more you have just one central? That would make things easier and would (in my opinion) cover most cases, but having it be generic and for any number of pieces would be better

HaoboGu:
Thanks for the reference! I'll take a look. Currently I'm focusing on implementing basic functionalities of the split keyboard(there is still lots of work to do..). I'll keep it in mind making sure we can add this feature with little effort. 

If you want to involve, I'd be very happy to help:D

PR #66, opened 2024-09-10
fix: correct grammar in error message

rmsthebest:
Hello,
Cool project!
Sorry for the driveby, I was just taking an initial look at the code and spotted the typo.

Not sure if you wanted to say `has died` or `is dead`. I guess both are true, but the way it is currently written with present tense is and past tense died is not grammatically correct.


HaoboGu:
Thanks a lot!

Issue #71, opened 2024-09-19
Keyboard code changes tracking


davidqge:
Following are the problems I am running into, and the change I think need to fix them:

- [x] mod tap is un-usable. I think the problem is while mod key is in delay mode, the second key was processed out of order. The channel should queue keypress before keymap processing, instead of after.
- [x] holding the key will keep sending key reports even though they haven't changed. The scroll of logs prevent look at the timing of wrong key presses. I think this can be solved by queue above too.
- [ ] keyboard output should change to keycode instead of hid report. To allow enter BLE password, and allow more flexibility in how to send report.
- [x] mod tap, mouse button 2, quick tap would press mouse button 2, not release it. I think above change the keyboard to output keycode would ensure that keycode are generated properly like non mouse keys.
- [ ] Scanner currently output high/low, this would short circuit on matrix without diode. Change to open drain low, which is designed for such task.

davidqge:
Here is log for mod taping TAB , released properly

`
19.520418 DEBUG matrix true 4 5
19.631786 DEBUG matrix false 4 5
19.631851 DEBUG Release tap hold, got TAP: Key(Tab), KeyState { pressed: true, changed: true, hold_start: Some(Instant { ticks: 19520480 }) }
19.642110 DEBUG Sending keyboard report: [0x2B, 0x0, 0x0, 0x0, 0x0, 0x0], modifier: 0
19.642214 DEBUG BLE notify 8 [0x0, 0x0, 0x2B, 0x0, 0x0, 0x0, 0x0, 0x0]
I (19642) NimBLE: GATT procedure initiated: notify;
I (19642) NimBLE: att_handle=31

19.643959 DEBUG Sending keyboard report: [0x0, 0x0, 0x0, 0x0, 0x0, 0x0], modifier: 0
19.650495 DEBUG BLE notify 8 [0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0]
I (19650) NimBLE: GATT procedure initiated: notify;
I (19650) NimBLE: att_handle=31
`
But mod tap mouse button are not released
`
210.628076 DEBUG matrix true 6 4
210.740198 DEBUG matrix false 6 4
210.740266 DEBUG Release tap hold, got TAP: Key(MouseBtn2), KeyState { pressed: true, changed: true, hold_start: Some(Instant { ticks: 210628131 }) }
[...]

PR #74, opened 2024-09-24
Permissive hold

davidqge:
Implemented the changes discussed here.
[Keyboard code changes tracking #71](https://github.com/HaoboGu/rmk/issues/71#issue-2536974707)

Tap Hold works perfect now.

No longer need to slow typing.


rmk/src/keymap.rs
```diff
@@ -185,41 +180,17 @@ impl<const ROW: usize, const COL: usize, const NUM_LAYER: usize> KeyMap<ROW, COL
...
 
-    /// Fetch the action in keymap, with layer cache
-    pub(crate) fn get_action_with_layer_cache(
-        &mut self,
-        row: usize,
-        col: usize,
-        key_state: KeyState,
-    ) -> KeyAction {
-        if key_state.is_releasing() {
-            // Releasing a pressed key, use cached layer and restore the cache
```
HaoboGu, line 196: Why removing layer cache? Removing it will lead the un-releasing key when switching layers

rmk/src/keyboard.rs
```diff
@@ -152,255 +143,188 @@ impl<'a, M: MatrixTrait, const ROW: usize, const COL: usize, const NUM_LAYER: us
...
-                        .borrow_mut()
-                        .get_action_with_layer_cache(row_idx, col_idx, ks);
-                    let tap_hold = match action {
-                        KeyAction::TapHold(tap_action, hold_action) => {
-                            Some((tap_action, hold_action))
+    pub(crate) async fn scan_matrix(&mut self) {
+        let mut ticker = Ticker::every(Duration::from_millis(1));
+        let mut holds: Vec<Holdingkey, 16> = Vec::new();
+        let mut holding = 0;
+        loop {
```
HaoboGu, line 197: This will limit the scan rate to 1KHZ. I want to improve the scan rate as fast as possible(currently it's set to 10KHZ, but I'll unlock it in the future), so I won't accept this change.

rmk/src/keyboard.rs
```diff
@@ -411,105 +335,32 @@ impl<'a, M: MatrixTrait, const ROW: usize, const COL: usize, const NUM_LAYER: us
...
-            if let Some(s) = key_state.hold_start {
-                let d = s.elapsed().as_millis();
-                if d > 200 {
-                    self.process_key_action_normal(hold_action, key_state, sender)
-                        .await;
-                }
-            }
+    async fn process_key_action_tap(&mut self, action: Action, mut pressed: bool) {
+        if pressed {
+            self.process_key_action_normal(action, pressed).await;
```
HaoboGu, line 367: A proper delay(> 1ms) should be set between two reports of tap

rmk/Cargo.toml
```diff
@@ -81,9 +81,9 @@ features = ["split"]
 cortex-m = { version = "0.7" }
 
 [features]
-default = ["col2row"]
```
HaoboGu: If you want to use row2col, just use `default-features = false`. No need to change default setting, since it's a breaking change

rmk/src/keymap.rs
davidqge: That info is stored in new struct HoldingKey. Just more convenient to have related info stored together, including hold_start. 
Also use less memory. Instead of store it for every location in matrix, it only need space for 16 holding keys.

rmk/src/keyboard.rs
davidqge: Since it's doing more things now, I just set it a conservative value, to make sure don't starve other code.

Yes, should be easy to change to 10KHz. But BLE and USB limit report to every 7ms, I think.

rmk/src/keyboard.rs
davidqge: Any delay adds to the additional 0.1ms delay you set before. By removing all the additional delays, I made scanner more responsive at guaranteed 1k per second.

This is the speed write to channel, no delay needed. Only the communication task, write to USB or BLE need delay after each report.

I think the channel should be a lot longer than 8. To avoid delay caused by channel full. Especially the Macro playback, need very long channel. Or move the macro playback to communication thread, so we can accommodate long macro without a long channel.

rmk/Cargo.toml
davidqge: You are right. I was doing global changes. Shouldn't change the default features.

rmk/src/keyboard.rs
HaoboGu: The report interval is typically 7.5ms for BLE and 1ms/0.125ms for USB2.0. 

Ideally this ticker should be removed to reach the maximum scanning speed. Then the scan rate can be much larger than actually report rate, ensuring best performance. Since there're lots of .await points, so other tasks won't be starved. 

The potential issue is when the report generating speed is larger than host request speed, then everything blocks. That's what I mentioned as `flooding report`

rmk/src/keymap.rs
HaoboGu: What will happen if user holds more than 16 keys?

rmk/Cargo.toml
davidqge: Removed the default feature, got this error: 
main.rs
```
/examples/use_rust/nrf52832_ble/src/main.rs: 61,
mismatched types
expected array `[[[KeyAction; 4]; 3]; _]`
   found array `[[[KeyAction; 3]; 4]; 2]`
```
The keymap needs to be rotated.

I removed this open drain patch for now.

HaoboGu: Remove this default feature means that row2col matrix are used.

All other examples uses col2row by default, so removing it leads to this error

davidqge:
removed commit that change default feature.
restored 10k/s scan
added full sized keymap to reproduce stack protection fault

davidqge:
Reverted the part of change in Matrix trait, so split code can build.

rmk/src/keymap.rs
davidqge: new press are ignored. Until some keys are released.
How many simultaneously pressed keys do you want?

rmk/src/keyboard.rs
davidqge: block happens when channel between keyboard and communication task full. That's why I want to increase that one.
Other than occasional burst, caused by releasing key trigger the many previously hold keys, human won't able to type that fast to cause problem.

rmk/src/keymap.rs
HaoboGu: I want to record all pressed keys on the keyboard, since the number of keys varies, a fixed number of pressed key is not good imo.

rmk/src/keyboard.rs
HaoboGu: Yeah but it's still possible, when processing a macro or a mouse(trackpad), which may send lots of report continuously. Merging report if possible may help. Increasing the channel size would also help, but it's hard to say what size is good.

HaoboGu:
This PR contains lots of modifications, I suggest to split this to separate PRs, where each one focus on only one thing. 

btw force-push makes it harder to track what's changed between commits. If you want to make the commit history clean, splitting them to different PRs and squash merging is a better way imo.

HaoboGu:
hey @davidqge, I'm thinking about decoupling matrix and keyboard these days as you said, as well as the compatibility of other types of input devices. 

The following is my current thoughts:

1. The scanning, key processing and communication should be three tasks, the event flow is: scanning -> processing -> communication
2. Scanning task and key processing task emit one report per action, while communication task merges reports which are within one communication interval
3. The scanning task and key processing task then could be fully async, no more busy loops when idle
4. All other input devices are also independent tasks, sending events to key processing task
5. Then we can regard keyboard matrix as an input devices, which makes it possible to run a batch of input device tasks together

There are some potential drawbacks:
1. The increasing inter-task communication may have negative compact on latency
2. The increasing number of tasks makes the latency less unpredictable
3. The size of communication channels is hard to determine: we don't want one slow processing blocks all other reports
4. Tap/hold still needs special processing

I've written a POC implementation in `feat/input_device` branch. Because part of this PR does similar thing, so I'd like to get some feedbacks of this refactoring. If you have any ideas/suggestions/questions, feel free to comment :D

davidqge:
I think only 2 is needed. You need channel between them because delay caused by usb/ble send.
Scanning and Tap Hold processing should be in one task. And Macro playback, Magic processing, and report on the communication task.
Store action in channel rather than report, that gave you more flexibility. Put Macro playback after channel will reduce the need for large channel.

HaoboGu:
Yeah i agree with u about it. but I still decoupled the matrix and key processing tasks, which takes matrix itself as an input device and makes it possible to add more input devices with little effort. all other devices can be configured from vial too.

Another benefit is that I finally makes all tasks fully async, including mouse and taphold processing, which simplifies the split code a lot and theoretically has lower power consumption. 

The bad side is just like what you said, it may require large channels, and it may perform worse on latency and even worse when the channel is blocking. anyway, there is quite a lot room for improvement.

PR #70, opened 2024-09-19
Esp32c6 Vial and storage works

davidqge:
Last change Also improved full sized keyboard key loading time.

PR #77, opened 2024-09-25
Ble vial service for Windows

HaoboGu:
This PR tries to make vial recognize RMK BLE keyboard. This is currently possible only for Windows because Windows regard BLE HID devices as a normal HID device and use BLE driver automatically.

A successful try is https://github.com/HaoboGu/rmk/pull/70. #70 cannot work with vial-web since vial-web uses report_id 0. 

This PR tries to extract BLE vial service to a separate one and makes vial-web recognize RMK keyboard.

Note: This PR is actually a temporary solution only for Windows. To achieve general vial-over-BLE, the BLE vial service should be a customized service(not HID!). This is because Apple devices don't expose BLE-HID services at all. So, a host tool(like a wireless version of via/vial) needs to be implemented finally.

### DIFF
--- a/.github/workflows/build esp.yml
+++ b/.github/workflows/build esp.yml
@@ -23,19 +23,31 @@ jobs:
         run: espup install
       - name: Build esp32c3_ble
         working-directory: ./examples/use_rust/esp32c3_ble
-        run: cargo +esp build --release
+        run: |
+          cargo +esp build --release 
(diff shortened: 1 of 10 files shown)


HaoboGu:
@davidqge I just made esp32c6 recognized by [vial.rocks](https://vial.rocks), could you help to verify this?

Issue #39, opened 2024-05-27
Macropad (RP2040) 'matrix-less' switch support?


erdii:
Heya :wave: 

I was wondering: does this firmware support the "matrix-less" inputs on the Adafruit Macropad [1]?
The macropad has 12 keyboard switches, each connected to it's own GPIO pin.

Cheers,
Josh

[1] https://www.adafruit.com/product/5128

HaoboGu:
Hi!

Direct pin is not supported right now, but adding this feature isn't hard imo. If you really want it, I can add it in next weeks, or you can do a PR and I'm happy to review it :D

erdii:
Ah that's a bummer.

I haven't written any rust actively and my experience only revolves around gargabe-collected languages so far, so I'm not sure if my rust-foo is good enough to contribute without getting overwhelmed by the steep learning curves of rust, cross-compiling rust for embedded and embassy but I'll dabble a few more hours to try and find out. :)

Thanks for creating this project and being a maintainer of it and I'd like to acknowledge that I don't have any entitlement to ask you for adding this feature (in the next weeks or at all). 

That said, If you'd like to share any pointers on where to start, I'd happily take them. I'd also happily download a version of rmk that includes the feature.

Cheers! :)

HaoboGu:
You can follow the first two sections of [user guide](https://haobogu.github.io/rmk/guide_overview.html) to setup Rust and RMK environment. 

If you're not familiar with Rust nor Rust embedded, the official [Rust book](https://doc.rust-lang.org/book/) and [the Rust embedded book](https://docs.rust-embedded.org/book/) are the good start points of your Rust journey.

Happy coding! :D

HaoboGu:
I just did some investigation. Current `matrix.rs` implementation is limited to traditional keyboard matrix, but it can be abstracted easily to a trait which is used in `keyboard.rs`:

```rust
pub(crate) trait KeyboardMatrix {
    fn new() -> Self;
    fn get_key_state(&mut self, row: usize, col: usize) -> KeyState;
    fn update_timer(&mut self, row: usize, col: usize);
    async fn scan(&mut self);
    #[cfg(feature = "async_matrix")]
    async fn wait_for_key(&mut self); 
}
```

Then the direct pin matrix can be implemented with very little effort. I currently just doesn't have enough time to implement it. But if anyone want to do, I'd happy to give suggestions & guidance about it.

davidqge:
[...]

PR #116, opened 2024-10-29
Add DirectPinMatrix

decaday:
I have implemented a DirectPinMatrix, including the DirectPinMatrix struct, modules, configuration file options, procedural macro entry functions, examples, and documentation.

Related issue:  #39 .


rmk/src/direct_pin.rs
```diff
@@ -0,0 +1,319 @@
...
+                futs[row_idx] = Some(select_array(futs_row.map(|option| option.unwrap())));
+            }
+            let _ = select_array(futs.map(|option| option.unwrap())).await;
+        } else {
+            let mut futs: [Option<_>; ROW] = [const { None }; ROW];
+
+            for (row_idx, direct_pins_row) in self.direct_pins.iter_mut().enumerate() {
+                let mut futs_row: [Option<_>; COL] = [const { None }; COL];
+                for (col_idx, direct_pin) in direct_pins_row.iter_mut().enumerate() {
+                    futs_row[col_idx] = Some(direct_pin.wait_for_high());
```
HaoboGu: Can be simplified with:

```rust
if self.low_active {
    futs_row[col_idx] = Some(direct_pin.wait_for_low());
} else {
    futs_row[col_idx] = Some(direct_pin.wait_for_high());
}
```

decaday: `wait_for_low` and `wait_for_high` return different type of `Future`.
Simplifing like this will lead to [type inference error](https://github.com/HaoboGu/rmk/pull/116/commits/0a6dd26847c0b6eafb708e4417e6a3e8fddeb2d6).

docs/src/keyboard_configuration.md
```diff
@@ -228,11 +245,25 @@ usb_enable = true
...
+# `matrix_type` is optional. Default is "normal"
+matrix_type = "normal"
 # Input and output pins
 input_pins = ["PIN_6", "PIN_7", "PIN_8", "PIN_9"]
 output_pins = ["PIN_19", "PIN_20", "PIN_21"]
 # WARNING: Currently row2col/col2row is set in RMK's feature gate, configs here do nothing actually
 
+# Direct Pin Matrix is a Matrix of buttons connected directly to pins. It conflicts with the above.
+matrix_type = "direct_pin"
+direct_pins = [
```
HaoboGu, line 257: This type of direct pin matrix seems invalid now:


```toml
direct_pins = [
    ["PIN_0", "PIN_1", "PIN_2"],
    ["PIN_3", "_", "PIN_5"]
]
```

It's common that the key matrix doesn't have keys at all positions.
Normal pin matrix doesn't have this issue, since the scanning algorithm automatically ignore those positions. But for direct pin matrix, we should allow users ignore some positions in the matrix(even though users can still fill a dummy pin here to bypass)

decaday: Done! 
https://github.com/HaoboGu/rmk/pull/116/commits/0305aaa1c815d6bb870c6595fc77c55cd72fe716

rmk/src/direct_pin.rs
decaday: I'm now using a 1-D heapless::Vec. The code is shorter, but simplifying it as you mentioned seems challenging.

PR #123, opened 2024-11-04
One-shot key support

### LINKED ISSUE #122: One-shot key support
Reference:

- zmk sticky key: https://zmk.dev/docs/keymaps/behaviors/sticky-key
- qmk one-shot key: https://docs.qmk.fm/one_shot_keys

pcasotti:
Closes #122

### DIFF
--- a/rmk/src/keyboard.rs
+++ b/rmk/src/keyboard.rs
@@ -245,7 +245,7 @@ impl<'a, const ROW: usize, const COL: usize, const NUM_LAYER: usize>
                     .await;
             }
             KeyAction::OneShot(oneshot_action) => {
-                self.process_key_action_oneshot(oneshot_action).await
+                self.process_key_action_oneshot(oneshot_action, key_event).await
             }
             KeyAction::LayerTapHold(tap_action, layer_num) => {
                 let layer_action = Action::LayerOn(layer_num);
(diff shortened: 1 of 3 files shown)


pcasotti:
Still needs to implement timeout, also needs to filter the events after the one shot so other modifiers or layer actions don't release the modifier.

Current implementation releases the modifier on the next key press, doing it on key release would be difficult since we would need to process every other key press inside `process_key_action_oneshot`. Maybe the other approach discussed [here](https://github.com/HaoboGu/rmk/issues/122#issuecomment-2453224924) can give more flexibility.

HaoboGu:
> Current implementation releases the modifier on the next key press, doing it on key release would be difficult since we would need to process every other key press inside process_key_action_oneshot. Maybe the other approach discussed https://github.com/HaoboGu/rmk/issues/122#issuecomment-2453224924 can give more flexibility.

The issue here is we want some keys not to be affected by one-shot key, so we have to do another full processing logic inside `process_key_action_oneshot`.

pcasotti:
@HaoboGu The core feature seems to be done. I'd like to know what are your ideas on how the one shot should work on some edge cases, I'll need that to add to the documentation as well.

pcasotti:
What about specifying which type of actions affect the one shot state? For example, pressing a modifier (or one shot modifier) after a one shot layer should disable the layer, so that `OSL(1) -> LShift (hold)  -> key` will result in `LShift + key` if `LShift` is on layer `1` and `key` is on layer `0`. But pressing a modifier after a one shot modifier should do nothing.



> > @HaoboGu The core feature seems to be done. I'd like to know what are your ideas on how the one shot should work on some edge cases, I'll need that to add to the documentation as well.
> 
> Cool! I'm thinking about can we have one-shot behavior for any `Action`? Currently only `Action::Modifier(m)` and `Action::LayerOn(n)` are supported.
> 
> For example, `Action::LayerOff(n)` could also be used as one-shot behavior, which is very similar with `Action::LayerOn(n)`

Sure, i'll work on that.

HaoboGu:
> What about specifying which type of actions affect the one shot state?

Yes, I think we should have this. Some `Action`s will have no effect when making it one-shot.

pcasotti:
Ok, just now realized that [zmk](https://zmk.dev/docs/keymaps/behaviors/sticky-layer#advanced-usage), has the answer, I think this is a good rule to follow.

HaoboGu:
If you find it difficult to achieve, we can also merge this PR as-is, since vial supports only OSL and OSM 😅

pcasotti:
> If you find it difficult to achieve, we can also merge this PR as-is, since vial supports only OSL and OSM 😅

Yeah, i think we can leave the "All `Actions`" part to another issue. I'll just change the way OSL works to match zmk as I linked above

PR #125, opened 2024-11-06
Refactor tap/hold processing, add better HRMs support out-of-the-box

### LINKED ISSUE #118: Better HRM(home row mod) support
Reference:

- QMK tweak for better HRM: https://github.com/stasmarkin/sm_td
- ZMK config with "Timeless HRM": https://github.com/urob/zmk-config

### LINKED ISSUE #124: Tap hold implementation issue
I've noticed there's a slight issue in the logic for the tap hold implementation. Specifically, relating to this section of code in the `process_key_action_tap_hold` function:
```rust
                embassy_futures::select::Either::Second(e) => {
                    if e.row == key_event.row && e.col == key_event.col {
                        // If it's same key event and releasing within 200ms, trigger tap
                        if !e.pressed {
                            let elapsed = self.timer[col][row].unwrap().elapsed().as_millis();
                            debug!("TAP action: {}, time elapsed: {}ms", tap_action, elapsed);
                            self.process_key_action_tap(tap_action, key_event).await;

                            // Clear timer
                            self.timer[col][row] = None;
                        }
                    } else {
                        // A different key comes within the threshold, trigger hold + that key

                        // Process hold action first
                        self.process_key_action_normal(hold_action, key_event).await;

                        // The actual processing is postponed because we cannot do recursion on async function without alloc
                        // After current key processing is done, we can process events in queue until the queue is empty
                        if self.unprocessed_events.push(e).is_err() {
                            warn!("unprocessed event queue is full, dropping event");
                        }
                    }
                }
```
Correct me if I'm wrong, but the `else` here can also be triggered by a key release event? What I've noticed is that if I'm typing fast, the tap of a tap hold key fails to trigger, this is because I press the tap hold key before releasing another key, so when I then release the other key it causes the code to go into this `else` condition, and then when I release the tap hold key (expecting the tap action to occur) it doesn't do anything since the else triggered the `hold` action.

I'm not quite familiar enough with Rust to make a PR with a fix, but in my head it sounds like what we'd want to do is loop until we timeout or get a _keypress_ event. If we get a key release event we should just process that action normally and then go back to waiting for a timeout or keypress.

### LINKED ISSUE #162: Hold after tap

HaoboGu:
Current tap/hold implementation has some issues, see #118, #124. This PR proposes an approach which addresses the issue, as well as HRM config.

The proposed solution is actually a mixture of [urob's zmk-config](https://github.com/urob/zmk-config), [Achordion](https://getreuer.info/posts/keyboards/achordion/index.html) and [sm_td](https://github.com/stasmarkin/sm_td). Many respect to the authors! 

The following is a simple description of new tap/hold processing:
When the keyboard processes tap/hold keys, it checks `prior_idle_time` first for fast typing, and delays the decision of tap/hold until the new incoming key's release plus a small time interval `post_wait_time`.

Closes #118 
Closes #124
Closes #162


HaoboGu:
@MarkuBu @azzamsa Hey, this is my first implementation of HRMs, could you have a try? Any feedbacks are welcome!

You can edit your `Cargo.toml` to use this branch:

```diff
- rmk = { version = "0.4.1", features = [
+ rmk = { git = "https://github.com/haobogu", branch = "feat/hrm", features = [
```

MarkuBu:
Sounds good. Will try it asap, but probably not before weekend

HaoboGu:
Tested, works good for now. So I'm merging this.

PR #132, opened 2024-11-10
feat(scripts): added via2rmk.py

whitelynx:
Added a straightforward script to convert from VIA/Vial layout syntax to RKM's keymap definitions. It can output either a TOML `keymap = [...]` section or a Rust `get_default_keymap()` function.


HaoboGu (commented):
lgtm

thanks!

PR #147, opened 2024-11-13
Tri Layer support

### LINKED ISSUE #143: Tri Layer support
Reference:
- QMK Tri Layers: https://docs.qmk.fm/features/tri_layer
- ZMK Conditional Layers: https://zmk.dev/docs/keymaps/conditional-layers

pcasotti:
Closes #143

- [x] Find a way to add Tri Layer options
- [x] Add options in toml and macros
- [x] Implement Tri Layer
- [x] Update docs

Implementation will be similar to ZMK, ignore `TriLayerUpper` and `TriLayerLower` key codes and use the layer state instead.
It will be limited to 2 activation layers ("upper" and "lower") and one Tri Layer configuration for now.

### DIFF
--- a/rmk-config/src/toml_config.rs
+++ b/rmk-config/src/toml_config.rs
@@ -10,6 +10,8 @@ pub struct KeyboardTomlConfig {
     /// Layout config.
     /// For split keyboard, the total row/col should be defined in this section
     pub layout: LayoutConfig,
+    /// Behavior config
+    pub behavior: Option<BehaviorConfig>,
     /// Light config
     pub light: Option<LightConfig>,
     /// Storage config
@@ -122,6 +124,11 @@ pub struct LayoutConfig {
     pub cols: u8,
     pub layers: u8,
     pub keymap: Vec<Vec<Vec<String>>>,
+}
+
+/// Configurations for actions behavior
+#[derive(Clone, Debug, Default, Deserialize)]
+pub struct BehaviorConfig {
     pub tri_layer: Option<TriLayerConfig>,
 }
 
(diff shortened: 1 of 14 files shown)


rmk-config/src/toml_config.rs
```diff
@@ -122,6 +122,15 @@ pub struct LayoutConfig {
     pub cols: u8,
     pub layers: u8,
     pub keymap: Vec<Vec<Vec<String>>>,
+    pub tri_layer: Option<TriLayerConfig>,
```
HaoboGu, line 132: should we have another field for advanced keyboard config? such as tri-layer/HRM/etc, corresponds to KeyboardOptionConfig？

pcasotti: Ok, I was also thinking about giving it a better name instead of `KeyboardOptionConfig`, maybe `BehaviorConfig`?

HaoboGu: I like this name

HaoboGu: so the section in toml file should be ［behavior］?

pcasotti: yes, but should it be inside the layout section or a separate top level? `[behavior]`vs `[layer.behavior]`

HaoboGu: I prefer a separate top level `[behavior]`, since there are lots of non-layer behaviors we can customize

HaoboGu:
lgtm!

HaoboGu:
let me merge this, thanks a lot!

PR #150, opened 2024-11-18
Add option to customize one shot timeout

pcasotti:
Add config under `[behavior]` to customize OSM and OSL timeout.

I believe tap/hold and HRM timeout should wait for #125


HaoboGu:
lgtm, thanks!

Issue #160, opened 2024-12-01
direct_pin split


willpuckett:
When attempting to build for a direct_pin split, I'm receiving the following error:

```
   Compiling embassy-nrf v0.2.0
   Compiling nrf-softdevice v0.1.0 (https://github.com/embassy-rs/nrf-softdevice?rev=d5f023b#d5f023ba)
   Compiling rmk v0.4.4
error: Parse `keyboard.toml` error: missing field `input_pins`
 --> src/peripheral.rs:6:1
  |
6 | #[rmk_peripheral(id = 0)]
  | ^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: this error originates in the attribute macro `rmk_peripheral` (in Nightly builds, run with -Z macro-backtrace for more info)

error: Parse `keyboard.toml` error: missing field `input_pins`
 --> src/central.rs:6:1
  |
6 | #[rmk_central]
  | ^^^^^^^^^^^^^^
  |
  = note: this error originates in the attribute macro `rmk_central` (in Nightly builds, run with -Z macro-backtrace for more info)

error: `#[panic_handler]` function required, but not found

error: could not compile `ehrbl` (bin "peripheral") due to 2 previous errors
warning: build failed, waiting for other jobs to finish...
error: could not compile `ehrbl` (bin "central") due to 2 previous errors
```

I built one half of the board successfully last night as a numpad.

My `keyboard.toml` for the split:

```
# Documentation: https://haobogu.github.io/rmk/keyboard_configuration.html
[keyboard]
name = "ehrbl"
product_name = "ehrbl"
vendor_id = 0x4c4b
product_id = 0x4643
manufacturer = "haobo"
chip = "nrf52840"
board = "XIAO BLE"

[layout]
# WARNING: Currently row2col/col2row is set in RMK's feature gate, configs here do nothing actually
# row2col = true
rows = 4
cols = 8
layers = 3
keymap = [
    [
        ["B", "F", "D", "L", "P", "O", "U", "Y"],
        ["C", "S", "T", "R", "N", "A", "I", "H"],
        ["V", "W", "G", "M", "K", ",", ".", "'"],
        ["_", "_", "LT(2, Backspace)", "LT(1, Space)", "E", "Enter", "_", "_"]
    ],
    [
        ["KpEnter", "Kp7", "Kp8", "Kp9", "KbVolumeDown", "KbMute", "KbVolumeUp", "_"],
[...]

PR #140, opened 2024-11-11
Support Direct Pin matrix for split keyboards.

DivineGod:
* [x] Update config to handle Matrix for split firmware
* [x] Implement direct_pin support in split central run function
* [x] Implement direct_pin support in split peripheral run function
* [x] Add peripheral examples
* [x] Add central examples
* [x] Update split config examples
* [ ] Add non-config example
* [x] Update relevant documentation

PR #163, opened 2024-12-03
fix(direct-pins): rust cannot infer the specific type of options for shadowing variables

drindr:
Support multi "_" GPIO in each row in direct-pins matrix.
Related issue and PR: #140 and #160


HaoboGu:
Looks good, thanks!

PR #169, opened 2024-12-06
split: replace unwraps with expect

wez:
I was facing an inscrutable unwrap error when porting my last-release-compatible keyboard.toml to master.

This commit switches the unwraps for expects that give some more useful diagnostics


HaoboGu:
Thanks a lot!

PR #173, opened 2024-12-10
Add py32f07x use_rust example

decaday:
I've written a USB Driver for PY32 in https://github.com/py32-rs/py32-hal/pull/19.
Running RMK works perfectly.
Supported MCUs: `PY32F070`, `PY32F071`, `PY32F072`


examples/use_rust/py32f07x/Cargo.toml
```diff
@@ -0,0 +1,65 @@
...
+    "time-driver-tim3"
+] }
+
+
+embassy-executor = { version = "0.6.0", features = [
+    "defmt",
+    "arch-cortex-m",
+    "executor-thread",
+    "integrated-timers",
+    "task-arena-size-32768",
```
HaoboGu: py32f072 has only 16K RAM, so setting task arena size to 32768 might not work.

decaday: Fixed. Thanks!

HaoboGu:
Thanks!!!

PR #175, opened 2024-12-11
Speedup the startup by using iterator to load keymap

### LINKED ISSUE #165: Fix latency after boot
related issue: https://github.com/tweedegolf/sequential-storage/pull/65

HaoboGu:
Closes #165


rmk/src/storage/mod.rs
```diff
@@ -562,44 +561,31 @@ impl<F: AsyncNorFlash, const ROW: usize, const COL: usize, const NUM_LAYER: usiz
         &mut self,
         keymap: &mut [[[KeyAction; COL]; ROW]; NUM_LAYER],
     ) -> Result<(), ()> {
-        for (layer, layer_data) in keymap.iter_mut().enumerate() {
```
pcasotti, line 565: I believe the similar code in `initialize_storage_with_config` can be refactored as well.

HaoboGu: I'm afraid not, `initialize_storage_with_config` is executed only once, it stores default keymap to the storage. The storing can only be done one by one.

PR #189, opened 2024-12-15
Initial migration from const-generics to typenum

Ben-PH:
This is an initial PoC of what a migration to use of `typenum` and `GenericArray` might look like.

### DIFF
--- a/rmk/Cargo.toml
+++ b/rmk/Cargo.toml
@@ -70,6 +70,8 @@ esp-idf-svc = { version = "0.49", default-features = false, optional = true }
 
 # Document feature
 document-features = "0.2"
+typenum = { version = "1.17.0", features = ["no_std"] }
+generic-array = "1.1.1"
 
 [package.metadata.docs.rs]
 features = ["split"]
--- a/rmk/src/debounce/default_bouncer.rs
+++ b/rmk/src/debounce/default_bouncer.rs
@@ -1,4 +1,6 @@
 use embassy_time::Instant;
+use generic_array::{sequence::GenericSequence, ArrayLength, GenericArray};
+use typenum::NonZero;
 
 use crate::matrix::KeyState;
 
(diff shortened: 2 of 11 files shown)


thomasaarholt:
(Not the developer, just curious)

Could you expand on what the motivation for moving away from const generics is here?

Ben-PH:
typenum allows you to do things like specify legal values at compile time, as well as removes the need to choose an integer type until runtime.

For example, if you wanted to implement a "sub-matrix" extension trait, you could constrain it so its width and height is less than the matrix it's extending. when you are placing it in the "super-matrix", you can constrain it so that it's left most column, plus its width is less than the width of the super-matrix, and similarly for the highest row. 

this is all at compile time.

as for integer types, you do not need to choose usize, u32, etc., until an actual machine-representation of the integer is required.

thomasaarholt:
Thank you for the explanation!

Ben-PH:
Another way to put it:

const generics allow you to specify "use u32/usize/u64/etc here", typenum says "use any integer value here", and you can specify what "legal" is (non-zero, even/odd, the result of arithmetic, etc)

oh yeah, that's the other thing: you can do arithmetic with typenum. no need for `SIZE` because you can just derive that within the type system my multiplying the width-type by the height-type.

HaoboGu:
I've looked through the PR, afaict, `typenum` brings us the compile-time checking and calculation. But now I tend to not use it because:

1. The last release of `typenum` was over 1 year ago, I'm not sure whether it's been actively maintained.
2. The compile-time checking part can be done in proc-macro, by checking values set in `keyboard.toml`. We've already had some, like: https://github.com/HaoboGu/rmk/blob/6ea066a128b0f1d04c5ab383ba78a98137fa6556/rmk-macro/src/keyboard_config.rs#L372-L386
3. The compile-time calculation part could be covered by https://github.com/rust-lang/rust/issues/76560, I prefer to use official Rust feature, rather than a third-party crate. `typenum` has a long history, it appeared even before `const_generics`. But for the future, I tend to use features provided by official Rust

Ben-PH:
After our discussion in DMs, I'd be happy to re-write the PR so that it only affects the internal code-base, and not the public facing API.

PR #192, opened 2024-12-16
Add basic POC input device support and rotary encoder implementation as an example

HaoboGu:
update: input device support will be separated to several PRs. This PR adds traits for input devices and input processors, implements those traits for rotary encoders(as an example)


HaoboGu:
It seems that input devices requires lots of API changes.. It's a breaking change and current API isn't good for end users. So I'll revert the API changes and keep only traits and sample rotary encoder implementations.

To support the input device well, the whole architecture of the project needs to be upgraded. I'll do it in other PRs.

Issue #180, opened 2024-12-12
Unexpected Behavior on Direct Split


willpuckett:
I've been testing RMK on a direct split, and I'm having a few issues, hopefully due to user error...

I'm building from the main branch. For compatibility with direct split and improved hrms, my Cargo.toml has:

```
[dependencies]
rmk = { git = "https://github.com/HaoboGu/rmk",  features = [
...
```

I have also ensured the Xiaos are on the latest soft device.

After flashing RMK, the left (central) portion of the keyboard works as anticipated. The right half, however, only outputs keypresses on the top two of four rows. 

If I hold a key on the top two rows of the right (peripheral) side and press a previously non-working key on one of the two bottom rows, both keypresses are sent. Holding a key on the left side and pressing one of the misbehaving keys only outputs the left key.

I was afraid I might have misconfigured my row/col offsets, but referencing #164 it seems they are correct.

It also doesn't seem to be a hardware problem: the keys are working normally in my zmk build both before and after flashing rmk.

I double checked my split.peripheral.matrix and everything seems to be entered correctly.

Changing the key assignments of the misbehaving keys in Vial has no effect.

Any thoughts or directions are greatly appreciated! 

My current keyboard.toml:

```
# Documentation: https://haobogu.github.io/rmk/keyboard_configuration.html
[keyboard]
name = "ehrbl"
product_name = "ehrbl"
vendor_id = 0x4c4b
product_id = 0x4643
manufacturer = "haobo"
chip = "nrf52840"
board = "XIAO BLE"

[layout]
# WARNING: Currently row2col/col2row is set in RMK's feature gate, configs here do nothing actually
# row2col = true
rows = 4
cols = 8
layers = 3
keymap = [
    [
        ["B", "F", "D", "L", "P", "O", "U", "Y"],
        ["C", "S", "T", "R", "N", "A", "I", "H"],
        ["V", "W", "G", "M", "K", "Comma", "Dot", "Quote"],
        ["__", "__", "LT(2, Backspace)", "LT(1, Space)", "E", "Enter", "__", "__"]
    ],
    [
        ["KpEnter", "Kp7", "Kp8", "Kp9", "KbVolumeDown", "KbMute", "KbVolumeUp", "__"],
        ["Kp0", "Kp4", "Kp5", "Kp6", "Left", "Down", "UP", "Right"],
        ["KpDot", "Kp1", "Kp2", "Kp3", "Home", "PageDown", "PageUp", "End"],
[...]

PR #200, opened 2024-12-17
fix(direct): the size of futs for waiting is incorrect

drindr:
Split-direct-keyboard cannot start scanning when some keys are pressed in direct matrix
some futures might skip the `select`

related issue:  #180


HaoboGu:
wooooooooooooow that seems the cause！ thanks a lot！

willpuckett:
Typing on my direct split now... All keys are working!

PR #203, opened 2024-12-18
ci: Optimize GitHub Actions workflow to use self-hosted runners.

liyang8246:
- Add determine-runner step, use self-hosted runner or fallback to latest Ubuntu version.
- Remove cargo-bins/cargo-binstall, use cargo install instead.
- Add Python setup step for Python 3.11.
- Add Rust toolchain setup step.


HaoboGu:
Thanks！

PR #208, opened 2024-12-19
Use 2M PHY

HaoboGu:


HaoboGu:
After enabling 2M PHY: 

idle current: 28uA
active current: 2.86mA

compared with 1M PHY:

idle current: 28uA
active current: 2.91mA

PR #211, opened 2024-12-20
Support log over USB serial

HaoboGu:


HaoboGu:
Known issue: the logger requires USB connection, using USB logger makes it more complex when switching between USB and BLE.

One solution is to run the usb logger + keyboard whenever the USB is connection, but if the communication type is set to BLE, the report won't be sent to the USB reporter. Though, this needs the architecture update, which will be done later. So this PR now blocks.

HaoboGu:
I'll split this PR to two, the first adds `defmt` feature and the second adds usb logger.

PR #214, opened 2024-12-20
ci: github token ref moved from secrets to vars

liyang8246:
- forked branches can't access secrets

need to add the token to the location shown in the figure below
image


HaoboGu:
thanks

PR #218, opened 2024-12-24
fix(config) check if either board or chip is set

gnullme:
Fixes the check if either board or chip is set in config.


HaoboGu (approved):
Thanks!

Issue #221, opened 2024-12-26
ModifierTapHold from keyboard.toml


hariseldon78:
Hello, thanks for this software. I am trying to implement the home row mods, and i would like to setup them in the keyboard.json. I think there is no way to do that at the moment, looking around in the code i only found the "LT(" parsing code.
Is there any plan of supporting it, and what will be the syntax? If it's already defined I could go on and try to post a PR myself.

HaoboGu:
Yes, ModifierTapHold can only be set by vial right now. Adding it should be easy, though.

hariseldon78:
ok. For adding it we could go 2 routes: supporting the already familiar syntax from QMK, which is
'''
LCTL_T(<keycode>)
LALT_T(<keycode>)
LSFT_T(<keycode>)
RCTL_T(<keycode>)
RALT_T(<keycode>)
RSFT_T(<keycode>)
'''

or a single pattern, like `MTH(<modifier>,<keycode>)`.

Which one would you prefer? (in my opinion the second one is cleaner, or we could do both)

hariseldon78:
ok, I'll just do both, and then we can remove the ones we don't like. I'll start tonight!

HaoboGu:
I prefer the second. It could be Mt(modifier_combination, key)

PR #222, opened 2024-12-27
Modifier tap hold in keyboard toml

hariseldon78:
This pull request closes issue #221.

I have been able to test the MTH(A,LShift) and it works correctly, but for some reason the TH(X,Y) is not working yet: has it been tested before? I see that Via does not have a binding for the TapHold.

I used all caps for MTH and TH for consistency with the other Layer actions, and i used the ordering <key>, <modifiers> for consistency with WM and LM.

### DIFF
--- a/docs/src/keyboard_configuration.md
+++ b/docs/src/keyboard_configuration.md
@@ -154,7 +154,7 @@ The key string should follow several rules:
 
   The definitions of those operations are same with QMK, you can found [here](https://docs.qmk.fm/#/feature_layers). If you want other actions, please [fire an issue](https://github.com/HaoboGu/rmk/issues/new).
 
-4. For modifier-tap-hold, use `MTH(key, modifier)` where the modifier can be a chain like explained on point 1
+4. For modifier-tap-hold, use `MTH(key, modifier)` where the modifier can be a chain like explained on point 1. For example for a Home row modifier config you can use `MTH(F,LShift)`
 
 5. For generic key tap-hold, use `TH(key-tap, key-hold)`.
 
--- a/docs/src/keyboard_configuration.md
+++ b/docs/src/keyboard_configuration.md
@@ -154,7 +154,7 @@ The key string should follow several rules:
 
   The definitions of those operations are same with QMK, you can found [here](https://docs.qmk.fm/#/feature_layers). If you want other actions, please [fire an issue](https://github.com/HaoboGu/rmk/issues/new).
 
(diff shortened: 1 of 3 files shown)


hariseldon78:
i also fixed Lgui and Rgui to LGui to RGui, let me know if that was a bad idea.

docs/src/keyboard_configuration.md
```diff
@@ -151,9 +151,12 @@ The key string should follow several rules:
     7. Use `"TT(n)"` to create a layer activate or tap toggle action, `n` is the layer number
     8. Use `"TG(n)"` to create a layer toggle action, `n` is the layer number
     9. Use `"TO(n)"` to create a layer toggle only action (activate layer `n` and deactivate all other layers), `n` is the layer number
-    
+
   The definitions of those operations are same with QMK, you can found [here](https://docs.qmk.fm/#/feature_layers). If you want other actions, please [fire an issue](https://github.com/HaoboGu/rmk/issues/new).
 
+4. For modifier-tap-hold, use `MTH(key, modifier)` where the modifier can be a chain like explained on point 1. For example for a Home row modifier config you can use `MTH(F,LShift)`
```
HaoboGu: I prefer `MT`, which is same with QMK, and consistent with `LT`, what do you think?

rmk-macro/src/layout.rs
```diff
@@ -37,6 +37,43 @@ fn expand_row(row: Vec<String>) -> TokenStream2 {
     quote! { [#(#keys), *] }
 }
 
+/// Get modifier combination, in types of mod1 | mod2 | ...
```
HaoboGu, line 76: Please add documentation here to specify what the five `bool`s represent. And there's a minor format issue: 
`){` -> `) {`

HaoboGu:
> I see that Via does not have a binding for the TapHold.

Yes, via/vial supports only ModifierTapHold and LayerTapHold. Only basic keycodes are supported in normal TapHold, and they're not exposed right now.

docs/src/keyboard_configuration.md
hariseldon78: sure, i didn't know about the QMK MT macro. I'll keep TH for the other one.

rmk-macro/src/layout.rs
hariseldon78: I decided to use a struct and refactor away the tuples. Also i implemented the macro expansion in the struct impl. Let me know if I did something wrong, I'm not expert on macros. The code compile, I'll test it now on the hardware.

hariseldon78: i confirm the hardware test is working for MT(A,LShift) .

HaoboGu (approved):
Thanks very much!

PR #224, opened 2024-12-29
fix(macro): use defaults defined in `TapHoldConfig`

pcasotti:
Use the default values defined in `TapHoldConfig` when generating from macros, insted of duplicating the values.


HaoboGu (approved):

HaoboGu:
Oh, thanks a lot!

Issue #225, opened 2025-01-01
Making more modules public


hyranno:
Requesting to make more modules public.
Hello and Happy New Year BTW.

### Context
I'm trying to make a keyboard with unconventional keyscan.
Then I wanted to write new struct implementing `matrix:MatrixTrait`, but `matrix` and relevant modules are not public.
So I request to make more modules public.

If you merge the codes for device into the rmk, making modules public is not necessary for it.
But making modules public is helpful when writing codes for minor devices which is not for merge.

Making lot of modules public may let new users confused.
If you want to emphasize commonly-used modules, how about making prelude module which re-exports commonly-used modules?

### Additional info
I'm working on my keyboard at https://github.com/hyranno/dflip_daisy .
I'm trying to make sequential keyscan using pins for column-clock, row-clock and reset.

HaoboGu:
Making `matrix` public makes sense. Could you please open a PR for it?

hyranno:
Can I take some time?
I haven't fully listed things to open to implement `MatrixTrait` and `run_rmk_...` using it.

hyranno:
Opened https://github.com/HaoboGu/rmk/pull/228
This should be enough to implement new `Matrix` and `run_rmk_...` outside rmk itself.

PR #228, opened 2025-01-03
Change/open modules public

hyranno:
PR for https://github.com/HaoboGu/rmk/issues/225
This should be enough to implement new `Matrix` and `run_rmk_...` outside rmk itself.


HaoboGu:
@hyranno My initial idea is to use `run_rmk` for APIs, and hide complex `initialize_xxx_keyboard_xx`. It seems that just making them public solves the problem for now, but I believe that there's something more we can do to make the entry APIs better. 

I have to admit that it's a very complicated problem, since the different MCUs require different input parameters, like USB, BLE, flash, etc. Do you have any ideas?

HaoboGu:
I'm merging it, leaving the problems to future PRs

PR #235, opened 2025-01-05
Handle embedded_io_async::Read/Write Guarantee

grabpot:
embedded_io_async::Read/Write __do not__ guarantee "that all available buffer space is filled", and could read/write less than buf.len(), and to handle this:
* Messages use postcard::from_bytes_cobs() to encode messages with COBS, using 0x00 byte as message separator
* SplitReader loops over serial.read() until 0x00 separator byte read
  * Loop handles receiving < 1 message on serial.read(), but also > 1 && < 2 messages (partial messages stored) and > 2 && < 3 (max. messages limited by buf.len()) messages (messages returned in vector)
  * Is "cancel-safe", with state stored in self, so no issue to be dropped by select() if interrupted by Timer/KeyEvent
* SplitWriter loops over serial.write() until all bytes written
* SPLIT_MESSAGE_MAX_SIZE, which sets buffer size, increased by 2 to account for COBS overhead

__NOTE__: I've only tested this with an rp2040 and my custom half-duplex UART driver. I don't expect any issues, but it should be tested with the standard BufferedUart or equivalent serial driver.


HaoboGu:
> embedded_io_async::Read/Write do not guarantee "that all available buffer space is filled", and could read/write less than buf.len()

So what's the problem? I think postcard could solve it automatically, deserialize the message from a partial filled buffer.

grabpot:
The main problem is that the serial Read trait isn't guaranteed to read all the bytes in a message. This is very likely if the UART bus is slower than the monitor cycle, but still quite likely if the UART buffer isn't a multiple of the message size (the internal ring buffer in BufferedUart will only return a contiguous slice, so when it gets to the end of the buffer, it won't return all the data in the buffer, even if its available).

The code could call postcard::take_from_bytes() after each serial read() and loop until it doesn't error, then save any used slice for the next read if necessary. This would skip (or replace) the second while loop, but otherwise handling a remaining partial message and the possibility of two messages is still required. I thought explicitly identifying complete messages using the COBS separator was more robust than just trying to deserialise any returned bytes.

HaoboGu:
> The main problem is that the serial Read trait isn't guaranteed to read all the bytes in a message

I think it is guaranteed. The buffer of `BufferredUart` is set to `SPLIT_MESSAGE_MAX_SIZE`. Could you give me a repo or a working example so that I can repro the issue?

grabpot:
I'm not sure, maybe I misunderstood, but:
1. The [trait](https://docs.rs/embedded-io-async/latest/embedded_io_async/trait.Read.html) says "it is not guaranteed that all available bytes are returned". Even if embassy_rp::BufferedUart did guarantee it, I'm not sure it's a given other drivers, e.g. for STM32, that implement the trait do too.
2. When I asked in the Embassy Matrix a related question about the embassy_hal_internal::RingBuffer behaviour, dirbaio referred to the trait and replied "in general when reading stuff you have to write a loop, you can't rely you'll receive the entire "message" in a single read call."
3. I'm not sure the size of a postcard serialized message is even fixed. Anything larger than a u8 (admittedly not used today) is encoded as a [Varint](https://docs.rs/postcard/latest/postcard/index.html#variable-length-data). Otherwise, RMK could use read_exact() from the trait to read the required number of bytes.
4. The `SPLIT_MESSAGE_MAX_SIZE` is set to SplitMessage::POSTCARD_MAX_SIZE __+ 4__, so it's big enough to fit two messages (or was, at least until 0.4.4, I didn't check the new conection state messages)

For an example, I wrote my own half-duplex UART driver as the Sofle keyboard isn't wired for full-duplex UART, and this doesn't (and can't see 3. above) return all the bytes of a message on read(): https://github.com/grabpot/rmk-sofle-dvorak

rmk/src/split/driver.rs
```diff
@@ -19,7 +20,7 @@ pub(crate) enum SplitDriverError {
 
 /// Split message reader from other split devices
 pub(crate) trait SplitReader {
-    async fn read(&mut self) -> Result<SplitMessage, SplitDriverError>;
+    async fn read(&mut self) -> Result<Vec<SplitMessage, 2>, SplitDriverError>;
```
HaoboGu: For the trait, each `read` call should read only one `SplitMessage`, not two. It defines "expected behavior", should not be affected by the serial implementation

rmk/src/split/serial/mod.rs
```diff
@@ -42,44 +43,90 @@ pub(crate) async fn run_serial_peripheral_monitor<
...
         }
-        let message: SplitMessage = postcard::from_bytes(&buf).map_err(|e| {
-            error!("Postcard deserialize split message error: {}", e);
-            SplitDriverError::DeserializeError
-        })?;
-        Ok(message)
+
+        let mut start_byte = 0;
+        let mut end_byte = start_byte;
+        while end_byte < self.n_bytes_part {
```
HaoboGu: The buffer size is `SPLIT_MESSAGE_MAX_SIZE`, so this second loop is not needed -- we can assume that after the first `while` loop, there's a complete `SplitMessage` received.

rmk/src/split/serial/mod.rs
```diff
@@ -42,44 +43,90 @@ pub(crate) async fn run_serial_peripheral_monitor<
...
+                .serial
+                .read(&mut self.buffer[self.n_bytes_part..])
+                .await
+                .inspect_err(|_e| self.n_bytes_part = 0)
+                .map_err(|_e| SplitDriverError::SerialError)?;
+            if n_bytes == 0 {
+                return Err(SplitDriverError::EmptyMessage);
+            }
+
+            self.n_bytes_part += n_bytes;
```
HaoboGu: Check if `n_bytes_part > self.buffer.len()`

grabpot:
__This is WIP and still needs to be tested.__

* Reverted SplitReader::read() to return a single message, which removes the need for a second loop in read(), and now uses postcard::take_from_bytes_cobs() to just extract a single message and store the unused bytes.
* Added min/max for buffer index < 0 and/or > buffer.len()
* SPLIT_MESSAGE_MAX_SIZE reverted to SplitMessage::POSTCARD_MAX_SIZE + 4; should be large enough to fit single message with COBS encoding but not second message.
* Tidied up serial.read() error handling to remove inspect_err()

rmk/src/split/serial/mod.rs
```diff
@@ -42,44 +42,81 @@ pub(crate) async fn run_serial_peripheral_monitor<
...
+                        error!("Postcard deserialize split message error: {}", e);
+                        self.n_bytes_part = 0;
+                        SplitDriverError::SerializeError
+                    },
+                )?;
+            (message, unused_bytes.len())
+        };
+
+        self.buffer.copy_within(
+            (self.n_bytes_part - n_bytes_unused).max(0)..self.n_bytes_part,
```
HaoboGu: max(0) won't work because both values are unsigned. 

Will n_bytes_parts always be larger than n_bytes_used?

grabpot: Thanks, good point. Will change to saturating_sub().

Yes, I believe so, or at least larger or equal to n_bytes_used. We pass a slice of length n_bytes_part to postcard::take_from_bytes_cobs(), and it splits it with split_at_mut() so unused_bytes.len() should always be smaller or equal to n_bytes_parts. I think the most common case is probably n_bytes_unused is zero, then copy_within is given a zero length range and does nothing.

HaoboGu: if so, a plain "-" should be enough?

rmk/src/split/serial/mod.rs
```diff
@@ -42,44 +42,81 @@ pub(crate) async fn run_serial_peripheral_monitor<
...
-            error!("Postcard deserialize split message error: {}", e);
-            SplitDriverError::DeserializeError
-        })?;
+
+        let (message, n_bytes_unused) = {
+            let (message, unused_bytes): (SplitMessage, &mut [u8]) =
+                postcard::take_from_bytes_cobs(&mut self.buffer[..self.n_bytes_part]).map_err(
+                    |e| {
+                        error!("Postcard deserialize split message error: {}", e);
+                        self.n_bytes_part = 0;
```
HaoboGu: Do we need to clear the buffer when there's deserialization error?

rmk/src/split/serial/mod.rs
```diff
@@ -42,44 +42,81 @@ pub(crate) async fn run_serial_peripheral_monitor<
...
-            .await
-            .map_err(|_e| SplitDriverError::SerialError)
+        let mut remaining_bytes = bytes.len();
+        while remaining_bytes > 0 {
+            let sent_bytes = self
+                .serial
+                .write(&bytes[bytes.len() - remaining_bytes..])
+                .await
+                .map_err(|_e| SplitDriverError::SerialError)?;
+            remaining_bytes -= sent_bytes;
```
HaoboGu, line 117: Is it possible that `sent_bytes` > `remaining_bytes`? If so, `remaining_bytes` could be incorrect. If not, it's safe.

HaoboGu: Looked at it again, I think it's safe.

rmk/src/split/serial/mod.rs
grabpot: No, any bytes after the zero terminator byte might still be valid. take_from_bytes_cobs() won't return unused_bytes if there's an error, but we could get the position of the terminator, then store the bytes after that?

HaoboGu: I'm not sure that when the `take_from_bytes_cobs` returns error, is the `SENTINEL` still a valid byte in the slice?

grabpot: no, it's not, take_from_bytes_cobs modifies the slice so we pass a clone of the buffer instead.

HaoboGu:
Look good now, @grabpot do you have any other thing about this PR?

HaoboGu:
Thanks very much!

PR #237, opened 2025-01-05
fix(scripts/via2rmk.py): added missing key and modifier names

whitelynx:
Found some missing keys and mods used in [the Svalboard Vial config](https://github.com/svalboard/vial-qmk/blob/vial/keyboards/svalboard/keymaps/vial/vial.json).


HaoboGu:
Thanks!

PR #243, opened 2025-01-07
Fix via2rmk Rust output

whitelynx:
Rust output was incorrect compared to examples; this should fix that, and correct/improve the help message a bit.


HaoboGu:
Thanks!

PR #245, opened 2025-01-10
feat(rmk): added matrix sample and scan delay config

whitelynx:
Allow configuring matrix input sample delay for slow input settle times, and scan delay for adjusting matrix polling rate.

Needed for Svalboard, since its optical sensors respond pretty slowly compared to a traditional key matrix.

Also, added `active_low` feature flag, which inverts output pin logic levels. (needed by Svalboard for its PFET electronics)


whitelynx:
Closing in favor of #246 - you can just implement a custom matrix to adjust timing.

whitelynx:
Closing in favor of #246 - you can just implement a custom matrix to adjust timing.

PR #254, opened 2025-02-06
Adapt BLE Advertisement Name

AdinAck:
Right now, running the default setup project for NRF52840 results in adv data which is too large due to the default keyboard name being too long. To avoid this, use `adapt_name` instead of `full_name`.

### DIFF
--- a/rmk/src/ble/nrf/advertise.rs
+++ b/rmk/src/ble/nrf/advertise.rs
@@ -1,5 +1,5 @@
 use nrf_softdevice::ble::advertisement_builder::{
-    AdvertisementDataType, Flag, LegacyAdvertisementBuilder, LegacyAdvertisementPayload,
+    AdvertisementDataType, Error, Flag, LegacyAdvertisementBuilder, LegacyAdvertisementPayload,
     ServiceList, ServiceUuid16,
 };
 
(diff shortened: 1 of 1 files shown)


HaoboGu:
When using `adapt_name`, will the displayed keyboard name be truncated if it's too long?

AdinAck:
Yep, according to the bluetooth spec this is the desired behavior. At the very least it will prevent the default project from panicing (since the builder is executed at runtime rather than compile time).

HaoboGu:
I didn't encounter any panic when building the adv data, did you use the built-in example?

AdinAck:
Haha, the name it chooses is the project name, clearly "keyboard-test-rmk" is too long.

HaoboGu:
I'm thinking about keeping using `full_name`, but reminding users to change their long project name at the compile time. 🤔

As an user, I would think it's a bug if the displayed name is not what I expect

AdinAck:
Sure, I personally also disagree with the notion of just truncating the name rather than coming up with a shorter name.

How about using `.try_build()` instead of `.build()` so you can catch the error and panic with a more useful message?

HaoboGu:
I prefer to have a compile-time error other than panic at the run-time. 

I think we can keep the current try_build way, but add an extra compile-time check as well?

AdinAck:
Well the current behavior is to panic at runtime. My changes only improve the error message. I agree the advertisement data should be built at compile time (like the scan data) but it is wrapped in a non-const function, I don't know enough about the system to know why that is. If you can make that function const, then the builder will naturally result in a compile time error.

HaoboGu:
cool! I'm merging this PR first, then I will add that const fn

HaoboGu:
thanks!

PR #256, opened 2025-02-15
Fixed issues that caused compilation to fail on esp32-s3 when using direct pins

B83C:
For the first commit, it was discovered that esp-idf-srv::hal::gpio::Output, in fact, does not implement embedded_hal::digital::OutputPin at all. The trait was only implemented for PinDriver.
For the second commit, the iterator returned more keymaps than was defined. The patch was hacky, but it works, review required.

### DIFF
--- a/rmk/src/storage/mod.rs
+++ b/rmk/src/storage/mod.rs
@@ -580,10 +580,12 @@ impl<F: AsyncNorFlash, const ROW: usize, const COL: usize, const NUM_LAYER: usiz
             {
                 match item {
                     StorageData::KeymapKey(key) => {
-                        assert!(key.layer < NUM_LAYER);
-                        assert!(key.row < ROW);
-                        assert!(key.col < COL);
-                        keymap[key.layer][key.row][key.col] = key.action;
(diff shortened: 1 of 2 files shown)


rmk/src/storage/mod.rs
```diff
@@ -580,10 +580,12 @@ impl<F: AsyncNorFlash, const ROW: usize, const COL: usize, const NUM_LAYER: usiz
             {
                 match item {
                     StorageData::KeymapKey(key) => {
-                        assert!(key.layer < NUM_LAYER);
-                        assert!(key.row < ROW);
-                        assert!(key.col < COL);
-                        keymap[key.layer][key.row][key.col] = key.action;
+                        if key.layer < NUM_LAYER && key.row < ROW && key.col < COL {
```
HaoboGu, line 583: those asserts can be removed after  the if statement is added

B83C: yes, forgot to :)

HaoboGu (approved):

HaoboGu:
thanks

PR #262, opened 2025-02-19
Update py32f07x example

decaday:
Hello! I’ve updated the `py32f07x` example’s dependency on py32-hal to version 3.0.

Due to limited RAM and some odd stack memory issues, I’ve split this PR into two commits for easier testing.

- The first commit does not add storage support but works fine.
- The second commit adds storage support but causes a stack overflow.

Enabling storage support required us to disable `defmt`.

It also caused the stack pointer (SP) to drop to 0x2000668 when entering the `main` function, which is already within the .bss section (and will cause the program to fail to run if using flip-link).

It increased stack memory usage by 9KB, which is abnormal.

Note: defmt is not completely disabled in the current commit, and can be used in the current main.rs to get the current stack position. (ROM is not overflow under this condition but RAM is still overflow)

If the issue cannot be resolved, it may be best to merge only the first commit (I’ll remove the second commit).


HaoboGu:
cooool! I'll test it soon

decaday:
This PR is out of date.
Closed because #397 is marged.

PR #265, opened 2025-02-22
Fix esp32 BLE serial number characteristic UUID

hcabel:
The serial number string for Esp32 BLE was stored in a characteristic with the wrong UUID 0x2a50 (PnP ID) instead of 0x2A25 (Serial Number String) [see here](https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/Assigned_Numbers/out/en/Assigned_Numbers.pdf?v=1740135081988) page 92.
Note: that later in the code your setting the PnP ID, essentially overriding whatever serial number was there anyway

### DIFF
--- a/rmk/src/ble/esp/server.rs
+++ b/rmk/src/ble/esp/server.rs
@@ -93,7 +93,7 @@ impl BleServer {
         block_on(server.get_service(BleUuid::from_uuid16(0x180a)))
             .unwrap()
             .lock()
-            .create_characteristic(BleUuid::from_uuid16(0x2a50), NimbleProperties::READ)
+            .create_characteristic(BleUuid::from_uuid16(0x2a25), NimbleProperties::READ)
             .lock()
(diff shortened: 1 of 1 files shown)


HaoboGu:
Good catch! It should be 0x2A25. And there's another wrong characteristic UUID below, would you mind fixing it as well? Thanks a lot

> Note: that later in the code your setting the PnP ID, essentially overriding whatever serial number was there anyway

IIRC PnP ID contains `vid_source`, `vender_id`, `product_id` and `product_version`, what do you mean that it will override the serial number?

hcabel:
🤦Haha Yeah I had missed the second one (doing it right now)

Never mind, wrong guess ^^ I though that setting the serial number to 0x2a50 then calling `hid.pnp(...` (also using 0x2a50) would override the previously set serial number, but no because one is set on the server and the other on the hid device

HaoboGu:
thanks！

PR #270, opened 2025-03-02
minor fixes

tib888:
Fix in mouse_report button release: &= 0b0111, like bit patterns were wrong (0b11110111 is correct), fixed in nicer format: !(1 << 3)

Likely fix in default_debouncer: I guess the elapsed_ms should be added not 1

small optimizations 

small typo fixes

### DIFF
--- a/rmk/src/keycode.rs
+++ b/rmk/src/keycode.rs
@@ -45,78 +45,34 @@ impl ModifierCombination {
             .with_ctrl(ctrl)
     }
 
-    /// Convert modifier combination to a list of modifier keycodes.
-    /// Returns a list of modifiers keycodes, and the length of the list.
-    pub(crate) fn to_modifier_keycodes(self) -> ([KeyCode; 8], usize) {
-        let mut keycodes = [KeyCode::No; 8];
-        let mut i = 0;
-        if self.right() {
-            if self.ctrl() {
-                keycodes[i] = KeyCode::RCtrl;
-                i += 1;
-            }
-            if self.shift() {
-                keycodes[i] = KeyCode::RShift;
-                i += 1;
-            }
-            if self.alt() {
-                keycodes[i] = KeyCode::RAlt;
-                i += 1;
-            }
-            if self.gui() {
-                keycodes[i] = KeyCode::RGui;
-                i += 1;
-            }
-        } else {
-            if self.ctrl() {
-                keycodes[i] = KeyCode::LCtrl;
-                i += 1;
-            }
-            if self.shift() {
-                keycodes[i] = KeyCode::LShift;
-                i += 1;
-            }
-            if self.alt() {
-                keycodes[i] = KeyCode::LAlt;
-                i += 1;
-            }
-            if self.gui() {
-                keycodes[i] = KeyCode::LGui;
-                i += 1;
-            }
-        }
-
-        (keycodes, i)
-    }
-
     /// Get modifier hid report bits from modifier combination
-    pub(crate) fn to_hid_modifier_bits(self) -> u8 {    
+    pub(crate) fn to_hid_modifier_bits(self) -> u8 {
         let mut hid_modifier_bits = 0;
-        if self.right() {
+        if !self.right() {
             if self.ctrl() {
-                hid_modifier_bits |= 1 << 4;
+                hid_modifier_bits |= HidModifierBit::RCtrl as u8;
             }
             if self.shift() {
-                hid_modifier_bits |= 1 << 5;
+                hid_modifier_bits |= HidModifierBit::RShift as u8;
             }
             if self.alt() {
-                hid_modifier_bits |= 1 << 6;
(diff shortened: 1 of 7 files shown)


rmk/src/keycode.rs
```diff
@@ -90,13 +90,35 @@ impl ModifierCombination {
     }
 
     /// Get modifier hid report bits from modifier combination
-    pub(crate) fn to_hid_modifier_bits(self) -> u8 {
-        let (keycodes, n) = self.to_modifier_keycodes();
+    pub(crate) fn to_hid_modifier_bits(self) -> u8 {    
         let mut hid_modifier_bits = 0;
-        for item in keycodes.iter().take(n) {
-            hid_modifier_bits |= item.as_modifier_bit();
+        if self.right() {
```
HaoboGu: This seems to duplicate the logic in `to_modifier_keycodes`?

tib888: Yes, but to_modifier_keycodes() in not used anymore (to_hid_modifier_bits was its only user) , so it could be deleted, and keep the direct implementation in to_hid_modifier_bits...

HaoboGu: OK, can you remove it. And I prefer to use `KeyCode::RCtrl.as_modifier_bit()` instead of plain `1 << 4`

tib888: Well, to be ownest, would rather delete KeyCode::RCtrl.as_modifier_bit() too since that is not used elsewhere either.
And more importantly, and its implementation is a hack, which relies on some magic coincidence: the order of these modifier bits is the same as they are ordered in the KeyCode enum:
    LCtrl = 0x00E0,
    LShift = 0x00E1,
    LAlt = 0x00E2,
    LGui = 0x00E3,
    RCtrl = 0x00E4,
    RShift = 0x00E5,
    RAlt = 0x00E6,
    RGui = 0x00E7,

    /// Returns the byte with the bit corresponding to the USB HID
    /// modifier bitfield set.
    pub(crate) fn as_modifier_bit(self) -> u8 {
        if self.is_modifier() {
            1 << (self as u16 as u8 - KeyCode::LCtrl as u16 as u8)
        } else {
            0
        }
    }

HaoboGu: ah, makes sense. Could you please remove both?

tib888: Sorry I remembered wrongly, they have some usage places, so they can not be deleted.

Like this code fragment, which used several places, so should be extracted into a function (sends every modifier press in a separate hid report):
```
            let (keycodes, n) = modifier.to_modifier_keycodes();
            for kc in keycodes.iter().take(n) {
                self.process_action_keycode(*kc, key_event).await;
            }
```

This function could report all modifier presses/releases in a single report:
```
    async fn process_modifiers(&mut self, pressed: bool, modifiers: ModifierCombination) {
        if pressed {
            self.register_modifier(modifiers.to_hid_modifier_bits());
        } else {
            self.unregister_modifier(modifiers.to_hid_modifier_bits());
        }
        self.send_keyboard_report().await;
    }
```

Or in my opinion it is not necessary to send a separate HID report just because some modifiers were pressed
'together' with other keys. So even this could be enough (see MacroOperation::Text extraction as an example)) for those cases:
```    
    fn decorate_with_modifiers(&mut self, pressed: bool, modifiers: ModifierCombination) {
        if pressed {
            self.register_modifier(modifiers.to_hid_modifier_bits());
        } else {
            self.unregister_modifier(modifiers.to_hid_modifier_bits());
        }        
    }
```

(After switching to this new function "to_modifier_keycodes()" could be deleted.)

What is your opinion?

tib888: Also I am not happy that ModifierCombination has this "right" flag just to fit in 5 bits, instead of having all 8 bits for each keys.

Not to mention that I would use a stack to push the active modifier combination when a new combination is applied temporarily by Action::Modifier, and would pop it back after the action finished.
(Same goes for macro executions, which at the beginning likely should clear all modifiers. And of course shoud be framed with a modifier+lock state push/pop.)

HaoboGu: I agree. Sending modifiers in separate HID reports is unnecessary. Please fix it :D

As for the 5 bit `ModifierCombination`, it was designed this way because the `KeyAction` is limited to 16bits, see:

https://github.com/HaoboGu/rmk/blob/01b5d5604a99c9108859d9cb2bf8d013b0fa385e/rmk/src/action.rs#L38-L45

But after porting RMK to vial, this design has not actually been used. I prefer to keep it for now, until we have the chance to review the `Action` and `KeyAction`.

tib888: Please accept this pull request as is. 
I have done the rest of the fixes we have talked about, but since it is a more risky refactor, I'd like to keep it separate.

HaoboGu:
LGTM, thanks! pls just ignore the ci error

Issue #253, opened 2025-02-04
Bootloader key


HaoboGu:
Provide a way to jump to MCU's bootloader.

Reference:

- qmk bootmagic: https://docs.qmk.fm/features/bootmagic
- zmk: https://zmk.dev/docs/keymaps/behaviors/reset
- embassy-boot-(rp/nrf/stm32): https://github.com/embassy-rs/embassy
- embassy-rp reboot example: https://github.com/embassy-rs/embassy/blob/d3059bbcf7fcf5acd84a8f672c0d0073e323f73b/examples/rp/src/bin/usb_serial_with_handler.rs#L33

PR #271, opened 2025-03-03
feat(boot): jump to Adafruit Nrf52 bootloader

drindr:
Add support for Bootloader key (only non-split or central in split support now)

configure in vial: 
- RESET for reset into bootloader
- 0x7c51 for normal reset

related issue: #253

### DIFF
--- a/rmk/Cargo.toml
+++ b/rmk/Cargo.toml
@@ -199,4 +199,4 @@ rp2040_pio = [
     "dep:fixed",
 ]
 
-adafruit = ["_nrf_ble"]
+adafruit_bl = ["_nrf_ble"]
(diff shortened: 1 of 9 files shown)


rmk/Cargo.toml
```diff
@@ -196,5 +196,7 @@ rp2040_pio = [
     "dep:rp-pac",
     "dep:pio",
     "dep:pio-proc",
-    "dep:fixed"
+    "dep:fixed",
 ]
+
+adafruit = ["_nrf_ble"]
```
HaoboGu: change the feature name to `adafruit_bl`?

HaoboGu:
cool, thanks!

PR #283, opened 2025-03-15
removed pio-proc, as it is an internal use crate.

gabevenberg:
Using pio directly instead. Compiles, and works on the central, still haven't gotten the peripheral working.


HaoboGu:
lgtm, thanks!

PR #286, opened 2025-03-16
hard coded constants replaced to use COMBO_MAX_LENGTH

tib888:
hard coded constants replaced to use COMBO_MAX_LENGTH
+ some minor fixes

will be useful if the combo MAX parameters will be part of the config file

### DIFF
--- a/rmk/src/via/vial.rs
+++ b/rmk/src/via/vial.rs
@@ -6,7 +6,7 @@ use num_enum::FromPrimitive;
 use crate::{
     action::KeyAction,
     channel::FLASH_CHANNEL,
-    combo::{Combo, COMBO_MAX_NUM},
+    combo::{Combo, COMBO_MAX_LENGTH, COMBO_MAX_NUM},
     keymap::KeyMap,
     storage::{ComboData, FlashOperationMessage},
     usb::descriptor::ViaReport,
@@ -55,7 +55,7 @@ const VIAL_PROTOCOL_VERSION: u32 = 6;
 const VIAL_EP_SIZE: usize = 32;
 const VIAL_COMBO_MAX_LENGTH: usize = 4;
 
-/// Note: vial uses litte endian, while via uses big endian
+/// Note: vial uses little endian, while via uses big endian
 pub(crate) async fn process_vial<const ROW: usize, const COL: usize, const NUM_LAYER: usize>(
     report: &mut ViaReport,
     vial_keyboard_Id: &[u8],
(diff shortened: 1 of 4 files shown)


HaoboGu:
VIAL supports only COMBO_MAX_LENGTH <= 4, so it's better to do some processing when vial reading combos

HaoboGu:
https://github.com/HaoboGu/rmk/blob/111461052b421ce94ccd4731f94fdc35487d8ebf/rmk/src/via/vial.rs#L262-L269

Currently only combos with length <=4 will be loaded, that can be confusing if a longer combo is set.

tib888:
I added protection for the cases when COMBO_MAX_LENGTH < VIAL_COMBO_MAX_LENGTH.
In this case the combo stored on the keyboard is simply not overwritten by Vial.

Of course it would be nice to handle better the buffer size mismatches - for example return error to vial, but i do not know the protocol...

The simplest solution is likely to keep COMBO_MAX_LENGTH == VIAL_COMBO_MAX_LENGTH == 4 and do not expose this parameter in keyboard config as it is quite reasonable number already.

However COMBO_MAX_NUM should be exposed in the config with a practical maximum limit: 255 - so vial will be able to acces them all.

HaoboGu:
> Of course it would be nice to handle better the buffer size mismatches - for example return error to vial, but i do not know the protocol...

Vial doesn't have an error handling mechanism, makes it quite hard.

> The simplest solution is likely to keep COMBO_MAX_LENGTH == VIAL_COMBO_MAX_LENGTH == 4 and do not expose this parameter in keyboard config as it is quite reasonable number already.

I agree.

> However COMBO_MAX_NUM should be exposed in the config with a practical maximum limit: 255 - so vial will be able to acces them all.

Making `COMBO_MAX_NUM` configurable is important. Especially for MCUs which doesn't have too much RAM. I'm going to do it soon

HaoboGu:
> Do you consider this pull request as acceptable?

Did you test how VIAL reads the combos with length > 4? I see in current implementation it just returns all data of the combo.

And, then `VIAL_COMBO_MAX_LENGTH` is large, the `input_data` might panic.

rmk/src/via/vial.rs
```diff
@@ -136,18 +136,19 @@ pub(crate) async fn process_vial<const ROW: usize, const COL: usize, const NUM_L
...
                         }
                         LittleEndian::write_u16(
-                            &mut report.input_data[9..11],
+                            &mut report.input_data
+                                [1 + VIAL_COMBO_MAX_LENGTH * 2..3 + VIAL_COMBO_MAX_LENGTH * 2],
                             to_via_keycode(combo.output),
                         );
                     } else {
-                        report.input_data[1..11].fill(0);
+                        report.input_data[1..3 + VIAL_COMBO_MAX_LENGTH * 2].fill(0);
```
HaoboGu, line 151: It will panic when `VIAL_COMBO_MAX_LENGTH` >= 15

tib888: I have checked its source code:
https://github.com/vial-kb/vial-gui/blob/main/src/main/python/protocol/combo.py

We can say that VIAL_COMBO_MAX_LENGTH is hard coded currently to 4.

My intent was not to change VIAL_COMBO_MAX_LENGTH or COMBO_MAX_LENGTH anyway, but to use these constants at the appropriate places instead of magic numbers.

HaoboGu: Thanks for the explanation!

HaoboGu (approved):

HaoboGu:
Thanks a lot!

PR #287, opened 2025-03-16
feat/joystick: add support for joystick

drindr:
The `NrfAdc` is a `input_device` sampling analog signal with NRF's SAADC.

The `JoystickProcessor` process the analog signal and map them into the mouse report with some parameters.

Usage (deprecated. see comments blow):
```rust
let saadc = init_adc([p.P0_31.into(), p.P0_29.into()], p.SAADC);
saadc.calibrate().await;
let mut adc_dev = NrfAdc::new(saadc, 20);
let mut joy_proc =
    JoystickProcessor::new([0, 1]/* ADC channel id */, [[40, 0], [0, 40]]/* transform */, [30930, 31087]/* bias */, 8/* resolution */, &keymap);
```
full configuration: https://github.com/drindr/rmk-config/blob/958585736a1f49630f9aa0e780f002fa9d42bbbc/src/central.rs#L140-L144

TODO;
- [x] refactor the battery config with the `input_device`
- [x] an easy-to-use TOML configuration method
- [x] a detailed document because the joystick requires lots of parameters to configure
- [ ] configure parameters through `rmk-gui`


drindr:
```rust
let saadc = init_adc(
    [saadc::VddhDiv5Input.into(), p.P0_31.into(), p.P0_29.into()],
    p.SAADC,
);
saadc.calibrate().await;
let mut adc_dev = NrfAdc::new(saadc, [EventType::Battery, EventType::Joystick(2)], 20);
let mut batt_proc = BatteryProcessor::new(1, 5, &keymap);
let mut joy_proc = JoystickProcessor::new([[80, 0], [0, 80]], [29130, 29365], 6, &keymap);
...
run_devices! (
    (matrix, adc_dev) => EVENT_CHANNEL,
),
run_processor_chain! {
    EVENT_CHANNEL => [joy_proc, batt_proc],
}
...
```

To adapt for the more concise `InputProcessor` API in the plan, modify the implementation of `NrfAdc`

drindr:
@HaoboGu
I modify the `rmk-macro` so that  `input_device` can be easily configured in the TOML config.
To make the definitions in `entry.rs` more consistence, I modified some implements.
Most of them is about the `processor` and the `battery`.
But I do not have some keyboards with some specific chips in the examples, so could you help me to test them?

As for the joystick's configuration, it is look like: 
```toml
[[split.central.input_device.joystick]]
name = "default"
pin_x = "P0_31"
pin_y = "P0_29"
pin_z = "_"
transform = [[80, 0], [0, 80]]
bias = [29130, 29365]
resolution = 6
# func = "mouse | n-direction key" # TODO: only mouse is supported now
```
The full configuration for keyboard with joystick:
https://github.com/drindr/rmk-config/blob/d7044c9d8edad984a4f3cf4e1a96d5dd8623c420/keyboard.toml#L297-L304

HaoboGu:
Sure, I'll test it this weekend

HaoboGu:
Did you write any documentation about configuring it in the `keyboard.toml`?

rmk-macro/src/keyboard_config.rs
```diff
@@ -89,14 +90,19 @@ pub(crate) struct KeyboardConfig {
 
 #[derive(Clone, Debug)]
 pub(crate) enum BoardConfig {
-    Normal(MatrixConfig),
     Split(SplitConfig),
-    DirectPin(MatrixConfig),
+    Single(SingleConfig),
```
HaoboGu: `Single` -> `Unibody`

drindr:
> Did you write any documentation about configuring it in the `keyboard.toml`?

Not yet. I will complete it as soon as possible.

rmk-macro/src/input_device/mod.rs
```diff
@@ -0,0 +1,54 @@
...
+
+    // generate ADC configuration
+    let ble_config = match &keyboard_config.communication {
+        CommunicationConfig::Ble(ble_config) | CommunicationConfig::Both(_, ble_config) => {
+            Some(ble_config.clone())
+        }
+        _ => None,
+    };
+    let (adc_config, adc_processors) = match &keyboard_config.board {
+        BoardConfig::UniBody(UniBodyConfig { input_device, .. }) => expand_adc_device(
```
HaoboGu, line 27: The `expand_input_device_config` should follow the structure of `InputDeviceConfig`:

```
pub struct InputDeviceConfig {
    pub encoder: Option<Vec<EncoderConfig>>,
    pub pointing: Option<Vec<PointingDeviceConfig>>,
    pub joystick: Option<Vec<JoystickConfig>>,
}
```

aka

```
expand_input_device_config:

- expand_encoder_config:
    - for encoder in encoders: 
        - expanded code 
- expand_joystick_config:
    - for joystick in joysticks: 
        - expanded code 
- expand_pointing_config:

...
```

I think this pattern is more clear

rmk/src/input_device/adc/nrf.rs
```diff
@@ -0,0 +1,89 @@
...
+            buf_state: 0,
+        }
+    }
+}
+
+impl<'a, const PIN_NUM: usize, const EVENT_NUM: usize> InputDevice
+    for NrfAdc<'a, PIN_NUM, EVENT_NUM>
+{
+    async fn read_event(&mut self) -> Event {
+        embassy_time::Timer::after_millis(self.polling_interval as u64).await;
```
HaoboGu: Is there any way to trigger the adc read? Now the adc samples the values forever, which will drain the battery quickly. 

The polling interval for joystick and battery should not be same -- joystick needs much higher polling freq.

rmk-macro/src/input_device/mod.rs
drindr: I chose to use `adc` rather than `joystick` because it processes the battery stuff as well.

HaoboGu: No problem for `expand_adc_config` containing both battery and joystick, and possibly others

rmk/src/input_device/adc/nrf.rs
drindr: Maybe recording the value, only accelerate the polling rate when the change of the value is larger than a threshold, like the sleep state in the key scan.

When there is no joystick device, the polling interval can be just set as a big number.

As for dividing the sampling of battery and joystick, I have 2 ideas.
The first one is sending the `p.SAADC` to the ADC device and reinitialize the `Saadc` instance before each sample.
The other is to modify the register directly before each sample because it seems the `embassy_nrf` do not provide API to change the configuration after `Saadc` initialized.

HaoboGu: I prefer to not poll the ADC at all, wait until the joystick moves. The joystick device could use the corresponding GPIOs as the interrupt source. If any edge is detected, drop GPIOs and create SAADC instance for reading adc values. Is it possible?

drindr: It will introduce a larger dead zone while it is inactivated, but on my small analog stick it seems acceptable. 🤔

HaoboGu: yeah, that's just trade-off. Switching polling rate is also acceptable. Maybe we need two modes, let user's decide. 

(sorry for increasing the workload😛, if you don't want to do it, just leave it as is

drindr: I will implement the switching polling rate method as sleep mode firstly.
Edge detection will be implemented as the waking for deep sleep mode in the future. Because to activate from the edge detection sometimes require moving the joystick backward due to the unknown current state, and I found the behavior of reusing pin in `Nrf` is different from my expectation.

drindr:
I think it is ready to merge as the first edition of the joystick. And I will add a feature about calibration in `rmk-gui` through extending `vial` commands later.

HaoboGu (approved):

HaoboGu:
Nice work, thanks!

PR #291, opened 2025-03-17
added artificial pull up resistor to pio tx line.

gabevenberg:
The rp2040 has weak pull up resistors, from 80k to 50k. This does not provide enough current to provide fast rise times at high baud rates with any moderately high capacitance, even as little capacitance as can be found with long traces, a few vias, or a longer TRRS cable. The solution is to also drive the line high at a weak drive current from the reciving side, providing plenty of current to drive the line high quickly while still being weak enough to be driven low from the tx side.

### DIFF
--- a/rmk/Cargo.toml
+++ b/rmk/Cargo.toml
@@ -96,7 +96,7 @@ features = ["split"]
 cortex-m = { version = "0.7" }
 
 [features]
-default = ["col2row", "split", "rp2040_pio", "rp2040_bl", "defmt"]
+default = ["col2row", "defmt"]
 
 ## If your PCB diode's direction is col2row, enable this feature. If it's row2col, disable this feature by `default-features = false`.
 col2row = []
--- a/rmk/Cargo.toml
+++ b/rmk/Cargo.toml
@@ -96,7 +96,7 @@ features = ["split"]
 cortex-m = { version = "0.7" }
 
 [features]
-default = ["col2row", "defmt"]
+default = ["col2row", "split", "rp2040_pio", "rp2040_bl", "defmt"]
 
 ## If your PCB diode's direction is col2row, enable this feature. If it's row2col, disable this feature by `default-features = false`.
 col2row = []
--- a/rmk/src/split/rp/uart.rs
+++ b/rmk/src/split/rp/uart.rs
@@ -213,7 +213,6 @@ impl<'a, PIO: Instance + UartPioAccess> BufferedUart<'a, PIO> {
             pin.set_schmitt(true);
             pin.set_pull(Pull::Up);
             pin.set_slew_rate(SlewRate::Fast);
-            pin.set_drive_strength(Drive::_12mA);
         }
     }
 
@@ -241,9 +240,8 @@ impl<'a, PIO: Instance + UartPioAccess> BufferedUart<'a, PIO> {
         cfg.fifo_join = FifoJoin::TxOnly;
         self.sm_tx.set_config(&cfg);
 
-        self.set_pin_tx();
-
         if self.full_duplex {
+            self.set_pin_tx();
             self.sm_tx.set_enable(true);
         }
     }
@@ -326,8 +324,9 @@ impl<'a, PIO: Instance + UartPioAccess> BufferedUart<'a, PIO> {
         // OEOVER set to INVERT, Direction::Out inverted to Direction:In
         self.sm_tx.set_pins(Level::Low, &[&self.pin_rx]);
 
+        let pin_tx = self.pin_tx.as_mut().unwrap_or(&mut self.pin_rx);
         // unset our fake-pull-up trickery
-        self.pin_rx.set_drive_strength(Drive::_12mA);
+        pin_tx.set_drive_strength(Drive::_12mA);
     }
 
     fn set_pin_rx(&mut self){
--- a/rmk/src/split/rp/uart.rs
+++ b/rmk/src/split/rp/uart.rs
@@ -322,12 +322,12 @@ impl<'a, PIO: Instance + UartPioAccess> BufferedUart<'a, PIO> {
     }
 
     fn set_pin_tx(&mut self){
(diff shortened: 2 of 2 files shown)


rmk/Cargo.toml
```diff
@@ -96,7 +96,7 @@ features = ["split"]
 cortex-m = { version = "0.7" }
 
 [features]
-default = ["col2row", "defmt"]
+default = ["col2row", "split", "rp2040_pio", "rp2040_bl", "defmt"]
```
gabevenberg: Temp so rust-analyzer includes the files Im working on, will revert before merge.

rmk/src/split/rp/uart.rs
```diff
@@ -290,9 +288,7 @@ impl<'a, PIO: Instance + UartPioAccess> BufferedUart<'a, PIO> {
         cfg.fifo_join = FifoJoin::RxOnly;
         self.sm_rx.set_config(&cfg);
 
-        // OEOVER set to INVERT, Direction::Out inverted to Direction:In
-        self.sm_rx.set_pin_dirs(Direction::Out, &[&self.pin_rx]);
-
+        self.set_pin_rx();
```
gabevenberg, line 289: Not sure about this one, set_pin_rx sets direction to `In`. Im not 100% sure what exactly pindirs does.

grabpot: set_pin_dirs() sets the pin to input or output mode - in input mode, the pin is not driven directly but with the pull-up, it's high, and in output mode, the pin drives the voltage, in this case to high on the rx side.

Without augmenting the pull-ups, the idea is both sides are open-drain and the tx side pulls the line low. Inverting the pin direction is a trick to allow the tx state machine to toggle output mode and pull the line low to transmit a 0 bit (by default, pin direction 0 would be input, not output).

The idea here is to set the rx side to drive the pin high, i.e. output mode, presumably because the built-in pull-ups are too weak.

rmk/src/split/rp/uart.rs
```diff
@@ -241,9 +241,7 @@ impl<'a, PIO: Instance + UartPioAccess> BufferedUart<'a, PIO> {
         cfg.fifo_join = FifoJoin::TxOnly;
         self.sm_tx.set_config(&cfg);
 
-        // OEOVER set to INVERT, Direction::Out inverted to Direction:In
-        self.sm_tx.set_pin_dirs(Direction::Out, &[pin_tx]);
-        self.sm_tx.set_pins(Level::Low, &[pin_tx]);
+        self.set_pin_tx();
```
grabpot: I think maybe this is only necessary if in full-duplex, i.e. inside the following if statement? In half-duplex, the same pin is later set to rx mode so this is unnecessary.

rmk/src/split/rp/uart.rs
```diff
@@ -315,12 +312,38 @@ impl<'a, PIO: Instance + UartPioAccess> BufferedUart<'a, PIO> {
...
         self.sm_rx.set_enable(true);
     }
 
+    fn set_pin_tx(&mut self){
+        self.sm_tx.set_pin_dirs(Direction::Out, &[&self.pin_rx]);
+        // OEOVER set to INVERT, Direction::Out inverted to Direction:In
+        self.sm_tx.set_pins(Level::Low, &[&self.pin_rx]);
+
+        // unset our fake-pull-up trickery
+        self.pin_rx.set_drive_strength(Drive::_12mA);
```
grabpot: In full-duplex, this should be self.pin_tx, and in half-duplex, self.pin_rx. self.pin_tx is an Option, and None in half-duplex, so I think we can unwrap it something like this:

let pin_tx = self.pin_tx.unwrap_or(self.pin_rx);

rmk/src/split/rp/uart.rs
grabpot, line 1: pin.set_drive_strength(Drive::_12mA); on Line 216 is probably unnecessary now as we're setting it separately for rx/tx pins.

tib888: I agree: on RX side 2mA is enough for any baud rate. 
The TX side hast to switch this to 4mA (or more) for the time of transition, then in full duplex mode switch back to 2mA for RX operation.

gabevenberg: > I agree: on RX side 2mA is enough for any baud rate. The TX side hast to switch this to 4mA (or more) for the time of transition, then in full duplex mode switch back to 2mA for RX operation.

Ive been following qmks [solution](https://github.com/qmk/qmk_firmware/blob/6d0e5728aa61b442885d48caf49d29e5c60e8197/platforms/chibios/drivers/vendor/RP/RP2040/serial_vendor.c#L160) closely, which uses 2mA for rx, and the full 12mA for tx.

rmk/src/split/rp/uart.rs
gabevenberg: Ok, thanks for the explination. Its still a bit weird that it can be in output mode while still being an input pin.

gabevenberg:
I have some HW coming to test this without the risk of frying my expensive keyboard controllers, so please dont merge until I can properly test that this wont burn out any gpios.

gabevenberg:
> I have some HW coming to test this without the risk of frying my expensive keyboard controllers, so please dont merge until I can properly test that this wont burn out any gpios.

Tested on a breadboard with a 2 button 'keyboard'. No fried GPIOs, and UART kept chugging even with a 150pF cap on the line.

Finally tested on my keyboard, fully working, no controllers burnt!

grabpot (approved):
Looks good to me!

HaoboGu:
nice work!

mvirkkunen:
Is this mentioned somewhere in the documentation? I just spent like two hours debugging weird voltage levels on the half duplex serial port, because I had added pull-up and series protection resistors like I usually do... did not think that somebody would be using a GPIO as a strong pull-up resistor!

Issue #297, opened 2025-03-25
make the storage to be an optional feature


TuEmb:
The storage is consuming a lot of RAM, we can make it as an optional feature to use for some MCUs with limited RAM.
My idea is:
  - we will add an feature named: `storage`
  - we can use this feature by updating Cargo.toml: `rmk = { version = "0.5.2", features = ["storage"]}`

@HaoboGu How do you think about it ?

HaoboGu:
Yes, I agree. Adding a `storage` feature is good for MCUs which don't have enough RAM.

PR #298, opened 2025-03-25
make the storage as an optional feature and enable by default

TuEmb:
Hi @HaoboGu,

As https://github.com/HaoboGu/rmk/issues/297, I just added `storage` feature in this PR. Please help me review it
I put it in `default` because I don't want to disrupt other examples that are using `storage`.
To remove the storage feature we just need to change `Cargo.toml`:
```
[dependencies]
rmk = { version = "0.5.2", default-features = false  }
```

Then, in the source, we can go with:

```Rust
    // Keyboard config
    let rmk_config = RmkConfig {
        usb_config: keyboard_usb_config,
        vial_config: VialConfig::new(VIAL_KEYBOARD_ID, VIAL_KEYBOARD_DEF),
        ..Default::default()
    };

    // Initialize the keymap
    let mut default_keymap = keymap::get_default_keymap();
    let keymap = initialize_keymap(
        &mut default_keymap,
        rmk_config.behavior_config.clone(),
    )
    .await;

    // Initialize the matrix + keyboard
    let debouncer = DefaultDebouncer::<ROW, COL>::new();
    let mut matrix = DirectPinMatrix::<_, _, ROW, COL, SIZE>::new(direct_pins, debouncer, true);
    let mut keyboard = Keyboard::new(&keymap, rmk_config.behavior_config.clone());

    // Initialize the light controller
    let light_controller: LightController<Output> =
        LightController::new(ControllerConfig::default().light_config);

    // Run RMK
    join3(
        keyboard.run(),
        run_rmk(&keymap, driver, light_controller, rmk_config),
        run_devices! (
            (matrix) => EVENT_CHANNEL,
        ),
    )
    .await;
```

### DIFF
--- a/rmk/src/keymap.rs
+++ b/rmk/src/keymap.rs
@@ -1,13 +1,14 @@
 use crate::{
     action::KeyAction,
-    boot::reboot_keyboard,
     combo::{Combo, COMBO_MAX_NUM},
     config::BehaviorConfig,
     event::KeyEvent,
     keyboard_macro::{MacroOperation, MACRO_SPACE_SIZE},
     keycode::KeyCode,
-    storage::Storage,
 };
+#[cfg(feature = "storage")]
+use crate::{boot::reboot_keyboard, storage::Storage};
+#[cfg(feature = "storage")]
 use embedded_storage_async::nor_flash::NorFlash;
 use num_enum::FromPrimitive;
 
@@ -57,7 +58,7 @@ impl<'a, const ROW: usize, const COL: usize, const NUM_LAYER: usize>
             behavior,
         }
     }
-
+    #[cfg(feature = "storage")]
     pub async fn new_from_storage<F: NorFlash>(
         action_map: &'a mut [[[KeyAction; COL]; ROW]; NUM_LAYER],
         storage: Option<&mut Storage<F, ROW, COL, NUM_LAYER>>,
--- a/rmk/src/via/mod.rs
+++ b/rmk/src/via/mod.rs
@@ -1,14 +1,14 @@
 use crate::{
     boot,
-    channel::FLASH_CHANNEL,
     config::VialConfig,
     hid::{HidError, HidReaderTrait, HidWriterTrait},
     keyboard_macro::{MACRO_SPACE_SIZE, NUM_MACRO},
     keymap::KeyMap,
-    storage::FlashOperationMessage,
     usb::descriptor::ViaReport,
     via::keycode_convert::{from_via_keycode, to_via_keycode},
 };
+#[cfg(feature = "storage")]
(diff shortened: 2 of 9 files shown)


rmk/src/via/mod.rs
```diff
@@ -254,7 +258,9 @@ impl<
                 // Then flush macros to storage
                 let num_zero = count_zeros(&self.keymap.borrow_mut().macro_cache[0..end as usize]);
                 if size < 28 || num_zero >= NUM_MACRO {
+                    #[cfg(feature = "storage")]
```
HaoboGu: This feature gate can be added to the if block:

```rust
#[cfg(feature = "storage")]
if size < 28 || num_zero >= NUM_MACRO {
...
}
```

rmk/src/via/mod.rs
```diff
@@ -131,7 +131,9 @@ impl<
                 match ViaKeyboardInfo::try_from_primitive(report.output_data[1]) {
                     Ok(v) => match v {
                         ViaKeyboardInfo::LayoutOptions => {
+                            #[cfg(feature = "storage")]
```
HaoboGu:
```
#[cfg(feature = "storage")]
ViaKeyboardInfo::LayoutOptions => {
...
}
```

rmk/src/keymap.rs
```diff
@@ -1,13 +1,14 @@
...
-    boot::reboot_keyboard,
     combo::{Combo, COMBO_MAX_NUM},
     config::BehaviorConfig,
     event::KeyEvent,
     keyboard_macro::{MacroOperation, MACRO_SPACE_SIZE},
     keycode::KeyCode,
-    storage::Storage,
 };
+#[cfg(feature = "storage")]
+use crate::{boot::reboot_keyboard, storage::Storage};
```
HaoboGu, line 10: Should `reboot_keyboard` gated by `storage`?

TuEmb: I found that the `reboot_keyboard` function is called when we got an error in storage initialization
image

rmk/src/via/mod.rs
TuEmb: Fixed in latest commit

rmk/src/via/mod.rs
TuEmb: Fixed in latest commit

rmk/src/keymap.rs
HaoboGu: OK! That's fine

gabevenberg:
Is it possible to disable the stroage feature while using keyboard.toml?

HaoboGu:
> Is it possible to disable the stroage feature while using keyboard.toml?

you can try set enable=false under  storage section

PR #299, opened 2025-03-25
feat(toml): add `KeyAction::Transparent` to keyboard.toml

### LINKED ISSUE #277: `KeyAction::Transparent` is not accessible through `keyboard.toml`
Basically the title. There is no way to configure a key as transparent through `keyboard.toml`

pcasotti:
Closes #277


HaoboGu (approved):
thanks

Issue #285, opened 2025-03-15
"state fork" behavior


tib888:
I'm planning to implement similar functionality to the "mod-morph" of ZMK or to the "fork" of kanata.
(It is also similar to "Key Overrides" in QMK, which would have the benefit to have VIAL support already, but too complex and missing the "else branch").

To goal is to be able to behave differently based on some state of the keyboard, most typically: which modifiers are active, but it would be nice if the lock-led states, layer states could be also used as condition (or the detected OS type if that would be available).

The needed parameters:
- keycode which triggers this behavior
- the list of state parameters, which are the base of the fork 
  (like: LShift, LCtrl, LAlt, LGui, RShift, RCtrl, RAlt, RGui, CapsLock, ScrollLock, NumLock, ...)
- the KeyAction to emmit when none of the listed states are active
- the KeyAction to emmit when any of the listed states are active

The simplest example use case:
```
[behavior.forks]
forks = [
  { trigger = ''BackSpace", conditions = ["LShift", "RShift"], none_output = "BackSpace", any_output = "Delete"  } #use Shift+BackSpace as Delete key
]
```

A complex use case: when "User0" KeyAction arrives, emmit "á" or "Á" depending on the shift/capslock state (on a US keybord on Windows, handle the case when the numlock is not pressed):
```
[behavior.forks]
forks = [
  { trigger = ''User0", conditions = ["LShift", "RShift", "CapsLock"], none_output = "User1", any_output = "User2"  } #capitalized letter decision
  { trigger = ''User1", conditions = ["NumLock"], none_output = "User3", any_output = "User4"  } #numlock missing decision
  { trigger = ''User2", conditions = ["NumLock"], none_output = "User5", any_output = "User6"  } #numlock missing decision
]

macros = [
  #on windows on the numpad type Alt+160 to emmit "á":
  { trigger ="User3",  [tap "NumLock", press "LAlt", tap "Kp1", tap "Kp6", tap "Kp0", release "LAlt", tap NumLock] },
  { trigger ="User4",  [press "LAlt", tap "Kp1", tap "Kp6", tap "Kp0", release "LAlt"] },
  #on windows on the numpad type Alt+181 to emmit "Á":
  { trigger ="User5",  [tap "NumLock", press "LAlt", tap "Kp1", tap "Kp8", tap "Kp1", release "LAlt", tap NumLock] },
[...]

PR #300, opened 2025-03-26
State based fork implementation

tib888:
The implementation of #285 is ready:
Similar functionality to the "mod-morph" of ZMK or to the "fork" of kanata, but more states can be tested: modifiers, mouse buttons, lock-leds* (documentation of keyboard config is included in PR).

Known problems:
- the led state based decisions do not work yet, since the received led states are not stored currently in the KeyboardReport.leds 

Please test, feedbacks are welcome!

Chaining many forks to test more complex conditions is coming soon...


tib888:
The implementation of chainable forks is ready too.
Please review/accept the PR!

rmk-macro/src/behavior.rs
```diff
@@ -117,18 +118,190 @@ fn expand_combos(combos: &Option<CombosConfig>) -> proc_macro2::TokenStream {
...
+    modifiers_right_alt: bool,
+    modifiers_right_gui: bool,
+
+    leds_num_lock: bool,
+    leds_caps_lock : bool,
+    leds_scroll_lock: bool,
+    //leds_compose: bool,
+    //leds_kana: bool,
+    //leds_power: bool,
+    //leds_shift: bool,
```
HaoboGu: remove the commented lines if they're not used?

HaoboGu:
quite a complicated feature, please give me some time reviewing it .. 🥲

before that, could you run `cargo fmt` under the `rmk` folder to make all code refactored? thanks

tib888:
It is cargo fmt already. mod.rs and rotary_encoder.rs needs formatting, but I haven't touched those. Should I format those too?

rmk-macro/src/behavior.rs
tib888: This part of StateBitsMacro goes to HidLeds, what is stored in keyboard.led_states: HidLeds, and sent to the PC in the KeyboardReport.

On the other hand, LightController receives HID reports with LedIndicator and based on that drives the led pins currently.

I think LightController **should also update** keyboard.led_states, but to do this, I'd like to ask your help.
(LedIndicator struct is the same as HidLeds, so only one of them should be kept). 

(If this would be ready, the forks would be able to change behavior based on these "lock leds", which would be very useful in my particular use-case...)

HaoboGu:
You don't have to. Great work! I'll take a deeper look asap

rmk-macro/src/behavior.rs
HaoboGu: For led states, it worths a global value which could be accessed by both `Keyboard` and `LightController`. A `Mutex` might be needed

rmk/src/keyboard.rs
```diff
@@ -71,6 +72,10 @@ impl<const ROW: usize, const COL: usize, const NUM_LAYER: usize, const NUM_ENCOD
     }
 }
 
+/// led states for the keyboard hid report (its value is also received in a hid report by the light service)
+/// LedIndicator type would be nicer, but that does not have const expr default constructor
+pub(crate) static mut LOCK_LED_STATES: u8 = 0; //no need for mutex since a single byte is atomic.
```
HaoboGu: Use `AtomicU8` to avoid `static mut` and unsafe code

rmk/src/storage/mod.rs
```diff
@@ -228,6 +239,34 @@ impl Value<'_> for StorageData {
                 );
                 Ok(3 + COMBO_MAX_LENGTH * 2)
             }
+            StorageData::ForkData(fork) => {
+                if buffer.len() < 13 {
```
HaoboGu, line 243: < 15?

HaoboGu (commented):
Looks good overall! Could you please add some unittest also?

rmk/src/storage/mod.rs
tib888: Fixed this.

tib888:
> Looks good overall! Could you please add some unittest also?

I added some tests.

I run them with:
cargo test --tests --no-default-features --verbose
And some of them seem frozen (run more than 60s).

Also when I ran test_fork_with_held_mouse_button() test, it failed at ln 1831, while the same thing was fine on ln1768.
When I tried to debug, it was also fine. When I added a delay, it was also fine, but there is nothing time related in the test.
So, there is likely some strange async stuff happening...

tib888:
I have traced down this freeze problem to:
KEYBOARD_REPORT_CHANNEL.sender().send(report).await

My guess is that the channel gets full if many test run. so this should be disabled during test runs somehow.

HaoboGu:
Yes you are correct. Running a dummy task which receives from `KEYBOARD_REPORT_CHANNEL` would solve this.

tib888:
I let you do this, because I'm not an expert in these.
The nicest would be to be able to test in the unit tests, what hid reports sent into this dummy host...

HaoboGu:
Sure, would you please add all the test just as is? I'll add the dummy host stuffs.

tib888:
> Sure, would you please add all the tests just as is? I'll add the dummy host stuffs.

Everything is already in the PR. Just uncomment the body of test_fork_with_held_modifier() (and test_fork_with_held_mouse_button())!

I do not want to push it uncommented because that leads to infinite test run on github.

HaoboGu:
thanks!

PR #301, opened 2025-03-27
Move to TrouBLE

### LINKED ISSUE #206: TrouBLE migration
Tracking Issue:

- [x] Blocked upstream by:  https://github.com/embassy-rs/trouble/issues/11
WIP PR: https://github.com/embassy-rs/trouble/pull/194

There's also an RMK branch for this at: https://github.com/HaoboGu/rmk/tree/feat/trouble

HaoboGu:
[TrouBLE](https://github.com/embassy-rs/trouble) is a BLE stack written in Rust, with a future goal of qualification. Moving to TrouBLE has lots of benefits, the most important are:

1. Now RMK uses `esp-idf-*` + esp32-nimble for ESP32. After moving to TrouBLE, we can use `esp-hal` with USB support. `esp-hal` is also the officially supported one by Espressif.
2. We can have BLE support for more MCUs with a single BLE implementation based on TrouBLE. We don't have to maintain different code for nRF/ESP anymore.

Closes #206


HaoboGu:
Need to find a way to build without nrf softdevice

HaoboGu:
It seems that TrouBLE cannot be built without nrf softdevice, since the new nrf softdevice controller is not provided as a separate binary, it's compiled into the firmware instead. But with some tweaking, the generated uf2 could be compatible with nice!nano's bootloader.

okhsunrog:
What do you think about using apache-nimble controller instead of SoftDevice here? There's an example of the usage in the trouble repo

HaoboGu:
> What do you think about using apache-nimble controller instead of SoftDevice here? There's an example of the usage in the trouble repo

Now RMK depends on trouble-host only and is controller agnostic, that means you can use apachr-nimble as the controller. It seems that apache-nimble supports nrf52840 only, so at least nrf-sdc example for nrf52832, etc will not be removed

jelmansouri:
There is a section in the read me that mentions the need to flash the the nrf softdevide. Not sure if this is still relevant. 
 https://github.com/HaoboGu/rmk/tree/feat/trouble/examples/use_rust/nrf52840_ble_split#with-debug-probe

examples/use_rust/nrf52840_ble_split/memory.x
```diff
@@ -1,13 +1,11 @@
 MEMORY
 {
   /* NOTE 1 K = 1 KiBi = 1024 bytes */
-  /* FLASH : ORIGIN = 0x00000000, LENGTH = 1024K */
-  /*RAM : ORIGIN = 0x20000000, LENGTH = 256K */
-
-  /* These values correspond to the NRF52840 with Softdevices S140 6.1.1 */
-  /* FLASH : ORIGIN = 0x00026000, LENGTH = 824K */
+  /* These values correspond to the nRF52840 WITH Adafruit nRF52 bootloader */
+  FLASH : ORIGIN = 0x00001000, LENGTH = 1020K
```
jelmansouri, line 5: If I understand this correctly the idea is to override the SoftDevice part of the of the Adafruit bootloader. Laaooking at the adafruit boot loader it seems to require around 40k?

HaoboGu: yes, you're correct. This value comes from https://devzone.nordicsemi.com/f/nordic-q-a/97314/uf2-image-and-bootloader. I'd check it again, ensuring it's correct

HaoboGu: Checked again. The bootloader size is 40KB, but it's placed at 0xF4000, according to 

https://github.com/adafruit/Adafruit_nRF52_Bootloader/blob/4329614a8f2a020b31af36a1f3fd843d357eb619/linker/nrf52840.ld#L8-L14

so it's correct.

gibbz00:
LGTM!

okhsunrog:
@HaoboGu I've just read the comments above and I'm a little bit confused. So we're overriding softdevice now? now it mean we don't need it anymore? I thought the best option would be to use softdefive 7.3.0 with suitable bootloader, I found this one in your repo: https://github.com/HaoboGu/rmk-ble-keyboard/blob/main/firmware/xiao_nrf52840_ble_bootloader-0.9.2_s140_7.3.0.hex
but now I'm a little bit confused. I'm new to nordic microcontrollers. Do I need to flash softdevice v6, v7, or I don't need to flash it at all, and how presence of bootloader changes it? can I flash any bootloader and then softdevice of needed version or I need to use bootloader with built-in softdevice? the documentation about this and memory maps of nrf chips would be very useful. I'm be grateful for explanation or some hints

HaoboGu:
@okhsunrog Using new BLE stack TrouBLE, we don't need the SoftDevice pre-flashed into the microcontroller anymore, the controller part is now merged into the generated firmware. So the version of softdevice is not important -- we will overwrite it. You can flash the bootloader with or without softdevice, both are fine.

okhsunrog:
@HaoboGu I thought Trouble provides only the host part, not controller. I thought the controller part is still from SoftDevice. Do I understand it in a wrong way?

HaoboGu:
@okhsunrog Yes, you're correct. 

The controller part is [nrf-sdc](https://github.com/alexmoon/nrf-sdc), which is a binding of new `softdevice controller`. `softdevice controller` the controller used in Nordic's new NCS SDK and it's different from old `softdevice` which contains both host and controller implementation. With nrf-sdc, the softdevice controller is compiled together into the firmware, that means we cannot flash the controller part and application separately.

okhsunrog:
@HaoboGu oh, now it's clear to me, thanks! The memory.x in the nrf52840 example in the repo suits for nice!nano with Adafruit_nRF52_bootloader, right?

HaoboGu:
@okhsunrog Yes. I made it compatible with nice!nanos

HaoboGu:
Let's get it merged!

Issue #317, opened 2025-04-08
bug: KeyCode::RepeatKey maps to th!(Again, Meh)


patmuk:
When configuring the keycap with k!(RepeatKey) it results wrongly in th!(Again, Meh) (as can be seen in via.

I assume the byte encoding goes wrong, as `RepeatKey = 0x779` and `Again = 0x079`. Probably combining `Meh` + `Again` results in `0x779` as well.

HaoboGu:
Yes you are correct. 0x779 is th!(Again, Meh) in vial keycode. 

<img width="660" alt="Image" src="https://github.com/user-attachments/assets/b9841a19-c088-4b7d-8d99-0c049e6d7d54" />

vial keycode for repeat key is 0x7C79

patmuk:
Indeed :) But where is this call missing?

HaoboGu:
https://github.com/HaoboGu/rmk/blob/596feb4935ff07b527baff787c49c02a302d02b9/rmk/src/via/keycode_convert.rs#L12-L24

Now if the key is Repeat(0x779), `to_via_keycode` does nothing on it, resulting in via keycode 0x779 aka `th!(Again, Meh)`. So adding conversion from `0x779` to `0x7C79` is enough:

```rust
 KeyAction::Single(a) => match a { 
     Action::Key(k) => { 
         if k.is_macro() { 
             k as u16 & 0xFF | 0x7700 
         } else if k.is_user() { 
             k as u16 & 0xF | 0x7E00 
         } else if k.is_combo() || k.is_boot() { 
             // is_rmk() 's subset 
             k as u16 & 0xFF | 0x7C00 
         } else if k as u16 == 0x779 {
             0x7C79
         } else { 
             k as u16 
         } 
     } 
```

patmuk:
Doing so I discovered a deeper bug:
Basically
`KeyCode::from_primitive(1913)`
leads to `KeyCode::No`

(although `KeyCode::from_primitive(0x7779)` converts correctly to `KeyCode::RepeatKey`.

I assume that the num_enum crate doesn't handle gaps in the numeration in the `enum KeyCode`.

Nevertheless, hardcoding as you proposed would work ... tough there might be other keys which are mapped wrongly.

[...]

Issue #309, opened 2025-04-02
Feature: Repeat last pressed Key


patmuk:
It would be great to have a key mapped to repeat the last pressed key, as in [QMK](https://docs.qmk.fm/features/repeat_key).
[Ben Vallack shows in his video](https://youtu.be/5RN_4PQ0j1A?si=a_6nZVsoHU4xLdsS&t=997) how this improves ergonomics, especially when the last KeyAction needed a multiple keys involving combo.

HaoboGu:
Closing this since #318 has been merged

Issue #308, opened 2025-04-02
Feature: KeyAction::Macro


patmuk:
It would be great (and enable chording (#303), if a keypress (or a chord) would trigger multiple outputs.
```
KeyAction::Macro([k!(H), k!(E), k!(L), k!(L), k!(O)], Some(2));
```
would print `hello` for a single KeyAction, optionally only when layer 2 is active.

HaoboGu:
There is macro support in RMK, but only ASCII characters are supported. Macros can be configured in vial

HaoboGu:
ah. no, the macros can only be configured in vial now.

patmuk:
Ah, thanks for the clarification. I am looking into implementing macro support without vial in rmk.

tib888:
#284 is related

patmuk:
After taking a deeper look into the macros, I understood that both can be combined.

The macro processing is already implemented - however, the storage of a macro is in binary form (`Keymap.macro_cache: [u8]`). It is interpreted in the function `process_action_macro` in `keyboard.rs` (actually in `get_next_macro_operation`).

@HaoboGu could you instead move the "on the fly" deserialization where the other deserialization takes place?

If this is refactored, implementing the rest is not much work :)

The data structure could be:

(from `keycap.rs`)
```
pub struct KeyMap<
```
(...)
```
> {
    /// Layers
    pub(crate) layers: &'a mut [[[KeyAction; COL]; ROW]; NUM_LAYER],
```
(...)
```
    /// Macro cache
    pub(crate) macros: [[MacroOperation]; NUM_MACRO],
(...)
}
```

where `MacroOperation` is from `keyboard_macro.rs`.

patmuk:
@HaoboGu's answer, as I accidentally posted my last comment at the wrong place! 

> It is very hard because you have to know the length of `MacroOperation` list at the compile-time.
> 
> In current implementation, you can have 1 macro with length 255 or 25 macros with length 10 or 255 macros with length 1.
> 
> If we do the deserialization early, the max length of macro actions is `MACRO_SPACE_SIZE`/`max_size_of(MacroOperation)`, which might waste lots of RAM space.

patmuk:
Good point. Ok, let's start with a parallel macro structure. Maybe we can later convert this into the Vial macro (as well on the fly) - or we keep them separate.

tib888:
Currently, if you use vial, you can configure arbitrary macro sequence for example on the "M0" macro key.
[...]

PR #318, opened 2025-04-08
implements the repeat function #309

patmuk:
Implements the [repeat key function](https://docs.qmk.fm/features/repeat_key) #309 

However, due to #317 this currently maps to `KeyCode::Again` and not `KeyCode::RepeatKey` as it should.


patmuk:
@HaoboGu  done so far, except for docs and fixing #317

patmuk:
> Please use `rusty_fork_test!` for unit test. This is because the all tests shares same report channel, and the tests will block when the channel is full.
> 
> https://github.com/HaoboGu/rmk/blob/84ab3ff6b74ec191149639a2bd9cada7f7fef713/rmk/src/keyboard.rs#L1742-L1771

Ah, thanks! I added this to all tests (it was only on a few) - hope this is ok.

I did not updated the documentation - as there is no section about special keys ... feel free to point to a section, if I overlooked something! It might be good to mention [the documentation of QMK](https://docs.qmk.fm/features/repeat_key) and mention that `AltRepeat` is not implemented (?).

HaoboGu:
> I did not updated the documentation - as there is no section about special keys ... feel free to point to a section, if I overlooked something! It might be good to mention [the documentation of QMK](https://docs.qmk.fm/features/repeat_key) and mention that AltRepeat is not implemented (?).

Yes, the documentation is missing.. Feel free to add one ;)

patmuk:
After taking a deeper look into the macros, I understood that both can be combined.

The macro processing is already implemented - however, the storage of a macro is in binary form (`Keymap.macro_cache: [u8]`). It is interpreted in the function `process_action_macro` in `keyboard.rs` (actually in `get_next_macro_operation`).

@HaoboGu could you instead move the "on the fly" deserialization where the other deserialization takes place?

If this is refactored, implementing the rest is not much work :)

The data structure could be:

(from `keycap.rs`)
```
pub struct KeyMap<
```
(...)
```
> {
    /// Layers
    pub(crate) layers: &'a mut [[[KeyAction; COL]; ROW]; NUM_LAYER],
```
(...)
```
    /// Macro cache
    pub(crate) macros: [[MacroOperation]; NUM_MACRO],
(...)
}
```

where `MacroOperation` is from `keyboard_macro.rs`.

HaoboGu:
It is very hard because you have to know the length of `MacroOperation` list at the compile-time. In current implementation, you can have 1 macro with length 255 or 25 macros with length 10. But if we do the serialization early, the max length of macro actions is `MACRO_SPACE_SIZE`/`max_size_of(MacroOperation)`, which might waste lots of RAM space.

patmuk:
Ups, I accidentally posted my comment here instead of the right issue (#308). Will copy it there! This PR is finished. Please have a look at the docs and feel free to rewrite it!

HaoboGu (approved):
lgtm, thanks very much!

PR #320, opened 2025-04-08
Implement basic GraveEscape support

vitek:


rmk/src/keyboard.rs
```diff
@@ -1070,6 +1066,24 @@ impl<'a, const ROW: usize, const COL: usize, const NUM_LAYER: usize, const NUM_E
             .await;
     }
 
+    // process action key
+    async fn process_action_key(&mut self, key: KeyCode, key_event: KeyEvent) {
+        let key = match key {
+            KeyCode::GraveEscape => {
+                if self.held_modifiers.into_bits() == 0 {
```
HaoboGu, line 1073: Check shift modifier only?

vitek: I tried to follow original qmk behaviour I'm used to. But shift would work as well.

https://docs.qmk.fm/features/grave_esc#what-your-os-sees

HaoboGu: Cool

Issue #302, opened 2025-03-27
Combo issue


HaoboGu:
1. When there's overlapped combo configuration, such as `{ actions = ["J", "K"], output = "A" }` and `{ actions = ["J", "K", "L"], output = "B" }`, and processing `JKL`, the expected output should be `B`, but now `A` also can be triggered.
2. When using combinations of a combo like `{ actions = ["J", "K"], output = "LShift" }` and other keys, the `LShift` won't be triggered in some cases.

fxliang:
```toml
[behavior.combo]
timeout = "80ms"
combos = [
  {actions = ["F", "J"], output = "Enter"},
  {actions = ["S", "D"], output = "LShift"},
  {actions = ["K", "L"], output = "RShift"},
  {actions = ["E", "R"], output = "WM(LeftBracket, LShift)"},
  {actions = ["D", "F"], output = "WM(Kc9, LShift)"},
  {actions = ["C", "V"], output = "LeftBracket"},
  {actions = ["U", "I"], output = "WM(RightBracket, LShift)"},
  {actions = ["J", "K"], output = "WM(Kc0, LShift)"},
  {actions = ["M", "Comma"], output = "RightBracket"},
  {actions = ["A", "S", "D", "F"], output = "CapsLock"}
]
```

example, `ASDF` as `CapsLock`, when `SD` as `LShift`, `ASDF` sometime might be interrupted by `SD`

hold `SD` won't trigger a `LShift` holding

hitsmaxft:
https://github.com/hitsmaxft/rmk/commit/b24d83b905df2421beebf0f860d1054ffe1f4241

pcasotti:
> [@pcasotti](https://github.com/pcasotti) Could you please have a look?

Sure, I'll take a look later today

PR #322, opened 2025-04-09
Sort combo by length desc #302

hitsmaxft:
This is a simple solution for combo triggering order. Just to make it works.

This PR will prevent early triggering of j+k before j+k+l

### DIFF
--- a/rmk/src/keymap.rs
+++ b/rmk/src/keymap.rs
@@ -46,6 +46,11 @@ pub struct KeyMap<
     pub(crate) behavior: BehaviorConfig,
 }
 
+fn _reorder_combos(combos: &mut [Combo; COMBO_MAX_NUM]) {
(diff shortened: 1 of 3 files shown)


rmk/src/keymap.rs
```diff
@@ -46,6 +46,11 @@ pub struct KeyMap<
     pub(crate) behavior: BehaviorConfig,
 }
 
+fn _reorder_combos(combos: &mut [Combo; COMBO_MAX_NUM]) {
```
HaoboGu, line 49: Why use this separate function, other than `keymap.reorder_combos()`? For easier testing?

hitsmaxft: Yes, since it is called in constructor, I prefer a pure function

HaoboGu:
thanks

PR #327, opened 2025-04-10
Introduce matrix map for nicer configs

tib888:
Available for testing! Updated keyboard configuration documentation is included.

### DIFF
--- a/rmk-macro/src/keyboard_config.rs
+++ b/rmk-macro/src/keyboard_config.rs
@@ -1,12 +1,16 @@
+use pest::Parser;
+use pest_derive::Parser;
+
 use proc_macro2::TokenStream as TokenStream2;
 use quote::quote;
 use serde::Deserialize;
+use std::collections::HashMap;
 use std::fs;
 
 use crate::config::{
     BehaviorConfig, BleConfig, DependencyConfig, InputDeviceConfig, KeyboardInfo,
-    KeyboardTomlConfig, LayoutConfig, LightConfig, MatrixConfig, MatrixType, SplitConfig,
-    StorageConfig,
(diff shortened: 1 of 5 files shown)


rmk-macro/src/keyboard_config.rs
```diff
@@ -79,6 +91,12 @@ pub(crate) struct KeyboardConfig {
     pub(crate) chip: ChipModel,
     // Board config, normal or split
     pub(crate) board: BoardConfig,
+
+    // // Aliases for key maps
+    // pub (crate) aliases: HashMap<String, String>,
```
HaoboGu: Remove if they are not needed?

HaoboGu:
I found that pest is introduced to parse the keymap. How do you feel about it? especially for the learning curve. I worry that it may prevent people contributing on it if it's difficult.

tib888:
During parsing these are built:
```
sequence_to_grid: Option<Vec<(u8, u8)>>
grid_to_sequence: Vec<Vec<Option<u32>>>
```
Shuould they be stored in the LayoutConfig?
If you plan to create vial.json automatically, that should be based on these somehow.
But even without that I think this PR is acceptable.

rmk-macro/src/keyboard_config.rs
tib888: done.

HaoboGu:
oh I found I missed vial.json part. vial.json needs to be labelled for vial app to recognize. Now is it possible to generate vial.json if users have matrix map in toml config?

tib888:
Not yet, because I do not undursteand exactly how you imagine that.
I would reccommend to accept this PR as is, then in a next one solve that.

HaoboGu:
thanks!

PR #328, opened 2025-04-11
Raw joystick support

HigherOrderLogic:
**Note: this PR is untested**

Add support for raw joystick input, beside the mouse mapping.

If possible, I also would like to support Xinput as well.


rmk-macro/src/input_device/adc.rs
```diff
@@ -73,14 +73,24 @@ pub(crate) fn expand_adc_device(
                 });
                 let joy_ident = format_ident!("joystick_processor_{}", joystick.name);
                 processor_name.push(quote!(#joy_ident));
+
+                let map_to = match joystick.map_to.unwrap_or("mouse".into()).as_str() {
+                    "mouse" => quote! { rmk::input_device::joystick::MapTo::Mouse },
+                    "joystick" => quote! { rmk::input_device::joystick::MapTo::Joystick },
+                    unsupported => {
+                        panic!("Unsupported map_to type: {}", unsupported)
```
HigherOrderLogic: Is there a better, more elegant way to handle compile error here?

rmk/src/hid.rs
```diff
@@ -3,15 +3,39 @@ use core::{future::Future, sync::atomic::Ordering};
...
+            (usage = Y,) = {
+                #[item_settings data, variable, absolute] y = input;
+            };
+        };
+    }
+)]
+pub struct JoystickReport {
+    pub x: i8,
+    pub y: i8,
+}
```
HigherOrderLogic, line 34:
```suggestion
pub struct JoystickReport {
    pub x: i8,
    pub y: i8,
    pub pressed: boolean.
}
```

Should I add this new state to represent the user pressing the joystick (not Z axis, this is usually wire to the `SEL` pin of the joystick), or let user handle it just as another normal key?

rmk-macro/src/input_device/adc.rs
HaoboGu: panic at the compile-time is fine. Adding something making users to know which configuration is wrong would be better

rmk/src/hid.rs
HaoboGu: I prefer to handle it as a normal key in the key matrix now. 

I know that some boards connect the pin directly to gpio, making it a separate key out of the key matrix. A separate direct pin matrix should be added in this case. We could improve it later though.

HigherOrderLogic:
@drindr  
Just to make sure before buying component to test, your joystick driver support hall effect joystick (and variant like TMR) right?

drindr:
yep

HaoboGu:
I'm closing this PR since there's no activities for a while, please feel free to reopen it if you want

PR #331, opened 2025-04-12
Refactor/storage of combos n forks

patmuk:
I noticed that firmware configured combos & forks are copied in `keymap.rs`
```
        // If the storage is initialized, read keymap from storage
        let mut combos: heapless::Vec<Combo, COMBO_MAX_NUM> = Default::default();
        for (i, combo) in behavior.combo.combos.iter().enumerate() {
            combos[i] = combo.clone();
        }
```
(Same for fork.)

This PR uses the combos and forks from the configuration directly, using headless::Vec instead of Arrays.
Like before, the vectors are filled with default (empty) placeholders so Vial configured combos still work (I tested this).

The functionality of the code should not be changed at all, but the resulting firmware is smaller and the code a bit cleaner.

### DIFF
--- a/rmk/src/keyboard.rs
+++ b/rmk/src/keyboard.rs
@@ -358,13 +358,13 @@ impl<'a, const ROW: usize, const COL: usize, const NUM_LAYER: usize, const NUM_E
     /// The replacement decision is made at key_press time, and the decision
     /// is kept until the key is released.
     fn try_start_forks(&mut self, key_action: KeyAction, key_event: KeyEvent) -> KeyAction {
-        if self.keymap.borrow().forks.len() < 1 {
+        if self.keymap.borrow().behavior.fork.forks.len() < 1 {
             return key_action;
         }
 
         if !key_event.pressed {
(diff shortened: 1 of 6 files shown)


rmk/src/keyboard.rs
```diff
@@ -358,13 +358,13 @@ impl<'a, const ROW: usize, const COL: usize, const NUM_LAYER: usize, const NUM_E
     /// The replacement decision is made at key_press time, and the decision
     /// is kept until the key is released.
     fn try_start_forks(&mut self, key_action: KeyAction, key_event: KeyEvent) -> KeyAction {
-        if self.keymap.borrow().forks.len() < 1 {
+        if self.keymap.borrow().behavior.fork.forks.len() < 1 {
```
HaoboGu: use is_empty?

patmuk: good catch!

Did another linter warning ("use enumerate()") in three other places as well.

HaoboGu:
thanks!

PR #332, opened 2025-04-12
fix: align keycode with QMK

HigherOrderLogic:
Add more keycode alias to better align with QMK.

There are a lot that's still missing. If you want, I can add them all in this PR.


rmk-macro/src/keycode_alias.rs
```diff
@@ -4,343 +4,247 @@ use once_cell::sync::Lazy;
...
+    add_alias!("Menu");
+    add_alias!("Select" = "slct");
+    add_alias!("Stop");
+    add_alias!("Again" = "agin");
+    add_alias!("Undo");
+    add_alias!("Cut");
+    add_alias!("Copy");
+    add_alias!("Paste" = "pste");
+    add_alias!("Find");
+    add_alias!("KbMute" = "kb_mute", "mute"); // FIXME: mismatch with QMK. `mute` in QMK is mapped to `KC_AUDIO_MUTE`. The same for the other `Kb*` keys.
```
HigherOrderLogic: Should I change this?

image

HigherOrderLogic:
I have added all the missing keycode aliases.

I tried to maintain compatibility with the old code, but it seems like some are missing, and some are redundant. If you want to perfectly align with QMK and the [HID spec](https://www.usb.org/sites/default/files/documents/hut1_12v2.pdf) (which is better imo), I can do that too. However, it means that there'll be breaking changes.

HaoboGu:
> If you want to perfectly align with QMK and the [HID spec](https://www.usb.org/sites/default/files/documents/hut1_12v2.pdf) (which is better imo), I can do that too

Yes, it's better to be aligned with HID spec.

> However, it means that there'll be breaking changes.

Don't worry! The next release will be another breaking change ;)

rmk-macro/src/keycode_alias.rs
HaoboGu: Is it fixed?

HigherOrderLogic: I have made changed on my local machine, waiting on finishing the others one to push it onto Github.

HaoboGu:
Hi, is this pr considered complete?

HigherOrderLogic:
> Hi, is this pr considered complete?

Yes. If possible, some testing from the community would be great.

HaoboGu:
I've tested it. It works well!

However, when I'm configuring the keymap, I found that we should add some alias back, for example, `-`, `=`, `[`, `]`, etc. For me, using those symbols are more intuitive.

I think we don't have to be 100% aligned with QMK, having something that makes it easier to config is fine. Also, some documentations should be added to tell users the available symbols they could use.

HaoboGu:
Nice work! Would you mind adding those symbols back and adding a documentation which listing the available alias?

HigherOrderLogic:
> I found that we should add some alias back, for example, `-`, `=`, `[`, `]`, etc. For me, using those symbols are more intuitive.

While I agree that those are more intuitive and easy to read, it is easy to mistake similar keys. For example, for the minus symbol, there are 2 version, which is `KpMinus` and the normal `Minus`. If you still want it, we can go ahead and add the symbol alias for the normal variants, while ignoring their keypad and other counterparts.

> Also, some documentations should be added to tell users the available symbols they could use.

I'll add that later, once the aliases are finalized to avoid extra works.

HaoboGu:
> While I agree that those are more intuitive and easy to read, it is easy to mistake similar keys. For example, for the minus symbol, there are 2 version, which is KpMinus and the normal Minus.

We already have something similar,  for example 0-9 and kp0-kp9. I don't think it will be a major issue. And yeah, that's why we need a list or a table for it. I believe it's worth to do it for better user experience

HigherOrderLogic:
@HaoboGu 
How does it looks now? If there is nothing to change, I'll go ahead with the docs.

HaoboGu:
looks great！

HaoboGu:
It looks great for me now @HigherOrderLogic, please go for the docs :D 

Thanks a lot!

HaoboGu:
Thanks for your effort! Can I merge this PR now?

HigherOrderLogic:
> Thanks for your effort! Can I merge this PR now?

There's still some less important keycodes missing from the docs. I'll get back to it in a few days when I have time.

PR #336, opened 2025-04-14
fix/joystick: optimize the first polling and parameters about duration

drindr:
1. Get correct values rather than 0 in the first polling.
2. Replace the type of some integer parameters about time with `Duration`.


HaoboGu (approved):
lgtm, thanks!

Issue #303, opened 2025-03-30
Q: adding chording keyboard support


akavel:
Hi!
I'm developing a chording keyboard, with firmware written in Rust (current code at: https://github.com/akavel/clawtype; some photos of the current version at: https://www.printables.com/model/1231156-clawtype). I wonder, would you be open for me to try and add features of such a keyboard to RMK? If yes, would you have some guidelines on where and how to do this?

FWIW, I think it might make sense to make it some kind of a separate "major mode" or something; or maybe a layer mode, hard to say; the point is, in such a chording keyboard, chords are a primary means of operation. On the other hand, I do have one layer for use with a (gyroscope-based) mouse, where I "disconnect" two keys from the chording mode, to make them work as "normal" (non-chorded) mouse buttons. Notably, in this keyboard the keys must be non-matrix, each one needs to be connected to a separate GPIO, as I must be able to properly handle "full NKRO".

I'm writing to ask about your thoughts on whether this might be a useful thing to try to add to RMK, and you could have some suggestions how I could try to do that - or if you don't think so.

HaoboGu:
I'm quite open to this, RMK provides direct pin matrix and combo support which might be useful for your case. But as far as I can tell, there are still work to do, such as NKRO, configurable number of combos. etc.

I have to say, this could be hard to implement all those features, but if you want to try, I'd be happy to help.

akavel:
Thanks for the reply and openness!

So, my first important question would be: does rmk currently have any support for _NON-matrix_ keys?

HaoboGu:
Yes, there's `direct_pin` just for it. The direct pin is organized to matrix, but you can use a 1*N matrix.

patmuk:
That sounds great - I would like to implement chording as well for my sweep.
It looks to me like there is only a macro KeyAction missing - meaning outputting several keys when triggering a combo.

For example: 
Holding the "chording-key" and pressing H E would output "hello".
This could be something like
```
Combo::new([k!(H), k!(E)], KeyAction::Macro([k!(H), k!(E), k!(L), k!(L), k!(O)], Some(2));
```
where the "chording key" activates the layer 2.
`KeyAction::Macro` would need to be implemented - but that might not be hard?

tib888:
Can you show a preview, how the config file will look like?

patmuk:
> Can you show a preview, how the config file will look like?

Sure: In the example I will define two Macros:

In your `keymap.rs`:
```
[...]

Issue #313, opened 2025-04-02
How to create an Umlaut?


patmuk:
On my Mac I would like to print an Umlaut, like ä, ü and ö.
Is there a KeyCode for it? Or is there a KeyCode for inputting Unicode?

Otherwise the combination `" + u + a/u/o` does that on a Mac, but would require #308.

HaoboGu:
also related #170

HaoboGu:
I'm closing it as it's duplicate with #170

Issue #284, opened 2025-03-15
Macro configuration


tib888:
Macro configuration is not possible from keyboard.toml or at least the documentation is missing.

Issue #170, opened 2024-12-06
Unicode support like QMK's UCIS


petermarks:
Does RMK have any support for sending unicode characters, or is anything planned? According to https://docs.qmk.fm/features/unicode, this is non-trivial and not consistent across OSs.

Similar to Linux's compose key feature, I'd like to use 2, 3 or even longer character sequences to send unicode characters. eg. --. would output – (en dash),  :) would send 😊. To be clear, I want to implement this in the keyboard rather than sending a keycode for compose as that isn't supported on Windows.

I envision either:

1. Press a "compose" key, type a sequence,  then have these replaced by a unicode character. 
2. Type the sequence, then press "compose" and have the keyboard replace what I typed with a unicode character.

The second approach has to send the characters to the host, but keep a history in a buffer, in case the compose key is pressed. If it is, it backspaces the sequence and sends the unicode (in an OS specific way). The first approach can be implemented similarly (as is done in QMK, or can handle the whole thing in the keyboard before sending anything. I personally favour 2, I think, but either would be fine. I realise this is limited to text editing contexts, but that is a perfectly reasonable limitation.

Any thoughts?

HaoboGu:
Currently this feature is not planned. PRs are welcome!

Personally I prefer the first solution since it seems to be a "leading key" feature and more easily to implement.

petermarks:
Thanks @HaoboGu, I think you are probably right as it is more similar to other mechanisms.

Reading the QMK doc further, it seems Windows support for unicode input is limited and not considered reliable, so additional software on the host is needed. If that's the case, then I I'll just go with the standard compose key on Linux and use WinCompose to get the same functionality on Windows.

I may have a look at implementing it in RMK once I get a bit more familiar with the code.

tib888:
I'm looking for this functionality too since I like to use US layout for coding because all brackets and operators are easy to accces and  I use English in work anyway. 
I hate when I have to switch to HUN layout when rarely I have to type hungarian properly with this nice set of extra letters: "áéíóöőúüű" and their capital versions so 9 keys wasted for these and all symbols are accesible with AltGr+... combinations.
So I ASAP switch back to the US layout.

So my idea is to use a separate **script** for each letter, which may either:
- Type Alt+160 on Numpad to get "á" - unfortunately this will not work for "ő,ű", which were forgotten by the standardization comitee back then so they need unicode.
- Type Alt+hexaunicode - but this needs enabling this function in windows beforehand
[...]

PR #337, opened 2025-04-14
Feat/macro

patmuk:
Keyboard macros are supported by Vial configuration (only).

This PR refactors the code and adds the possibility to define keyboard macros via Rust, which are compatible with (and shown in) Vial. The user input is converted to the binary format used by Vial.

Open todos:
- [x] document usage
- [x] implement KeyAction::MacroTrigger as an additional way to configure macro triggers (additional to and compatible with KeyCode::M0-31)
- [x] refactor "process_macro_action" to avoid the key un-/registering


patmuk:
I just noticed the `cargo check` gives the same warning in `rmk/examples/use_rust/nrf52840_ble_split/` - where I updated my `central.rs` from to be compatible with master. Some change there must be responsible ... but I can't see what. `rmk/examples/use_rust/nrf52840_ble/` doesn't emit this warning.

patmuk:
> yes, I had it too. I've fixed on my branch. Will merge it soon

Ah, super :) Is it a small enough fix so you can tell me how? I would like to fix that in my own project as well ...

HaoboGu:
Removing unnecessary `mod`s in `peripheral.rs`, such as `mod macros` and `mod keymap`

patmuk:
> Removing unnecessary `mod`s in `peripheral.rs`, such as `mod macros` and `mod keymap`

Ah, great, that worked :) Thank!

patmuk:
@HaoboGu I am currently implementing `Action::TriggerMacro` as an alternative to `KeyCode::MacroX`.
I don't understand why the KeyEvent::Row and COL is needed in
`keyboard.rs line 1281`, at the function `register_keycode`. Is the key's position sent in the Hid report as well?

I am currently just creating a `KeyEvent{0, 0, false}` for the new `Action::TriggerMacro`.

HaoboGu:
> @HaoboGu I am currently implementing `Action::TriggerMacro` as an alternative to `KeyCode::MacroX`. I don't understand why the KeyEvent::Row and COL is needed in `keyboard.rs line 1281`, at the function `register_keycode`. Is the key's position sent in the Hid report as well?
> 
> I am currently just creating a `KeyEvent{0, 0, false}` for the new `Action::TriggerMacro`.

It prevents registering a same key multiple times. And also useful when keys are stuck. If a key is stuck, pressing and releasing the same key will clear the state.

patmuk:
All works - except macros over 8.
Meaning, using `KeyCode::Macro0 -7` or `Action::TriggerMacro(0 - 7)` works, assigned as a key as well as in a combo.
Anything over 7 does not work at all - the keycode is shown in vial (for example, of `TriggerMacro(8)` as `0x7708` - but pressing it doesn't do anything.

@HaoboGu or @tib888 could you please take a look into my code? Maybe you can easily see what I overlooked.

Here is how you can configure the Macros:
```
pub(crate) fn get_macro_sequences() -> [u8; MACRO_SPACE_SIZE] {
    define_macro_sequences(&[
        to_macro_sequence("Macro0"),
        to_macro_sequence("Macro1"),
        to_macro_sequence("Macro2"),
        to_macro_sequence("Macro3"),
        to_macro_sequence("Macro4"),
        to_macro_sequence("Macro5"),
        to_macro_sequence("Macro6"),
        to_macro_sequence("Macro7"),
        to_macro_sequence("Macro8"),
        to_macro_sequence("Macro9"),
        to_macro_sequence("Macro10"),
        to_macro_sequence("Macro11"),
        to_macro_sequence("Macro12"),
        to_macro_sequence("Macro13"),
        to_macro_sequence("Macro14"),
        to_macro_sequence("Macro15"),
        to_macro_sequence("Macro16"),
        to_macro_sequence("Macro17"),
        to_macro_sequence("Macro18"),
        to_macro_sequence("Macro19"),
        to_macro_sequence("Macro20"),
        to_macro_sequence("Macro21"),
        to_macro_sequence("Macro22"),
        to_macro_sequence("Macro23"),
        to_macro_sequence("Macro24"),
        to_macro_sequence("Macro25"),
        to_macro_sequence("Macro26"),
        to_macro_sequence("Macro27"),
        to_macro_sequence("Macro28"),
        to_macro_sequence("Macro29"),
        to_macro_sequence("Macro30"),
        to_macro_sequence("Macro31"),
        to_macro_sequence("Macro32"),
    ])
}
```

The macros automatically get bound to KeyCode::Macrox and Action::TriggerMacro(x), where x is the position in the array.

patmuk:
Argh, I found the problem: MAX_NUM_MACROS (which was 8).

I changed that to 256. This limit is basically because of `to_action_code` of Action has only 8 bits for the id left.
However, I think that `to_action_code` is used only in `to_key_action`, which is used nowhere (?).

Thus, it could be more macros ... but 256 is enough for me :)

As for the space it takes 256 bytes. And this array is filled up with zeros, so one can configure things via Vial. 
We could optimize that in three ways: 
1. limit the minimum size to the number of macros actually supported by Vial. That seems to be 16. However, as the content of a macro could be arbitrarily long, there is no correct number. I mean, if we set the limit to, for example, 16, it would allow for 8 macros Text with 1 letter (1 byte for the ascii letter + 1 byte terminal symbol) or 1 macro with 15 letters. So not really an option.
2. Rewriting the macro sequence when serializing into bytes: The MacroOperation::tap(key) seems to be equivalent to MacroOperation::Text(key). But `tap` is 3 bytes (2 bytes identifier and 1 byte ascii letter), while `text` is only 1 byte (the ascii letter). It is even worse for a capital letter, as `Press shift` and `release shift` have to be added, which are 6 bytes. So, if @HaoboGu you know if `tap` and `text` is the same I could convert the former to the latter for a smaller resulting binary. However, the user can choose which one he wants - and I guess that for text he would always anyways take MacroOperation::Text.
3. Make `MACRO_SPACE_SIZE` configurable. Not sure if we need `MAX_MACRO_NUM` at all.

I think (1) doesn't make sense, (2) is a minimal optimization with the risk that we overlook something (it is separate in Vial as well) and (3) is what we should do.

In the code I do check that the resulting binary is small enough.

patmuk:
@HaoboGu please have a look.
All is working now, I added documentation as well. Maybe start from there :)

This closes #303, and potentially a few additional issues as well, though I did not test for them.
#313 might work now, #170 probably as well.

Note that this might be a breaking change for some!

In the last days I removed some duplication: Instead of storing combos and forks in `keymap` and `BehaviorConfig` twice I changed the code (in another PR) and removed the cloning so that they are only stored in the `BehaviorConfig`.
The macro sequences binary (`macro_sequences`, formaly known as `macro_cache`) is in `BehaviorConfig` as well.

BehaviorConfig was in `keymap` and `RmkConfig`. I removed it from the latter, now it is in `keymap` only.

Thus two changes are needed:
`initialize_keymap_and_storage` and similar take a 
1. `&storage_config` instead of `rmkConfig.storage_config.clone()`
2. `behavior_config` instead of `rmkConfig.behavior.clone()`
Thus, is the `behavior_config` was silently created with 
```
RmkConfig{
..Default::default()
}
```
it now has to be explicitly defined with
```
let behavior_config = BehaviorConfig::default();
```

The examples and macros are all adapted to that.


I did not implement configuring macros via the toml configuration (#284) - this should be another PR.

Feel free to do so, it should be trivial having this code here.

I will be off for a two weeks vacation (to China and Japan, yeah!), will only sporadically check comments but won't be able to code anything.

Hope all is fine with this PR!

patmuk:
Oh, one more thing: I didn't implement any integration tests. Single functions (like conversions) are unit tested, but I did not see how I can intercept the key report to test if a macro sequence is actually triggered.
Though I stepped through a unit test without assert statements and tested it on my nice!nano aurora sweep.

tib888:
> 2\. The MacroOperation::tap(key) seems to be equivalent to MacroOperation::Text(key)

Look at keyboard.cs / process_action_macro().

'Tap' is not the same as 'Text' because you can not text the "F1" key can tap it, also all active modifiers are suppressed, while "texting". So for pure text I recomment using "Text", but more compex use cases may need Tap.

patmuk:
Ah, thanks, that is good to know! I thought modifiers are suppressed always :) Will update the documentation.

rmk/src/config/keyboard_macros/macro_config.rs
```diff
@@ -0,0 +1,26 @@
+/// Default macro space size
+/// the sum of alll macro elements + number of macro elements
```
tib888: alll -> all

docs/src/keymap/keyboard_macros.md
```diff
@@ -0,0 +1,212 @@
+## Keyboard Macros
+rmk supports keyboard macros: 
+Pressing a trigger to execute a sequence of keypresses.
+
+This can be configured via Vial or rust. 
+A configuration via the toml configuration file will be provided in the future.
+
+## macro operations
+The following operations, coming froim Vial, can be used to form a macro sequence.
```
tib888: froim -> from

docs/src/keymap/keyboard_macros.md
```diff
@@ -0,0 +1,212 @@
...
+        to_macro_sequence("type"),
+        Vec::from_slice(&[
+            MacroOperation::Press(KeyCode::Backspace),
+            MacroOperation::Text(KeyCode::I, false),
+            MacroOperation::Text(KeyCode::N, false),
+            MacroOperation::Text(KeyCode::G, false),
+        ])
+        .expect("too many elements"),
+    ])
+
```
tib888, line 192: What is the advantage of this 'Type' + 'ing'? 
Is this using less memory?
If one macro would be able to trigger other then it would, but i think that "recursion" does not work currently...

Also, "Press(KeyCode::Backspace)" is really needed there?

docs/src/keymap/keyboard_macros.md
```diff
@@ -0,0 +1,212 @@
...
+            MacroOperation::Tap(KeyCode::D),
+        ])
+        .expect("too many elements"),
+    ])
+}
+```
+This code defines two macro sequences which produce "Hello" and "World".
+(As mentioned above prefere the first Macro for text only output. The first macro sequence is 6 bytes long, the second 22 bytes.)
+
+For text output there is a convinience function: `to_macro_sequence(text: &str) -> heapless::Vec<MacroOperation, MACRO_SPACE_SIZE>`.
```
tib888: to_macro_sequence() -> if not able to handle umlaut (and other similar characters) that should be mentioned here.
(Or it could have a version which expect a coversion table for such language characters, for example   hungarian KeyCode::É is mapped on Semicolon, Á is mapped on KeyCode::Quote, etc.)

tib888:
> I will be off for a two weeks vacation (to China and Japan, yeah!), will only sporadically check comments but won't be able to code anything.
> 
> Hope all is fine with this PR!

I haven't tested yet, but seems to be a nice work!
Have a nice vacation!

HaoboGu:
> However, I think that to_action_code is used only in to_key_action, which is used nowhere (?).

yeah, it need to be removed. It's legacy code which is not used anymore.


> Make MACRO_SPACE_SIZE configurable. Not sure if we need MAX_MACRO_NUM at all.

This is what I'm going to do next! Making all those consts configurable is the next key feature

HaoboGu:
Great work! I'll take a look this week.

Have a good vacation :D

docs/src/keymap/keyboard_macros.md
patmuk: Thanks for finding the typos!

docs/src/keymap/keyboard_macros.md
patmuk: Oh, I tried to explain it better now.

I wasn't thinking of recursion. I avoided that as well, to not let the user run into en endless loop.

I meant having a single macro for `[Backspace] + ing`, which can be used after executing other macros for verbs.
For example, if one has a macro for `type` one would execute that and afterwards press the key to execute the `ing` macro, so that the `e` gets deleted and `ing` added, resulting in `typing`.

Adding ` write` we could reuse the `ing` macro to get `writing`. So instead of having one macro for each form (`type`, `typing`, `write`, `writing`) we would have only three (`type`,`write`, `[Backspace] ing`).

(This helps but doesn't work for all verbs: `read` -> `reaing` :) ). It would be nice to have a `ing`-modding function - but that is another feature. One would need to have a word table so that the keyboard knows hoe to transform each verb.)

docs/src/keymap/keyboard_macros.md
patmuk: Good point. I clarified that with an example.
Since it is OS and keyboard layout (in the OS) dependent I just mentioned how one could do that for their own setup.
I think the only way to do it OS & Layout independent would be by defining new `KeyCode`s.
However, the keyboard would have to have a way of detecting the OS and layout to send the right keys for a specific character.

patmuk:
> > However, I think that to_action_code is used only in to_key_action, which is used nowhere (?).
> 
> yeah, it need to be removed. It's legacy code which is not used anymore.
> 
> > Make MACRO_SPACE_SIZE configurable. Not sure if we need MAX_MACRO_NUM at all.
> 
> This is what I'm going to do next! Making all those consts configurable is the next key feature

Ah, good to know :) For now I implemented it for `Action::TriggerMacro` as well. Once we remove this function we can address more macros - but 256 should be enough for now :)

patmuk:
> > > However, I think that to_action_code is used only in to_key_action, which is used nowhere (?).
> > 
> > 
> > yeah, it need to be removed. It's legacy code which is not used anymore.
> > > Make MACRO_SPACE_SIZE configurable. Not sure if we need MAX_MACRO_NUM at all.
> > 
> > 
> > This is what I'm going to do next! Making all those consts configurable is the next key feature
> 
> Ah, good to know :) For now I implemented it for `Action::TriggerMacro` as well. Once we remove this function we can address more macros - but 256 should be enough for now :)

We could do that then the consts are configurable.

patmuk:
> Great work! I'll take a look this week.
> 
> Have a good vacation :D

Thanks!! And @HaoboGu  and @tib888  thanks for your feedbacks and review!

docs/src/keymap/keyboard_macros.md
```diff
@@ -0,0 +1,271 @@
...
+Note that ofther modifiers pressed outside of a sequence with `Text` are disabled.
+
+### Tap(KeyCode)
+Presses and releases a key. 
+Modifiers pressed outside of a macro sequence are considered as well.
+If you don't need this prefer `Text(KeyCode, bool)` above, as the resulting macro is 3 times smaller in size.
+
+### Press(KeyCode)
+Press (and hold) a keycode.
+Usefull for modifier keys.
```
HaoboGu: Useful?

docs/src/keymap/keyboard_macros.md
```diff
@@ -0,0 +1,271 @@
...
+### via the configuration file
+This is not yet supported.
+
+### via Rust
+A new field `keyboard_macros` has been added to the `BehaviorConfig` struct.
+Within it a field `macro_sequences` has to be set.
+This is in binary format (`[u8]`) and can only be as long as `MACRO_SPACE_SIZE`, which is set to 256.
+
+The maximum number of Macros depends on the length of the sequences:
+The space consued is MacroOperations * 3 + Number of Macros (where the operation `text` is only 1/3).
```
HaoboGu: consued -> consumed

rmk/src/config/keyboard_macros/keyboard_macro.rs
```diff
@@ -0,0 +1,412 @@
...
+#[derive(Debug, Clone)]
+pub enum MacroOperation {
+    /// 0x00, 1 byte
+    /// Marks the end of a macro sequence
+    /// Don't use it on your own,
+    /// will be automatically removed and added
+    /// by MacroOperations::define_macro_sequences()
+    End,
+    /// 0x01 01 + 1 byte keycode
+    Tap(KeyCode),
```
HaoboGu, line 24: `Tap(KeyCode::A)` is converted to `[0x01, 0x01, 0x04]`, am I right?

Why don't use 0x01 for `Tap`, 0x02 for `Press` and so on?

docs/src/roadmap.md
```diff
@@ -22,6 +22,13 @@ There are a bunch of things to do with RMK in the near future. I plan to ship 1.
   - [x] Direct pin
   - [x] 🟢 encoder
   - [x] 🔵 Input device
+  - [ ] 🔴 Macro support
+    - [x] 🔴 Macro definition via Rust
+    - [ ] 🟢 Macro definition via toml
+    - [ ] 🔴 Macro's for umlaute
+    - [ ] 🔵 Macro's for unicode
```
HaoboGu: Macro's -> Macros?

Do the umlaute and unicode needs to be done with macros? I mean, I think they can be implemented separately 🤔

docs/src/keymap/keyboard_macros.md
```diff
@@ -0,0 +1,271 @@
...
+    ])
+}
+```
+
+
+## Triggering a macro
+### binding
+A macro can be triggered in two ways:
+1. Using the `KeyCode::Macro0` - `KeyCode::Macro31`.
+1. Using the `Action::MacroTrigger(index)`, where index can be any number up to `MAX_NUM_MACRO` (which is set to 256).
```
HaoboGu: Is it possible to set `Action::MacroTrigger(index)` key in vial? Especially when index > 32.

Vial supports to set a raw code, for example 0xFFFF, so if it's possible, having some doc for it would be great.

docs/src/roadmap.md
patmuk: I am back, had fantastic holidays :)

I fixed the spelling - thanks!

Hmmm ... we should at least not do Umlaute and Unicode in this PR.

Doing it with Macros is one way - maybe not the only way, but no other way comes to my mind. 

I don't know deep enough how HID codes work ... but aren't they dependent on the keyboard setting in the OS?
Meaning that a specific code is interpreted as `:` in en-us and `ö` in de-de? Additionally, it depends on the keyboard setting itself (ANSI vs ISO-something) and OS (`@` is a different combination on MacOS than on Windows).

So, in my setting: MacOS with en-us keyboard, `ö` is printed when pressing ```Alt + U and O```, thus this would need a combo.

Regarding Unicode, there is a "unicode" keyboard in MacOS, where pressing `Alt` and a sequence of numbers emits a unicode symbol. This would need a macro as well.

... I changed it in the roadmap to 
```
[ ] support umlaute
[ ] support unicode
```
independent to the macro way of doing it.

It would be great if anybody knows this better than me. Otherwise, using macros like I described above works, have tested it.

But yes, in summary I changed the goals to be more solution independent - and if you want we can remove them from the roadmap altogether.

docs/src/keymap/keyboard_macros.md
patmuk: No - the `Action::MacroTrigger(index)` is not bound to a `KeyCode`. One could reserve additional keycodes in an unused space - but I did not implement that.

Basically when a key is pressed the algorithm checks if it was bound to `Action::MacroTrigger(index)` or `Action::Single(KeyCode)`.

For example, one can bind one key to `Action::MacroTrigger(0)` and another to `Action::Single(KeyCode::Macro0`.
Pressing either key would execute the first macro. But a switch would branch off at different places and execute nearly the same code later.

patmuk:
@HaoboGu from my POV this PR is ready to be merged. Though I am happy to integrate any feedback if anyone has further comments :)

docs/src/roadmap.md
HaoboGu: yes, the Unicode support depends on the OS. They're not in the HID spec.

HaoboGu:
great work! could you rebase to the latest main branch? then we can get it merged

patmuk:
> great work! could you rebase to the latest main branch? then we can get it merged

Thanks :) Sure, will do that today.

docs/src/roadmap.md
patmuk: Ok - I will add to the docs an example how to do it with macros on a Mac.

patmuk:
I have troubles with the latest main branch.
I thought first that the problem is because I merged wrong. However, I now changed my rust-based configuration to work on the main branch directly.
After applying changes from `example/use_rust/nrf52840_ble_split` to my `central.rs`, compiling works, but after flashing the keyboard does nothing. Probably it panics - but I don't have a probe to check that.
I am using nice!nanos on a bluetooth keyboard.

Are you sure that the configuration works? Is there a minimal default value set I can try?
Do you have tips how I can debug it?

HaoboGu:
tbh it's quite hard to debug panics if you don't have a debug probe.. I manually merged the changes and pushed to your branch. It works on my board, could you have a try? Just double check it. Thanks!

patmuk:
Oh, great, thanks! Will try that and thanks for the confirmation!

I could get the central half working by copying the example into my project and applying minimal changes (the matrix and pin config) - so the problem is definitively on my side.

HaoboGu:
Cool! So can we get this PR merged now?

tib8:
> so the problem is definitively on my side.
lets check your cargo.toml, lots of dependencies has been updated recently on main...

HaoboGu:
Oh, for the peripheral side, you might want to clear_storage first. The default address of the peripheral changed.

patmuk:
Thanks! I did that but it didn’t work … I set `clean_storage` to true and pressed the reset button on both n!nanos.  The central side works via bluetooth as well, but the right doesn’t. I will do that again later, just to be sure. 
Is the peripheral’s address set in both, central.rs run_peripheral_manager and in peripheral.rs?

HaoboGu:
> Is the peripheral’s address set in both, central.rs run_peripheral_manager and in peripheral.rs?

Now the central and the peripheral are auto-paired. Then they will store the peer's address in the flash. So manually setting the peer address is no longer needed.

patmuk:
Ah, cool :)
I cleaned the halfs with zmk's reset fw and flashed them again - but still the peripheral doesn't work.
The code which figures the bluetooth address is
```
fn ble_addr() -> [u8; 6] {
    let ficr = embassy_nrf::pac::FICR;
    let high = u64::from(ficr.deviceid(1).read());
    let addr = high << 32 | u64::from(ficr.deviceid(0).read());
    let addr = addr | 0x0000_c000_0000_0000;
    unwrap!(addr.to_le_bytes()[..6].try_into())
}
```
right? Identical in central.rs and peripheral.rs. 

I can see that central.rs reads the peripheral's address out of the storage ... but I fail to see how it ends up there?
Could it be that central tries to read before peripheral wrote it? And the tests did not reveal that, as after the first time it was writing the same address?

I am now double-checking if my central.rs and peripheral.rs is different in more than different input/output pins, row & col numbers and the removal of the encoder specific code (maybe adding this).

HaoboGu:
> Identical in central.rs and peripheral.rs.

Yes, it uses chip's unique address as the BLE address.

> Could it be that central tries to read before peripheral wrote it?

In this case central gets nothing. Then it scans for the peripheral to get the peripheral's address and saves the address to the storage. On the next boot, it will use the saved address.
Same on the peripheral side.

HaoboGu:
> start_addr is 0xA0000 in central.rs, but 0x60000 in peripheral.rs. 

that's fine

patmuk:
> > start_addr is 0xA0000 in central.rs, but 0x60000 in peripheral.rs.
> 
> that's fine

I changed that with no visible effect, still works.

Ok, will now finalize the PR and write you once it is ready for merge :)
Might ask here how to overwrite the default value for MACRO_SPACE_SIZE - but will look into it first myself :)

tib8:
I know I have already mentioned this, but don't you want to change TriggerMacro to take the text directly instead of a macro number, like: Action::PlayMacro("text")?
So you could have almost "infinite" number of macros, stored in the "code segment". 
Of course in this case they could be only reconfigured by RMK recompile, not by VIAL, but for your use case I think that is fine...

(KeyCode::MacroXX would work as currently, remain configurable in VIAL.)

patmuk:
> I know I have already mentioned this, but don't you want to change TriggerMacro to take the text directly instead of a macro number, like: Action::PlayMacro("text")? So you could have almost "infinite" number of macros, stored in the "code segment". 
> Edit: Never mind the above! Vial should show a keycode an there is no way to provide a keycode from "text"....

Oh yes, I recall that now :) I think that can be done - but it would break compatibility with vial. On one hand I am not sure what vial would show - in best case the same keycode for all "PlayMacro" bindings.

We could do that in another PR. After finishing this I will take care of my other project and return to configure my chording-macros at a later time. I might run into space issues - if so I might implement it this way :)

patmuk:
Could you please check if combos still work for you?

Both halfs work, but no combo is triggered. 

Additionally I have a fork definition, where the the trigger is `k!(KpEqual)` with the mod `shift` and target `wm!(I, LOPT)` (LOPT is a shortcut I defined). Compile works, but the flashed halfs don't put out any letter - I assume it panics. 

These two things lock like regressions to me ... the fork one is not as bad, as I didn't test it before (did not executed the fork, but the definition did not panic).

But no combo works for me - all looks fine except that the combos are not triggered. Looking now deeper into it, maybe I overwrote some timing setting.

patmuk:
The combos work!

I found my mistake - I commented out the macro definitions but used the triggers in the combos. This is where the panic came from!

So all fine :)

docs/src/keymap/special_characters_and_unicode.md
```diff
@@ -0,0 +1,34 @@
...
+Entering special characters usually requires a key combination, which depends on your operating system and chosen keyboard layout (setting in the OS).
+For example, in macOS with an en-US layout, you can define the following sequence to enter a `ü`:
+```
+pub(crate) fn get_macro_sequences() -> [u8; MACRO_SPACE_SIZE] {
+    define_macro_sequences(&[
+        Vec::from_slice(&[
+            MacroOperation::Press(KeyCode::LAlt),
+            MacroOperation::Tap(KeyCode::U),
+            MacroOperation::Release(KeyCode::LAlt),
+            MacroOperation::Tap(KeyCode::U),
```
tib8: Is this final MacroOperation::Tap(KeyCode::U), really needed here?
Also LAlt really works, or RAlt is needed?
(I'm just asking, I'm on windows...)

patmuk:
@HaoboGu all is ready to merge :) I documented how one can enter unicode and language specific characters using macros.
This PR closes the issues #308, #284, #303, #313, #170 - and maybe others I overlooked :)

docs/src/keymap/special_characters_and_unicode.md
patmuk: Yes: The first Alt+U is a dead key for telling to put an umlaut and u is the letter where to put it on.
But I see that it is confusing ... will change it to Alt+U and A for ä.

RAlt works as well :)

docs/src/roadmap.md
```diff
@@ -22,6 +22,16 @@ There are a bunch of things to do with RMK in the near future. I plan to ship 1.
...
   - [x] 🟢 encoder
   - [x] 🔵 Input device
+  - [x] 🔴 Basic Macro support
+    - [x] 🔴 Macro definition via Rust
+    - [ ] 🟢 Macro definition via toml
+    - [ ] make macro storage space configurable
+    - [x] add KeyAction::TriggerMacro so more than 32 macros can be triggered
+    - [x] 🔴 Support umlaute
+    - [x] 🔵 Support unicode
+  - [x] Macro support enhancement
```
HaoboGu: Duplicate here?

HaoboGu:
oh you can also add this to the `CHANGELOG.md`

docs/src/roadmap.md
patmuk: Oh yes, thanks for catching that!
Fixed that and added a changelog entry as well :)

Btw: The open box for "make macro storage space configurable" can most likely be closed. I am talking the value from the global definition you refactored. However, I did not come around to test this.

HaoboGu:
thanks again!

patmuk:
You're very welcome :) Thanks for your great support!

Issue #131, opened 2024-11-09
Vial Matrix Tester seems cannot be used


mangkoran:
It seems Vial Matrix Tester is cannot be used. It doesn't respond to any input.

![image](https://github.com/user-attachments/assets/6b6f3081-a063-4448-9c84-3bb87b3e57e5)

Pro Micro RP2040 (16M flash)
rmk v0.4.1

HaoboGu:
good find, it's just not implemented yet :(

whitelynx:
@HaoboGu any idea what pieces of the protocol need to be implemented in order for this to work? I'm not that familiar with Vial's protocol, and so far I haven't found the specific command(s) related to this feature.

HaoboGu:
> @HaoboGu any idea what pieces of the protocol need to be implemented in order for this to work? I'm not that familiar with Vial's protocol, and so far I haven't found the specific command(s) related to this feature.

It's `SwitchMatrixState` event, the qmk implement [this way](https://github.com/vial-kb/vial-qmk/blob/4b2ae32e871c07e50703f0af591c479e66dc7eab/quantum/via.c#L249). However, it's hard to do it in Rust. I don't know if there's a easy way to implement this 🥲

tib888:
Is there a plan to support "NKRO"?
Because that (if I know right) NKRO also needs the state storage of every keys, not only 6 + modifiers.
But of course there is a difference: vial needs the phisical keys, while NKRO needs the "after mapped to FullSize" keys pressed state... 

Also, I was wondering why register_keycode() has this strange code inside:
```
        // First, find the key event slot according to the position
        let slot = self.registered_keys.iter().enumerate().find_map(|(i, k)| {
            if let Some((row, col)) = k {
                if key_event.row == *row && key_event.col == *col {
                    return Some(i);
                }
            }
            return None;
        });
```
Why is that dealing with row/col positions???
Maybe to use it for VIAL matrix tester?

pcasotti:
Adding a note here. Before we add support for vial matrix tester we need to implement the vial security feature, otherwise a malicious program can use the vial protocol to implement a keylogger

HaoboGu:
Closed by #516

mangkoran:
Sorry for necrobump. I can confirm this issue has been fixed with the PR above, which is released in [v0.8.0](https://github.com/HaoboGu/rmk/releases/tag/rmk-v0.8.0)

[...]

PR #340, opened 2025-04-15
feat/vial-lock(WIP): add security feature for vial

drindr:
Support vial's lock and unlock method

To avoid malicious program's behavior as @pcasotti mentioned in #131, the security feature of vial is required.
And I'd like to add a calibration method for the joystick with the extended vial, this feature should be implemented first as well for the safety reason.

### DIFF
--- a/rmk/src/via/vial.rs
+++ b/rmk/src/via/vial.rs
@@ -1,6 +1,8 @@
 use core::cell::RefCell;
 
 use byteorder::{BigEndian, ByteOrder, LittleEndian};
+use embassy_sync::{mutex::Mutex, signal::Signal};
+use heapless::Vec;
 use num_enum::FromPrimitive;
 
 use crate::action::KeyAction;
@@ -13,6 +15,7 @@ use crate::{
     channel::FLASH_CHANNEL,
     combo::COMBO_MAX_LENGTH,
     storage::{ComboData, FlashOperationMessage},
+    RawMutex,
 };
 
 /// Vial communication commands. Check [vial-qmk/quantum/vial.h`](https://github.com/vial-kb/vial-qmk/blob/20d61fcb373354dc17d6ecad8f8176be469743da/quantum/vial.h#L36)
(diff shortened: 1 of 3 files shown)


rmk/src/via/vial.rs
```diff
@@ -304,3 +330,74 @@ fn vial_combo_mut(combos: &mut heapless::Vec<Combo, COMBO_MAX_NUM>, idx: usize)
...
+        }
+    }
+    pub fn lock(&mut self) {
+        self.in_progress = false;
+        self.locked = true;
+    }
+}
+
+#[allow(dead_code)]
+pub struct MatrixStatus<const ROW: usize, const COL: usize> {
```
HaoboGu, line 360: Are MatrixStatus and VialStatus used only in vial? If that, I think VialStatus can be calculated directly in keyboard.rs for saving RAM.

Now we have to save a copy of MatrixStatus in Keyboard and convert it to VialStatus in Signal

HaoboGu: But MatrixStatus may be very useful when processing keys🤔

drindr: To save RAM, I think it is useful to implement all of them in one-bit-per-bool like `VialStatus.matrix_data`. Matrix is saved in a raw 2d bool array in `MatrixStatus` now. As for avoid duplication, maybe a static mutex works for that.
The reason why I put the `MatrixStatus` in the `keyboard.rs` now is just to ensure each key has been released normally before entering matrix test. Of course it might be useful for key processing.

HaoboGu: I'm considering to put `MatrixStatus` in `Keymap`, so the `VialService` could access the `MatrixStatus` just like reading keymap. No duplication and no `Signal`, but `Keymap` might need to be renamed LOL

drindr: I'm trying to wrap a static array of AtomicU8 and use each bit to save the status of each key. Maybe it can be aligned with Keymap by using the same or similar interface? Or just place the MatrixStatus in the Keymap module?

HaoboGu: Yes, I think so.

And there's some crates you can use, such as https://crates.io/crates/bit-vec and https://crates.io/crates/bitvec

drindr:
migrate to #516

PR #359, opened 2025-05-03
Use ble_addr in nrf split peripheral clear_storage

### LINKED ISSUE #355: Can the peripheral use the same ble_addr() approach as the central?
Of course, this is not released yet, but I was wondering if the peripheral could also use the same address generation approach as the central?

Specifically, I am referring to the following function:

```rust
fn ble_addr() -> [u8; 6] {
    let ficr = embassy_nrf::pac::FICR;
    let high = u64::from(ficr.deviceid(1).read());
    let addr = high << 32 | u64::from(ficr.deviceid(0).read());
    let addr = addr | 0x0000_c000_0000_0000;
    unwrap!(addr.to_le_bytes()[..6].try_into())
}
```

Reference: [peripheral.rs#L115-L115](https://github.com/HaoboGu/rmk/blob/ab9a28bcb887deeae7b7de81a4125b5a75811541/examples/use_rust/nrf52840_ble_split/src/peripheral.rs#L115C5-L115C64)

Thanks!

IniterWorker:
Close #355

### DIFF
--- a/examples/use_rust/nrf52840_ble_split/src/peripheral.rs
+++ b/examples/use_rust/nrf52840_ble_split/src/peripheral.rs
@@ -10,7 +10,7 @@ use embassy_nrf::gpio::{Input, Output};
 use embassy_nrf::interrupt::{self, InterruptExt};
 use embassy_nrf::peripherals::{RNG, SAADC, USBD};
 use embassy_nrf::saadc::{self, AnyInput, Input as _, Saadc};
-use embassy_nrf::{bind_interrupts, rng, usb, Peri};
+use embassy_nrf::{Peri, bind_interrupts, rng, usb};
 use nrf_mpsl::Flash;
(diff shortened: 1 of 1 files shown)


examples/use_rust/nrf52840_ble_split/src/peripheral.rs
```diff
@@ -129,6 +136,7 @@ async fn main(spawner: Spawner) {
     let storage_config = StorageConfig {
         start_addr: 0x60000, // 384K
         num_sectors: 32,     // 128K
+        clear_storage: true, // Ensure clean ble peer addr
```
HaoboGu: I prefer not to add this because it will make the peripheral clear saved central on every reboot. For new users it might be confusing.

This option should only be added when needed, for example, upgrading from old version to newer. Having a comment on it could be great.

HaoboGu (approved):
thanks!

Issue #347, opened 2025-04-24
RMK: Thoughts on Design, Direction, and Customization


IniterWorker:
Hey folks,

I've recently started working with RMK, writing firmware for my [Lily58](https://github.com/IniterWorker/rmk-lily58-keyboard). Like any good dev with questionable boundaries, I immediately went off the beaten path and started customizing things.

This post is mostly thoughts and feelings about where RMK is now and where it *could* go. It’s a bit philosophical, a bit technical, and definitely coming from someone who is new to keyboards' firmware and more into Rust. Let’s roll.

---

### The Setup: My Custom Lily58

So here’s my setup: the Lily58 has **63 LEDs**—58 under keys and 5 on the back panel. What I want is to create lighting effects **based on events**: user interactions, keyboard states, layer switches, whatever I can hook into.

In RMK today, the lighting system is fairly new, but it feels like it was designed around a “one-size-fits-all” model. For example:

```rust
pub(crate) struct LightService<'d, P: OutputPin, R: HidReaderTrait<ReportType = LedIndicator>> {
    pub(crate) enabled: bool,
    light_controller: &'d mut LightController<P>,
    reader: R,
}
```

It’s pin-based, non-async, and hard-wired into how RMK expects you to do lighting. Want to drive a bunch of WS2812s with an async external service? Not really in the cards.

I’m aiming for is something like this for a more Event Driven way with handler(s) that will allow me to write my own integration of the keyboard behaviour based on my hardware.

I want my LED logic to live outside RMK, in my keyboard integration repo—not inside the RMK core. And that brings me to the bigger topic…

---

### RMK vs QMK: A Shift in Philosophy

QMK is a beast. It’s taught us all how to get keyboards working, fast. But it’s built around **you configure what features you want**, not **you implement what your keyboard needs**. You dig through headers, tweak defines, enable stuff like you’re turning on modules in a spaceship.

And that works—until it doesn’t. Want something QMK doesn’t support? Time to fork and patch.

What I’d love to see from RMK is the opposite mindset: **you start with a clean core and plug in what you want**. Traits over globals. Events over conditionals. Modular crates instead of monolithic code.

---

### Where RMK Could Go (In My Ideal World)

Right now, RMK puts things like LED and display logic directly into the core. That’s fine if you’re building one universal firmware—but it makes real customization tricky.

Instead, imagine this layout:

```
// rmk
rmk          ← matrix scanning, debouncing, traits
rmk-macro         ← reusable keyboard concepts

[...]

PR #361, opened 2025-05-06
feat(controller): add `Controller` trait

pcasotti:
Creation off the `Controller` interface as discussed in #347, the interface should be similar to `InputDevice` with focus on user customization.


rmk/src/channel.rs
```diff
@@ -15,6 +15,9 @@ use {crate::split::SplitMessage, embassy_sync::pubsub::PubSubChannel};
 
 pub const EVENT_CHANNEL_SIZE: usize = 16;
 pub const REPORT_CHANNEL_SIZE: usize = 16;
+pub const CONTROLLER_CHANNEL_SIZE: usize = 16;
```
HaoboGu: Could you make these constants configurable? Just like https://github.com/HaoboGu/rmk/pull/351 does

HaoboGu (approved):

HaoboGu:
> I am wondering how it will work with split keyboard. Do you plan to run it only on central or peripheral will be able to listen to it ?

Both central and peripheral should be able to run the controller. The only difference of splits (I think) is the split driver should also be a subscriber of `ControllerEvent`, making `ControllerEvent` exchangeable between central and peripheral.

pcasotti:
I was thinking about adding a "main loop" for controllers, I believe some controllers (display, backlight) would probably run some kind of animation, which would need to be running constantly instead of only reacting to events.

```rust
pub trait Controller {
    async fn process_event(&mut self, event: ControllerEvent);
    async fn update(&mut self);
}
```
then on the macro we do
```rust
let sub = unwrap!($channel.subscriber());
loop {
    if let Some(e) = sub.try_next_message_pure().await {
        $dev.process_event(e).await,
    }
    $dev.update().await;
}
```

HaoboGu:
good idea! I think a "polling rate" might be needed to avoid controllers eats all CPU time. A sleep strategy should be considered as well.

pcasotti:
I added separate traits for event-driven and polling controllers, I think we don't need the `run_controllers!` macro anymore, that part is implemented on the trait.

HaoboGu:
But how do users run multiple different controllers(polling or event based) together?

pcasotti:
> But how do users run multiple different controllers(polling or event based) together?

By running their respective run functions, the implementation of the channel communication (creating the subscriber from the channel and receiving the messages) is left to the controller.

We can provide common controller implementations so the user can do:
```rust
let display = DisplayController::new(/*...*/);
let led = LedController::new(/*...*/);

join3(
    display.run_polling(),    // display polling events
    led.run_controller(),     // led event based
    run_rmk(
        // ...
    ),
)
.await;
```
If the user has unsupported hardware they can implement the traits for their own controller.

pcasotti:
A controller can implement both traits, as different users might want to run in different ways. In config we might have
```toml
[[controllers]]
type = "display"
run = "polling"
```
so we know when to use `run_polling` or `run_controller` (need to come up with a better name for these)

pcasotti:
What kinds of events would exist? We can maybe mirror some of zmk events
```
activity_state_changed.c
battery_state_changed.c
ble_active_profile_changed.c
endpoint_changed.c
hid_indicators_changed.c
keycode_state_changed.c
layer_state_changed.c
modifiers_state_changed.c
mouse_button_state_changed.c
position_state_changed.c
sensor_event.c
split_peripheral_status_changed.c
usb_conn_state_changed.c
wpm_state_changed.c
```

pcasotti:
I created a proof of concept [here](https://github.com/pcasotti/rmk/tree/feat/display-controller) of what a i2c display controller could look like.

rmk/src/controller/mod.rs
```diff
@@ -0,0 +1,174 @@
...
+    async fn process_event(&mut self, event: Self::Event);
+
+    /// Block waiting for next message
+    async fn next_message(&mut self) -> Self::Event;
+
+    /// Event loop
+    async fn run_controller(&mut self) {
+        loop {
+            let event = self.next_message().await;
+            self.process_event(event).await;
```
HaoboGu, line 54: An idea comes to my mind: maybe only a `EventController` is needed, and we can add `polling` mode for it.

Now the `EventController` has event loop: `run_controller`. For polling mode, just select between the event loop and the polling loop, like:

```rust
async fn run_controller_in_polling_mode(&mut self) {
    select(embassy_time::Timer::after(Self::INTERVAL), self.next_message()).await {
        First(_) => self.update().await;
        Second(event) => self.process_event(event).await; // Process event
    }
}
```

Then the `PollingController` can be:

```rust
pub trait PollingController: EventController {
    /// Interval in which `poll` and `update` will be called
    const INTERVAL: embassy_time::Duration;

    /// Update periodically
    async fn update(&mut self);

    /// Polling mode
    async fn run_controller_in_polling_mode(&mut self) {
       ...
    }
}
```

What do you think?

pcasotti: If I'm not wrong this implementation of `run_controller_in_polling_mode` will not work. If we have messages coming in intervals shorter than `Self::INTERVAL`, `update()` might never be called, we would need to subtract the elapsed time from INTERVAL and keep track of the remaining time. Also `process_event` in `EventController` might have a different implementation in `PollingController`.

I believe what we want is:
```
EventController:
    Event => save_state, update_peripheral
    
PollingController:
    Event => save_state
    INTERVAL => update_peripheral
```
It looks like we can use the same trait because the methods are the same, but I don't think so. `update_peripheral` can have different implementations in both traits (e.g. `PollingController` might update a frame of an animation in `update_peripheral`, while updating an animation in an `EventController` doesn't make sense). We could also share the `save_state`, but `EventController` might not want to store state.

The current implementation for `PollingController` isn't right either, it only looks for events in fixed intervals, while would be nice to receive events "immediately".

Maybe I'm overthinking it

HaoboGu: I think the difference between `EventController` and `PollingController` is how the `update` is triggered. `PollingController` has an extra trigger source, aka the timer. The function of the two traits can be shared, but the implementation depends.

I don't mean that `PollingController` "inherits" from `EventController`, in practice they are separately implemented for different devices. But they do have something similar. 

Maybe a single `Controller` trait is enough, let users/implementers choose whether their device is polling-based or event-based

pcasotti: But with a single `Controller` trait we would still have `INTERVAL` and `update` even if the user wants only an event-based controller, is there a way around it?

pcasotti: My only issue with `PollingController: EventController` is the `process_event` implementation, I believe they should have different implementations so `PollingController` would be:
```rust
pub trait PollingController: EventController {
    /// Interval in which `poll` and `update` will be called
    const INTERVAL: embassy_time::Duration;

    /// Update periodically
    async fn update(&mut self);

    /// Process the received event
    async fn process_event(&mut self, event: Self::Event);

    /// Polling mode
    async fn run_controller_in_polling_mode(&mut self) {
       ...
    }
}
```
What do you think?

HaoboGu: I don't think the `PollingController` should be a subtrait of `EventController`. It should be a separate trait.

I'm imagining something like the following:

```rust
trait Controller {
   type ControllerType;
   ... // Common fns
}

pub trait EventController: Controller<ControllerType = Self> {}

pub trait PollingController: Controller<ControllerType = Self> {
  const INTERVAL: embassy_time::Duration;
  async fn run_controller_in_polling_mode(&mut self) {
    ...
  }
  ... // Any other additional functions
}
```

pcasotti: That looks good, what would `ControllerType` be used for?

HaoboGu: I wanted to include the type of the controller(polling, event) to the trait. It seems that it's not easy, making it quite complex.. I want to keep this trait as simple as possible, so just ignore it :)

pcasotti: If we have `EventController: Controller {}` do we even need `EventController`?

pcasotti:
I think that's it. I'll create an issue to track implemented `ControllerEvent` sources.

rmk/src/controller/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+///     ),
+/// )
+/// .await;
+/// ```
+pub trait PollingController: Controller {
+    /// Interval between `update` calls
+    const INTERVAL: embassy_time::Duration;
+
+    /// Process the received event
+    async fn process_event(&mut self, event: Self::Event);
```
HaoboGu: I prefer to move `process_event` to `Controller`, even though they're usually differently implemented in `PollingController` and `EventController` -- that's why there're two different controller types based on the `Controller`.

rmk/src/controller/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+/// let c1 = MyController{};
+/// let c2 = MyController{};
+/// // Bind devices to channels
+/// let controller_future = run_controllers! {
+///     (c1, c2) => rmk::channel::CONTROLLER_CHANNEL,
+/// };
+///
+/// ```
+#[macro_export]
+macro_rules! run_controllers {
```
HaoboGu: This macro cannot be used for `PollingController`, since `PollingController` requires `update`, am I right?

rmk/src/controller/mod.rs
pcasotti: Should `EventController` and `PollingController` be mutually exclusive then?

rmk/src/controller/mod.rs
pcasotti: I was going to remove this macro but forgot, it's useless now we have `event_loop()` and `polling_loop()`.

rmk/src/controller/mod.rs
HaoboGu: I think so

PR #373, opened 2025-05-19
Allow disabling storage feature in cargo to work with a keyboard.toml

gabevenberg:


rmk-macro/src/keyboard.rs
```diff
@@ -42,7 +42,13 @@ pub(crate) fn parse_keyboard_mod(item_mod: ItemMod) -> TokenStream2 {
...
     if let Some(m) = toml_config.clone().matrix {
         if m.row2col {
-            eprintln!("row2col is enabled, please ensure that you have updated your Cargo.toml, disabled default features(col2row is enabled as default feature)");
+            eprintln!("row2col is enabled, please ensure that you have updated your Cargo.toml, disabling default features(col2row is enabled as default feature)");
+        }
+    }
+
+    if let Some(s) = toml_config.clone().storage {
+        if !s.enabled {
+            eprintln!("storage is disabled, please ensure that you have updated your Cargo.toml, disabling default features(storage is enabled as default feature)");
```
HaoboGu: I think we can check whether `keyboard_config.storage.enabled == is_feature_enabled(rmk_features, "storage")` here, if not, just panic with a message.

Then, when expanding code, only `keyboard_config` is needed, and `rmk_features` doesn't need to be passed everywhere

gabevenberg: Id perhaps argue that then we shouldn't have an `enabled` flag at all, and leave it all up to the cargo feature...

HaoboGu: 🤔 hmmm, you're correct, at least in `StorageConfig`, the enabled field can be removed. But when generating RMK projects from a `keyboard.toml`, this option is still valuable for determining whether the `storage` feature is enabled.

I think we can do it in a separate PR

gabevenberg: For now, should I be checking based off the cargo feature or the toml field?

HaoboGu: I'd like to make the macro check the toml field for now ;)

gabevenberg:
I will say I have a bit of an issue with how default features are being used, whenever one wants to disable just one default feature, they have to figure out what the other default features are, and add them as features manually. Ill do a write up in an issue.

HaoboGu:
Is this PR finished? It looks good now to me.

gabevenberg:
Yup, its finished.

PR #386, opened 2025-05-29
docs: Improve the style of the code block and update the style of the custom block

Menra11:
docs: Improve the style of the code block and update the style of the custom block


HaoboGu:
nice work, thanks

PR #394, opened 2025-06-04
doc(layout): fix mouse key code in the example

lotem:
Error building the example layout config

    error[E0599]: no variant or associated item named `MouseBtn12` found for enum `KeyCode` in the current scope


HaoboGu:
Thanks ❤️

PR #402, opened 2025-06-07
fix(controller): update modifiers correctly

pcasotti:
Currently modifier changed events are only sent together with the HID report, this can cause incorrect behavior for intermediate modifiers that happen between HID reports (one shot, etc.)


HaoboGu:
lgtm, thanks!

PR #403, opened 2025-06-08
feat: move controller behind a feature flag

pcasotti:
I believe it's a good idea to move controller stuff behind a feature flag since it might increase latency when sending events. We might also want per-event features in the future.


HaoboGu:
moving controllers behind a feature gate is good. but using feature to set the supported event is too complex for users.

HaoboGu:
thanks!

PR #413, opened 2025-06-13
Feat/controller event

### LINKED ISSUE #366: Implement controller event sources
Now we have the `Controller`, we need to actually send [ControllerEvent](https://github.com/HaoboGu/rmk/blob/3d68f036a4cdc3da5c2ec37aa890a336474f74f6/rmk/src/event.rs#L101) messages to the controllers.

- [ ] Key event
- [ ] Battery percent changed
- [ ] Changing state changed
- [ ] Ble profile changed
- [ ] Layer changed
- [X] Modifier changed
- [ ] Typing speed
- [ ] Connection type
- [ ] Split peripheral connection status

pcasotti:
Closes #366


pcasotti:
- [X] Key event
- [x] Battery percent changed
- [x] Changing state changed
- [x] Ble profile changed
- [x] Layer changed
- [X] Modifier changed
- [x] Typing speed
- [x] Connection type
- [x] Split peripheral connection status

pcasotti:
I wonder if wpm calculations should be implemented as a `Controller` that can also send events

HaoboGu:
> I wonder if wpm calculations should be implemented as a `Controller` that can also send events

looks reasonable. but I don't know how it impacts ram usage, extra publisher and subscriber are added in this case?

HaoboGu:
and is it possible to calculate the number of publishers and subscribers at the compile time?

pcasotti:
> looks reasonable. but I don't know how it impacts ram usage, extra publisher and subscriber are added in this case?

Yes, there will be an extra publisher and subscriber, but from what I can tell they're just wrappers around the channel.

> and is it possible to calculate the number of publishers and subscribers at the compile time?

The best I can think of is to manually keep track of how many we use internally, and add that to user defined controllers (via `constants.rs`). From reading the source code, looks like there isn't much impact in using extra publishers/subscribers, it just dictates the size of an internal `heapless::Vec`

HaoboGu:
> Yes, there will be an extra publisher and subscriber, but from what I can tell they're just wrappers around the channel.

Sounds great!

> The best I can think of is to manually keep track of how many we use internally, and add that to user defined controllers (via constants.rs)

For built-in controllers, yeah I think it's possible to track how many controllers are enabled in `keyboard.toml` and then calculate the number of used publisher/subscriber. It's not only about the ram usage, it causes the run-time panic or hanging(when the number of pub/sub reaches the limit). So it would be very nice if we could have some way to detect it.

rmk/src/keyboard.rs
```diff
@@ -227,6 +227,10 @@ impl<'a, const ROW: usize, const COL: usize, const NUM_LAYER: usize, const NUM_E
         if self.combo_on {
             if let Some(key_action) = self.process_combo(key_action, key_event).await {
                 debug!("Process key action after combo: {:?}, {:?}", key_action, key_event);
+
+                #[cfg(feature = "controller")]
+                send_controller_event(&mut self.controller_pub, ControllerEvent::Key(key_event));
```
EthanOlpin: question: would it be reasonable to include the `KeyAction` with `Key` events? Currently it seems like it could be difficult for consumers to track/determine the relevant state if they, for example, want to display typed keys on an OLED.

pcasotti: Good question, if we do this we could maybe remove layer and modifier changed events

rmk/src/lib.rs
```diff
@@ -195,18 +197,29 @@ pub async fn run_rmk<
     light_controller: &mut LightController<Out>,
     rmk_config: RmkConfig<'static>,
 ) -> ! {
+    #[cfg(feature = "controller")]
```
pcasotti: @HaoboGu Is this the right place to initialize the controller?

HaoboGu: I think the controller initialization should be controlled by users's `main.rs`, just like input devices. For toml config users, the controller initialization code is generated by rmk-macro according to the config. It also makes it possible for users to write their own controllers.

The drawbacks is that it makes the user-side code quite complex. I'm not sure whether this way is clear enough(for both users and developers), what do you think about it?

pcasotti: I imagined some controllers might be mandatory like "you need to have the `WpmController` if the `controller` feature is active, or if we implement `VialService` as a controller, while the user could manage his own controllers himself.

pcasotti: I guess we can leave this out for now, I'll drop this commit.

HaoboGu:
yeah, the mixed way seems acceptable for controllers which have no user-side configuration. But for controllers like LightController or RgbController, they require users pass additional pin config to run_rmk, which is not easy to manage

pcasotti:
> yeah, the mixed way seems acceptable for controllers which have no user-side configuration. But for controllers like LightController or RgbController, they require users pass additional pin config to run_rmk, which is not easy to manage

So we keep `WpmController` inside `run_rmk`? I'll move it to `run_keyboard` then, seems like a better place

HaoboGu:
> So we keep `WpmController` inside `run_rmk`? I'll move it to `run_keyboard` then, seems like a better place

TBH I'm not 100% sure it's a good choice, because for those built-in controllers, we have to use feature gate to control whether a controller is enabled or not. When the number of built-in controller increases, the number of feature gates increases as well, making features hard to manage.

But for now, we don't have many build-in controllers, it looks good for me. So let's do it.

rmk/src/lib.rs
```diff
@@ -285,20 +287,27 @@ pub(crate) async fn run_keyboard<
...
+    let mut wpm_controller = WpmController::new();
+
     match select4(
         communication_task,
         #[cfg(feature = "storage")]
         select(storage_fut, via_fut),
         #[cfg(not(feature = "storage"))]
         via_fut,
+        #[cfg(feature = "controller")]
```
HaoboGu, line 300: Now enabling `controller` automatically enables `wpm_controller`, is it right?

pcasotti: yes

pcasotti: In the future this could be optional when we start to work on configuration for controllers.

HaoboGu:
nice work, thanks!

PR #430, opened 2025-07-01
Add Delay for Buffered Key Events to Ensure HID Recognition

hitsmaxft:
Description:
In certain scenarios, key events are sent too quickly for the HID layer to reliably register state changes, especially when using buffered input behaviors like tap-hold or combos. This patch introduces a minimum delay (e.g., 1ms) to ensure that HID devices have enough time to process key transitions.
Problem:

For example:

    A press event on key A is queued in the event buffer.

    After just 10ms, a release event for the same key A arrives.

    Tap-hold logic detects this and flushes the buffer, sending a tap action for key A.

If this transition (press → release ) happens too quickly, the HID system may miss the state change and fail to report the key correctly.
Solution:

    Introduce a manual delay (at least 1ms) during key buffering (e.g., tap-hold and combo behaviors).

    Ensure that any emitted key event has a visible time gap before the next state change is sent.

Testing:

    The test module now verifies that all reported key events include a minimum inter-event delay of 1ms to ensure HID stability.

### DIFF
--- a/rmk/src/keyboard.rs
+++ b/rmk/src/keyboard.rs
@@ -22,7 +22,7 @@ use core::cmp::Ordering;
 use core::fmt::Debug;
 
 use embassy_futures::select::{select, Either};
-use embassy_futures::yield_now;
+use embassy_futures::{join, yield_now};
 use embassy_time::{Duration, Instant, Timer};
 use heapless::{Deque, FnvIndexMap, Vec};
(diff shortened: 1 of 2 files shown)


HaoboGu:
TBH I suspect that it's not a HID issue. As long as the `pressed` and `released` report are sent, even if they are fired within 1ms, they will be queued and sent to the host(maybe in the adjacent HID report), so no HID report will be skipped.

Issue #428, opened 2025-06-30
Permissive hold key is lost


lotem:
Modified https://github.com/lotem/rmk-sweep
to build with https://github.com/HaoboGu/rmk/commit/531d80c8fe724e708a5c9db5efcb69d03d78b234
with configuration

```toml
[behavior]
tap_hold = { permissive_hold = true }
```

Key sequences like these can output nothing.

    MT(LShift, Z) down
    S down
    S up
    MT(LShift, Z) up, this event is optional for reproducing the issue

Expected: uppercase `S`.
Actual: nothing if `S` is typed quickly after MT; "quickly" <= the default `hold_timeout` (250ms).

Not reproducible if `S` is typed after a longer delay.
Not reproducible with LT.

---

With `tap_hold.mermissive_hold = false` (default value), the same key sequence outputs lowercase `s` while expecting uppercase `S`.

HaoboGu:
I can repro it using the following test:

```rust
#[test]
fn test_tap_hold_quick_tap_hold() {
    key_sequence_test! {
        keyboard: create_test_keyboard_with_config(BehaviorConfig {
            tap_hold: tap_hold_config_with_hrm_and_permissive_hold(),
            .. BehaviorConfig::default()
        }),
        sequence: [
            [2, 1, true, 10], // Press th!(A, lshift)
            [2, 3, true, 200],  // Press D
            [2, 3, false, 100], // Release D  <-- Release D after "permissive hold" interval, but also after the hold-timeout
            [2, 1, false, 100], // Release A
        ],
        expected_reports: [
            [KC_LSHIFT, [0, 0, 0, 0, 0, 0]], // Hold LShift
            [KC_LSHIFT, [kc_to_u8!(D), 0, 0, 0, 0, 0]], // Press D
            [KC_LSHIFT, [0, 0, 0, 0, 0, 0]], // Release D
            [0, [0, 0, 0, 0, 0, 0]], // All released
        ]
    };
}

// Report output of the above test:
[KC_LSHIFT, [0, 0, 0, 0, 0, 0]],
[KC_LSHIFT, [0, 0, 0, 0, 0, 0]],
[0, [0, 0, 0, 0, 0, 0]],
```


Changing the third action to `[2, 3, false, 50]` makes the testing pass.

lotem:
Fixed. Thanks.

PR #431, opened 2025-07-01
fix: tap hold timeout should fire other key pressing

hitsmaxft:
Fix issue #428 

Tap Hold timeout is not process properly and leak following single key. 

I combine timeout and permissive hold into same fire_all_holding_keys, and making it became more complex.


HaoboGu:
great work, thanks!

PR #438, opened 2025-07-05
fix: improve pest parser rule to handle "No" action with trailing whitespace

hitsmaxft:
- Changed no_action rule to atomic (@) to capture only "No" without consuming following characters
- Added comprehensive test cases for "No" action parsing
- Ensures "No " (with space) and "No\n" (with newline) are correctly parsed as no_action
- Distinguishes between "No" action and keycodes like "NoUsSlash" or "NonUsSlash"


HaoboGu:
Thanks!

PR #439, opened 2025-07-05
fix: No with whitespace at the end of layer.keys

### LINKED ISSUE #437: compile error: `No` followed by whitespace at the end of `layer.keys`
tested: rmk 0.7.4

reproducible code:
https://github.com/lotem/rmk-example/commit/792463d8d6161194e7b7ec76c3ccecd3e850f3d0

build log:

```shell
❯ cargo build --release
warning: Patch `trouble-host v0.1.0 ([https://github.com/embassy-rs/trouble.git?rev=cb4b2cd424cdb2af54bad52922798a58d8786d7f#cb4b2cd4)`](https://github.com/embassy-rs/trouble.git?rev=cb4b2cd424cdb2af54bad52922798a58d8786d7f#cb4b2cd4)%60) was not used in the crate graph.
Check that the patched package version and available features are compatible
with the dependency requirements. If the patch has a different version from
what is locked in the Cargo.lock file, run `cargo update` to use the new
version. This may also occur with an optional dependency that is not enabled.
   Compiling rmk v0.7.4
   Compiling rmk-example v0.2.0 (/Users/chengong/src/rmk-example)
error: custom attribute panicked
 --> src/main.rs:8:1
  |
8 | #[rmk_keyboard]
  | ^^^^^^^^^^^^^^^
  |
  = help: message: `"No\n"` is not a valid identifier

error: `#[panic_handler]` function required, but not found

error: could not compile `rmk-example` (bin "rmk-example") due to 2 previous errors
```

lotem:
Fixes #437

`no_action` should be an atomic rule.
Maybe the lookahead expression was hindered by WHITESPACE. https://docs.rs/pest/latest/pest/#whitespace-and-comment

### DIFF
--- a/rmk-config/src/keymap.pest
+++ b/rmk-config/src/keymap.pest
@@ -41,7 +41,7 @@ modifier_combination = { modifier_name ~ ("|" ~ modifier_name)* }
 simple_keycode = { keycode_name }
 
 // Rule 2: No Key
-no_action = { ^"No" ~ !ASCII_ALPHANUMERIC } // Case-insensitive "No" followed by non-alphanumeric
+no_action = @{ ^"No" ~ !ASCII_ALPHANUMERIC } // Case-insensitive "No" followed by non-alphanumeric
 
 // Rule 3: Transparent Key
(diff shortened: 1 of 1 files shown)


HaoboGu:
Thanks

PR #464, opened 2025-07-09
move KeyAction::WithModifier to Action::KeyWithModifier

alex-tdrn:
Marked this as draft since I'm going to sleep now and haven't had the time to actually test the changes with my layout. 
But I opened the PR to start discussion about 2 things I'm unsure of:
1. the `Action` and `KeyAction` serialization to u16 methods were tricky to update, so I thought I'd comment them out until later. Then I noticed that they were only depended upon by one test case, so I removed them completely in the first commit (I updated the test case to use the enums directly). What is the purpose of the u16 representation? If needed, I can revert the commit and add them back, but I think they would need to be extended to u32. Is there some future plan for these functions, and if so what constraints are there?
2.I hit a snag and realized that the `lm!` macro cannot work with my change. Not sure how to fix that one, except adding something like `KeyAction::Double(Action, Action)` but that seems very smelly haha. Maybe there's a different arrangement of the 2 enums that covers all use-cases 🤔

I'll try to continue working on this tomorrow after work again, or on the weekend at the latest ✌️

### DIFF
--- a/rmk/src/action.rs
+++ b/rmk/src/action.rs
@@ -56,8 +56,6 @@ pub enum KeyAction {
     Tap(Action),
     /// Keep current key pressed until the next key is triggered.
     OneShot(Action),
-    /// Action with the modifier combination triggered.
-    WithModifier(Action, ModifierCombination),
     /// General tap/hold action: (tap_action, hold_action)
     TapHold(Action, Action),
     /// Tap dance action, references a tap dance configuration by index.
@@ -72,6 +70,8 @@ pub enum Action {
     Key(KeyCode),
     /// Modifier Combination, used for oneshot keyaction.
     Modifier(ModifierCombination),
+    /// Key stroke with modifier combination triggered.
+    KeyWithModifier(KeyCode, ModifierCombination),
     /// Activate a layer
     LayerOn(u8),
     /// Deactivate a layer
--- a/rmk/src/action.rs
+++ b/rmk/src/action.rs
@@ -74,6 +74,8 @@ pub enum Action {
     KeyWithModifier(KeyCode, ModifierCombination),
     /// Activate a layer
     LayerOn(u8),
+    /// Activate a layer with modifier combination triggered.
+    LayerOnWithModifier(u8, ModifierCombination),
     /// Deactivate a layer
     LayerOff(u8),
     /// Toggle a layer
--- a/rmk/src/action.rs
+++ b/rmk/src/action.rs
@@ -43,16 +43,6 @@ impl EncoderAction {
 
 /// A KeyAction is the action at a keyboard position, stored in keymap.
 /// It can be a single action like triggering a key, or a composite keyboard action like tap/hold
-///
-/// Each `KeyAction` can be serialized to a u16 action code. There are 2 patterns of action code's bit-field composition of `KeyAction`:
-///
-/// - KeyActionType(8bits) + BasicAction(8bits)
-///
-/// - KeyActionType(4bits) + Action(12bits)
-///
-/// The `BasicAction` represents only a single key action of keycodes defined in HID spec. The `Action` represents all actions defined in the following `Action` enum, including modifier combination and layer switch.
-///
-/// The KeyActionType bits varies between different types of a KeyAction, see docs of each enum variant.
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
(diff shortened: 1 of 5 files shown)


HaoboGu:
> What is the purpose of the u16 representation? If needed, I can revert the commit and add them back, but I think they would need to be extended to u32. Is there some future plan for these functions, and if so what constraints are there?

Using u16 as the representation is due to some legacy reason and memory usage consideration, updating to u32 is fine for me.


> I hit a snag and realized that the lm! macro cannot work with my change. Not sure how to fix that one, except adding something like KeyAction::Double(Action, Action) but that seems very smelly haha. Maybe there's a different arrangement of the 2 enums that covers all use-cases 🤔

Add an action like `Action::LayerWithModifier(u8, ModifierCombination)`?

alex-tdrn:
I'm struggling a with updating `to_action_code` and `to_key_action_code`. I don't want to be annoying, but I'll ask again, what are these functions used for? It seems like it's leftover code from a time when this was the storage format used instead of the vial format? What is the argument for not just removing these functions?

Part of what's breaking my brain a bit is that with adding `Action::KeyWithModifier` and `Action::LayerOnWithModifier`, there's now a lot of overlap between the enum variants, and I keep getting it wrong. This is tricky code to get right (at least for me, but I'm also not used to bit twiddling to this degree). 

Even the current implementation has some bugs. For example, in `to_key_action_code`, serializing a `KeyAction::OneShot(Action::Key(KeyCode::M))` would be ambiguous with `KeyAction::OneShot(Action::Key(KeyCode::No))` and `KeyAction::Single(Action::Key(KeyCode::M))`. Another example is that `KeyAction::{Transparent, Single, Tap, OneShot}` are all ambiguous between each other. 
```rust
pub(crate) fn to_key_action_code(self) -> u16 {
        match self {
            KeyAction::No => 0x0000,
            KeyAction::Transparent => 0x0001, // <--- overlap
            KeyAction::Single(a) => a.to_action_code(), // <--- overlap
            KeyAction::Tap(a) => 0x0001 | a.to_action_code(), // <--- overlap
            KeyAction::OneShot(a) => 0x0010 | a.to_action_code(), // <--- overlap
            KeyAction::WithModifier(a, m) => 0x4000 | ((m.into_bits() as u16) << 8) | a.to_basic_action_code(),
            KeyAction::TapHold(tap, hold) => match hold {
                Action::LayerOn(layer) => {
                    if layer < 16 {
                        0x3000 | ((layer as u16) << 15) | tap.to_basic_action_code()
                    } else {
                        error!("LayerTapHold supports only layer 0~15, got {}", layer);
                        0x0000
                    }
                }
                Action::Modifier(m) => 0x6000 | ((m.into_bits() as u16) << 8) | tap.to_basic_action_code(),
                _ => 0x8000 | (hold.to_basic_action_code() << 15) | tap.to_basic_action_code(),
            },
            KeyAction::TapDance(index) => 0x5700 | (index as u16),
        }
    }
```
Even after updating the serialization, there's no guarantee that it's not broken in some way, since it's not used by any code (yet?).

I'm proposing to remove these, as in my first commit, since they are a maintenance burden with no benefit as of right now. If RMK will need its own binary format in the future, we can develop it then, with the advantage that we will know what constraints we need to respect, and with unit tests that check whether the serialization is correct.  @HaoboGu would that be acceptable to you?

HaoboGu:
yes, I agree. Removing them is fine. Most of them is legacy code that is not used anymore.

alex-tdrn:
> It looks good to me now. For LM!, I'm fine to leave it to the next PR, or you can also add it in this PR if you want.

Ah, I would definitely add LM as well, I wouldn't want anyone's config to break if they're following master. I was still in the process of adding that and testing on my keyboard today. But I ran into https://github.com/HaoboGu/rmk/issues/470 along the way and got distracted a bit haha. But I'm realizing I can properly test this PR without involving tap-hold. I'll definitely wrap this PR up before the weekend's over though! 🤞

alex-tdrn:
Okay this is ready now. I tested my change by testing simple shifted keys on my layout, and they work fine. Didn't test for `lm!` changes but I'm pretty confident this should be right

HaoboGu:
Could you fix the warning in CI?

alex-tdrn:
> Could you fix the warning in CI?

Should be fixed now 👍

HaoboGu:
Great, thanks a lot!

PR #480, opened 2025-07-16
update the problem of color occlusion, layout size and logo size

Menra11:
- update the problem of color occlusion
- reduce the layout size for large screens
- update logo size and remove the 'rmk' text of header


HaoboGu (approved):
thanks a lot！

PR #483, opened 2025-07-17
fix: overflow in `PollingController::polling_loop`

pcasotti:
Also missing import in `event.rs`, preventing compilation with `controller` feature


HaoboGu:
thanks!

PR #484, opened 2025-07-17
Battery report changes for controller

pcasotti:
These changes aim to improve battery reporting for controllers:

* Report the battery percentage instead of the measured adc value
* Report battery instantly after boot, without waiting for `polling_interval` in `NrfAdc`


HaoboGu (approved):
lgtm, thanks!

PR #487, opened 2025-07-19
feat: add macro config to toml

### LINKED ISSUE #284: Macro configuration

pcasotti:
Closes #284

for now it looks like this:
```toml
[[behavior.macros.macros]]
operations = [
    { operation = "down", keycode = "LSFT" },
    { operation = "text", text = "hello world1" },
    { operation = "up", keycode = "LSFT" },
]

[[behavior.macros.macros]]
operations = [
    { operation = "tap", keycode = "LCTL" },
    { operation = "tap", keycode = "LALT" },
    { operation = "tap", keycode = "DEL" },
]

[[behavior.macros.macros]]
operations = [
    { operation = "text", text = "ding!" }
]

[[behavior.macros.macros]]
operations = [
    { operation = "tap", keycode = "F1" },
    { operation = "delay", duration = "1000ms" },
    { operation = "tap", keycode = "PGDN" },
]
```


HaoboGu:
Maybe using `[behavior.macro.macros]` as the key is better? 

Edit: Now there's no other configurations for macro, so using `[behavior.macros]` is also fine. But if there is other configuration other than `macros` config, such as `tap interval of adjacent operation`, then the `[[behavior.macros]]` is not sufficient. What do you think?

pcasotti:
> Maybe using `[behavior.macro.macros]` as the key is better?

Yes, this looks better

rmk-config/src/lib.rs
```diff
@@ -346,6 +346,8 @@ pub struct BehaviorConfig {
     pub tap_hold: Option<TapHoldConfig>,
     pub one_shot: Option<OneShotConfig>,
     pub combo: Option<CombosConfig>,
+    #[serde(alias = "macro")]
+    pub macros: Option<MacrosConfig>,
```
HaoboGu, line 415: is there any particular reason for not using `pub macro: ..`?

pcasotti: It's reserved: `expected identifier, found reserved keyword`.
it could be defined as `pub r#macro:`, but then we need the `r#` notation everywhere to access it, e.g., `behavior.r#macro`

HaoboGu:
lgtm! Could you add some documentation about it?

PR #498, opened 2025-07-23
Fix permissive hold key rolling

### LINKED ISSUE #496: Conflicting rolling key issues with permissive_hold
Below are two scenarios that I am facing with my current layout using RMK. I'm replicating my current ZMK config and both scenarios work as expected for me on ZMK.

## Scenario 1 (only happens when `permissive_hold = true`)

I would like to type **"a+"**.

`KC_A` is on layer 0 (default), and `KC_KP_PLUS` is on layer 1. Activating layer 1 is done by holding `LT1(KC_SPACE)`.

I press `KC_A` > `LT1(KC_SPACE)` > `KC_KP_PLUS`

When I release `KC_A` before pressing `LT1(KC_SPACE)`, all goes well and I get **"a+"** at the end. However, if I press `LT1(KC_SPACE)` before releasing `KC_A`, I get the tap behavior even though I haven't released `LT1(KC_SPACE)`, ending up with **"a b"** (since the key for `KC_KP_PLUS` is a `KC_B` on layer 0).

## Scenario 2 (only happens when `permissive_hold = false`)

I would like to type **" b"**.

`KC_B` is on layer 0 (default). Space is sent by tapping on `LT1(KC_SPACE)`.

I press `LT1(KC_SPACE)` > `KC_B`

When I release `LT1(KC_SPACE)` before pressing `KC_B`, all goes well and I get **" b"** at the end. However, if I press `KC_B` before releasing `LT1(KC_SPACE)`, I end up with **"b "** instead of **" b"**. I believe this happens because `KC_B` is not a modkey so it computes on press, whereas `LT1(KC_SPACE)` computes the tap on release (it waits to check if it was a tap or a hold).

## Summary

I don't think Scenario 2 is a bug, since the behavior I want is the described at https://docs.qmk.fm/tap_hold#rolling-keys with PERMISSIVE_HOLD turned on, so it makes sense that it would not happen when I use `permissive_hold = false`.

Scenario 1 looks like a bug to me. The keypress of a non-modkey should not influence the tap-hold decision of a key pressed afterwards. If not a bug, it's at least something the user should be able to toggle on or off in their config. Please let me know if this already exists and I missed it.

Apart from that one issue, I'm able to daily drive a 36-key perfectly. VIA + mouse commands + wireless written in Rust, truly awesome.

HaoboGu:
Fixes #496

### DIFF
--- a/rmk/tests/keyboard_taphold_test.rs
+++ b/rmk/tests/keyboard_taphold_test.rs
@@ -741,5 +741,31 @@ mod tap_hold_test {
             };
         }
 
+        #[test]
+        fn test_issue_496() {
+            key_sequence_test! {
+                keyboard: {
+                    let behavior_config = BehaviorConfig {
+                        tap_hold: tap_hold_config_with_hrm_and_permissive_hold(),
+                        ..BehaviorConfig::default()
+                    };
+                    create_test_keyboard_with_config(behavior_config)
+                },
+                sequence: [
+                    [1, 1, true, 20],  // Press Q
+                    [4, 5, true, 150],  // Press lt!(1, space)
+                    [1, 1, false, 20],  // Release Q
+                    [1, 3, true, 10], // Press E(on layer 1 is W)
+                    [1, 3, false, 10], // Release E(should trigger W)
+                    [4, 5, false, 10], // Release lt!(1, space)
+                ],
+                expected_reports: [
+                    [0, [kc_to_u8!(Q), 0, 0, 0, 0, 0]], // Tap Q
+                    [0, [0, 0, 0, 0, 0, 0]], // All released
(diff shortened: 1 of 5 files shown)


rmk/tests/keyboard_taphold_test.rs
```diff
@@ -741,5 +741,31 @@ mod tap_hold_test {
             };
         }
 
+        #[test]
+        fn test_issue_496() {
```
guicostaarantes, line 745:
```suggestion
        fn test_taphold_with_previous_rolling_keypress() {
```

HaoboGu: cool, fixed in another commit :)

PR #506, opened 2025-07-24
fix: outdated version of `embassy-usb-logger`

pcasotti:
This prevented compilation with the `usb_log` feature due to incompatible versions with `embassy-usb`


HaoboGu:
thanks

PR #511, opened 2025-08-02
Update dependencies

HaoboGu:
- [x] embassy-executor
- [ ] embassy-executor for esp32, upstream PR: https://github.com/esp-rs/esp-hal/pull/3889
- [ ] embassy-futures
- [x] sequential-storage
- [x] nrf-sdc and nrf-mpsl
- [x] embassy-stm32
- [ ] bt-hci


HaoboGu:
I'll merge it as is, then open another PR for other deps

Issue #490, opened 2025-07-21
Proposal: Tap Hold Dance


HaoboGu:
## Definition

While implementing tap-dance(#479), I realized that there's a chance to merge tap-hold and tap-dance to a more powerful behavior -- it's call it tap-hold-dance（forgive me for the poor naming).

The tap-hold-dance key works like this:

- Tap a key multiple times to trigger different actions (just like tap-dance).
- After tapping, you can hold the key to trigger another action — the hold action changes based on how many times you tapped before.

The representation is actually same with current TapDance behavior:

```rust
struct TapHoldDance {
  tap_actions: [KeyAction; MAX_TAP_NUM],
  hold_actions: [KeyAction; MAX_TAP_NUM],
  tapping_term: Duration,
}
```

When the `MAX_TAP_NUM` = 1, it fallbacks to a normal tap-hold action, and when `hold_actions` is empty, it fallbacks to tap-dance action.

## Limitations

It sounds great, but it has some limitations still.

1. In Rust's type system, the `MAX_TAP_NUM` should be known at the compile time, and all `TapHoldDance` should have same length of `MAX_TAP_NUM`. That increases the memory usage A LOT, because if you have a tap-dance behavior with `MAX_TAP_NUM = 4`, then ALL normal tap-hold behavior requires 4x memory which will never been used.
2. Now tap-dance implementation accepts `KeyAction` as the behavior, that's why it's **u8 indexed**:
    ```rust
    enum KeyAction {
      ...
      TapDance(u8)
    }
    ```
    To avoid recursion we have to use the u8 as the index and find the corresponding `TapDance` from the `BehaviorConfig`. 

    Of course we can use `Action` in `tap_actions` and `hold_actions`, if we do that the `OneShot` keys(which is now a `KeyAction`) cannot be used in tap-hold-dance. It's lots work to do to migrate `OntShot` to `Action` too.
3. There are many options available to tap-hold, for example, `prior_idle_time`, `permissive_hold`, `chordal_hold`, etc. Those behaviors improves tap-hold keys. I'm not sure whether they can be kept easily for tap-hold-dance behavior.

tib8:
Would it be possible to configure the tap pattern like this?

struct Morse {
  actions: [(str&, KeyAction); MAX_MORSE_ACTION_NUM],  
  tapping_term: Duration,
}

Example:
(".", E) //tap
("-", T) //hold
("..", I) //double tap
(".-", A) //tap hold
[...]

Issue #512, opened 2025-08-02
Introduce a configuration matrix for key position specific information


tib888:
It could hold information about the location of the home row, the left/right hand, and maybe timing settigs, which all can be useful for fine tuning HRM. 
Other ideas are wellcome!

My idea is to config these in the optional 'third coordinate' of the matrix_map:

```
matrix_map = """
(2, 0) (2, 1) (1, 2) (1, 3) (1, 4) (1, 5)              (1, 6) (1, 7) (1, 8) (1, 9) (2, 10) (2, 11)
(3, 0) (3, 1) (2, 2) (2, 3) (2, 4) (2, 5)              (2, 6) (2, 7) (2, 8) (2, 9) (3, 10) (3, 11)
(4, 0) (4, 1) (3, 2) (3, 3) (3, 4) (3, 5)              (3, 6) (3, 7) (3, 8) (3, 9) (4, 10) (4, 11)
                           (4, 2) (4, 3) (4, 4)  (4, 7) (4, 8) (4, 9)
"""
```

'L'/'R' to set the sides, upper case for the home row, lower case elsewhere:
```
matrix_map = """
(2, 0, l) (2, 1, l) (1, 2, l) (1, 3, l) (1, 4, l) (1, 5, l)              (1, 6, r) (1, 7, r) (1, 8, r) (1, 9, r) (2, 10, r) (2, 11, r)
(3, 0, L) (3, 1, L) (2, 2, L) (2, 3, L) (2, 4, L) (2, 5, L)              (2, 6, R) (2, 7, R) (2, 8, R) (2, 9, R) (3, 10, R) (3, 11, R)
(4, 0, l) (4, 1, l) (3, 2, l) (3, 3, l) (3, 4, l) (3, 5, l)              (3, 6, r) (3, 7, r) (3, 8, r) (3, 9, r) (4, 10, r) (4, 11, r)
                                     (4, 2, l) (4, 3, l) (4, 4, l)  (4, 7, r) (4, 8, r) (4, 9, r)
"""
```

HaoboGu:
Adding "l/r" for tagging hands looks good to me, but I don't like using capitalized L/R for HRMs, because `l` and `L` in the matrix map are hard to distinguish

I prefer something like 

```toml
hrm_keys = "a s" # or ["A", "S"] or "mt!(A, LShift) mt!(S, LGui)`
```

or 

```toml
matrix_map = """
[...]

PR #514, opened 2025-08-03
Memory efficient morse

tib888:
WIP:
- Needs testing on HW

- KeyState improvement is needed to store the encoded morse pattern - see https://github.com/HaoboGu/rmk/issues/490#issuecomment-3129148302

- The Morse struct should be refactored: instead of ta,hold arrays lets have a single action map
pub struct MorseTable<const N: usize> {
    /// array of valid pairs: (encoded morse pattern, action to be triggered)
    pub(crate) actions: Vec<(u8, Action); N>
    ...
}

- The advanced morse behavior, which can be accessed by morse!(index) macro (just like td!) needs toml config support.
Example:
(".", E) //tap
("-", T) //hold
("..", I) //double tap
(".-", A) //tap hold
("-...", B) //hold tap tap tap

### DIFF
--- a/rmk/src/keymap.rs
+++ b/rmk/src/keymap.rs
@@ -1,5 +1,6 @@
 #[cfg(feature = "storage")]
 use embedded_storage_async::nor_flash::NorFlash;
+use postcard::fixint::be;
 #[cfg(feature = "controller")]
 use {
     crate::channel::{CONTROLLER_CHANNEL, ControllerPub, send_controller_event},
@@ -195,13 +196,24 @@ impl<'a, const ROW: usize, const COL: usize, const NUM_LAYER: usize, const NUM_E
     ) -> Option<Morse<TAP_DANCE_MAX_TAP>> {
         match key_action {
             KeyAction::TapHold(tap_action, hold_action) => Some(
-                if let Action::Key(keycode) = tap_action
+                if self.behavior.morse.enable_hrm //TODO instead of this let the HRM keycodes configurable!
+                    && let Action::Key(keycode) = tap_action
                     && keycode.is_home_row()
                     && let Action::Modifier(modifier) = hold_action
                 {
-                    Morse::new_hrm(*tap_action, *modifier)
+                    Morse::new_hrm(
+                        *tap_action,
+                        *modifier,
+                        self.behavior.morse.operation_timeout.as_millis() as u16,
+                    )
                 } else {
-                    Morse::new_tap_hold(*tap_action, *hold_action)
+                    Morse::new_tap_hold_with_config(
+                        *tap_action,
+                        *hold_action,
+                        self.behavior.morse.operation_timeout.as_millis() as u16,
+                        self.behavior.morse.mode,
+                        self.behavior.morse.unilateral_tap,
+                    )
                 },
             ),
             KeyAction::TapDance(idx) => self.behavior.tap_dance.tap_dances.get(*idx as usize).map(|td| td.0),
--- a/rmk/src/via/keycode_convert.rs
+++ b/rmk/src/via/keycode_convert.rs
@@ -500,54 +500,54 @@ mod test {
         // LT0(A) -> LayerTapHold(A, 0)
         let via_keycode = 0x4004;
         assert_eq!(
-            KeyAction::Morse(Morse::new_layer_tap_hold(Action::Key(KeyCode::A), 0)),
+            KeyAction::TapHold(Action::Key(KeyCode::A), Action::LayerOn(0)),
             from_via_keycode(via_keycode)
         );
 
         // LT3(A) -> LayerTapHold(A, 3)
         let via_keycode = 0x4304;
         assert_eq!(
-            KeyAction::Morse(Morse::new_layer_tap_hold(Action::Key(KeyCode::A), 3)),
+            KeyAction::TapHold(Action::Key(KeyCode::A), Action::LayerOn(3)),
             from_via_keycode(via_keycode)
         );
 
         // LSA_T(A) ->
         let via_keycode = 0x2604;
         assert_eq!(
-            KeyAction::Morse(Morse::new_hrm(
+            KeyAction::TapHold(
                 Action::Key(KeyCode::A),
-                ModifierCombination::new_from(false, false, true, true, false)
-            )),
+                Action::Modifier(ModifierCombination::new_from(false, false, true, true, false))
+            ), //hrm
             from_via_keycode(via_keycode)
         );
 
         // RCAG_T(B) ->
         let via_keycode = 0x3D05;
         assert_eq!(
-            KeyAction::Morse(Morse::new_modifier_tap_hold(
+            KeyAction::TapHold(
                 Action::Key(KeyCode::B),
-                ModifierCombination::new_from(true, true, true, false, true)
-            )),
+                Action::Modifier(ModifierCombination::new_from(true, true, true, false, true))
+            ),
             from_via_keycode(via_keycode)
         );
 
         // ALL_T(A) ->
         let via_keycode: u16 = 0x2F04;
         assert_eq!(
-            KeyAction::Morse(Morse::new_hrm(
+            KeyAction::TapHold(
                 Action::Key(KeyCode::A),
-                ModifierCombination::new_from(false, true, true, true, true)
-            )),
+                Action::Modifier(ModifierCombination::new_from(false, true, true, true, true))
+            ), //hrm
             from_via_keycode(via_keycode)
         );
 
         // Meh_T(B) ->
         let via_keycode = 0x2705;
         assert_eq!(
-            KeyAction::Morse(Morse::new_modifier_tap_hold(
+            KeyAction::TapHold(
                 Action::Key(KeyCode::B),
-                ModifierCombination::new_from(false, false, true, true, true)
-            )),
+                Action::Modifier(ModifierCombination::new_from(false, false, true, true, true))
+            ),
             from_via_keycode(via_keycode)
         );
 
@@ -638,39 +638,39 @@ mod test {
         assert_eq!(0xF04, to_via_keycode(a));
 
         // LT0(A) -> LayerTapHold(A, 0)
-        let a = KeyAction::Morse(Morse::new_layer_tap_hold(Action::Key(KeyCode::A), 0));
+        let a = KeyAction::TapHold(Action::Key(KeyCode::A), Action::LayerOn(0));
         assert_eq!(0x4004, to_via_keycode(a));
 
         // LT3(A) -> LayerTapHold(A, 3)
-        let a = KeyAction::Morse(Morse::new_layer_tap_hold(Action::Key(KeyCode::A), 3));
+        let a = KeyAction::TapHold(Action::Key(KeyCode::A), Action::LayerOn(3));
         assert_eq!(0x4304, to_via_keycode(a));
 
         // LSA_T(A) ->
-        let a = KeyAction::Morse(Morse::new_modifier_tap_hold(
+        let a = KeyAction::TapHold(
             Action::Key(KeyCode::A),
-            ModifierCombination::new_from(false, false, true, true, false),
-        ));
+            Action::Modifier(ModifierCombination::new_from(false, false, true, true, false)),
+        );
         assert_eq!(0x2604, to_via_keycode(a));
 
         // RCAG_T(A) ->
-        let a = KeyAction::Morse(Morse::new_modifier_tap_hold(
+        let a = KeyAction::TapHold(
             Action::Key(KeyCode::A),
-            ModifierCombination::new_from(true, true, true, false, true),
-        ));
+            Action::Modifier(ModifierCombination::new_from(true, true, true, false, true)),
+        );
         assert_eq!(0x3D04, to_via_keycode(a));
 
         // ALL_T(A) ->
-        let a = KeyAction::Morse(Morse::new_modifier_tap_hold(
+        let a = KeyAction::TapHold(
             Action::Key(KeyCode::A),
-            ModifierCombination::new_from(false, true, true, true, true),
-        ));
+            Action::Modifier(ModifierCombination::new_from(false, true, true, true, true)),
+        );
         assert_eq!(0x2F04, to_via_keycode(a));
 
         // Meh_T(A) ->
-        let a = KeyAction::Morse(Morse::new_modifier_tap_hold(
+        let a = KeyAction::TapHold(
             Action::Key(KeyCode::A),
-            ModifierCombination::new_from(false, false, true, true, true),
-        ));
+            Action::Modifier(ModifierCombination::new_from(false, false, true, true, true)),
+        );
         assert_eq!(0x2704, to_via_keycode(a));
 
         // ComboOff
--- a/rmk/src/keyboard.rs
+++ b/rmk/src/keyboard.rs
@@ -2112,12 +2112,11 @@ mod test {
     use rusty_fork::rusty_fork_test;
 
     use super::*;
-    use crate::action::ExtractedKeyActionKeyActionKeyActionKeyActionKeyActionKeyActionKeyActionKeyActionKeyAction;
+    use crate::action::KeyAction;
     use crate::config::{BehaviorConfig, CombosConfig, ForksConfig};
     use crate::event::{KeyPos, KeyboardEvent, KeyboardEventPos};
     use crate::fork::Fork;
     use crate::hid_state::HidModifiers;
-    use crate::morse::Morse;
     use crate::{a, k, layer, mo, th};
 
     // Init logger for tests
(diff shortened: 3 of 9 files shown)


HaoboGu:
Looks like the memory consumption is great again. Let me check the functionalities

rmk/src/keyboard/held_buffer.rs
```diff
@@ -114,6 +116,8 @@ pub enum KeyState {
 pub struct HeldKey {
     pub event: KeyboardEvent,
     pub action: KeyAction,
+    /// if the action is tap/hold related, the config converted to morse
+    pub morse: Option<Morse<TAP_DANCE_MAX_TAP>>,
```
HaoboGu, line 120: Can the morse key be read on-the-fly? If so we don't need the extra `Option<Morse<>>` in `HeldKey`, `process_key_action` and `process_key_action_inner`.

rmk/src/via/keycode_convert.rs
```diff
@@ -98,18 +100,14 @@ pub(crate) fn from_via_keycode(via_keycode: u16) -> KeyAction {
...
-            }
+            KeyAction::TapHold(Action::Key(keycode), Action::Modifier(modifier))
         }
         0x4000..=0x4FFF => {
             // Layer tap-hold.
             // Layer tap-hold is in hold on other key press mode by default
             let layer = (via_keycode >> 8) & 0xF;
             let keycode = KeyCode::from_primitive(via_keycode & 0x00FF);
-            KeyAction::Morse(Morse::new_layer_tap_hold(Action::Key(keycode), layer as u8))
+            KeyAction::TapHold(Action::Key(keycode), Action::LayerOn(layer as u8))
```
HaoboGu, line 111: Now the layer tap-hold, modifier tap-hold and home-row modifier tap-hold have different configurations, and when setting them in vial, they can be automatically set, I think it's sufficient for most cases. Your change removes it, is there any possibilities that we can keep this?

rmk/src/keyboard/held_buffer.rs
tib888: I think, yes, `pub morse: Option<Morse<TAP_DANCE_MAX_TAP>>`  (and also Into_morse) can be deleted if all branches of match on `HeldKey.action` is  implemented instead.
I just started from your morse branch and this was the simplest to do.

rmk/src/via/keycode_convert.rs
tib888: I'm not sure if I understand this question right.
We may reintroduce all vial actions into KeyAction, or keep only KeyAction::TapHold and based on its content do the separation for vial protocol only.
(I may have missed that in the opposite direction...)

HaoboGu: Sorry I didn't make it clear. Let me explain it again: 

when converting via_keycode(u16) to `KeyAction`, the previous code will assign different morse settings according to the morse action type. For example, when the vial keycode is `mt!(a, LShift)`, then the `hrm!` is used, which enables flow tap and unilateral_tap for HRM. Flow tap and unilateral tap will not be set if the vial keycode is `mt!(b, LShift)`. 

That's why there're `Morse::new_layer_tap_hold`, `Morse::new_modifier_tap_hold` and `Morse::new_hrm`

tib888: well, my home row is: "k s t l v j n a e i", so the hardcoded hrm decision based on "a s d f g h j k l ;" is not the best.
I would put this into the positional config instead. (https://github.com/HaoboGu/rmk/issues/512)
so if a tap-hold like key is found in home locations, that should change its config.
That is why I pass the position info into the `into_morse` function.

tib888: but until that is not ready, the hard-coded hack may be used there also...

HaoboGu: Yes, since I think "asdfghjkl;" is the home-row for 99% people(this might not be correct for RMK users ;)), so having it as the default would be nice for most people. 

For advanced users, the default settings should be override by the explicit config in `keyboard.toml`.

HaoboGu:
I tested this PR on my hardware, it works as expected.

tib888:
Then lets consider accepting the PR in its current state!
I started working on the continuation (listed in the opening comment), but that is much more risky.
(This KeyState state machine is quite complicated... I'll need help for real morse pattern support.)

rmk/src/keymap.rs
```diff
@@ -185,6 +187,29 @@ impl<'a, const ROW: usize, const COL: usize, const NUM_LAYER: usize, const NUM_E
...
+    ) -> Option<Morse<TAP_DANCE_MAX_TAP>> {
+        match key_action {
+            KeyAction::TapHold(tap_action, hold_action) => Some(
+                if let Action::Key(keycode) = tap_action
+                    && keycode.is_home_row()
+                    && let Action::Modifier(modifier) = hold_action
+                {
+                    Morse::new_hrm(*tap_action, *modifier)
+                } else {
+                    Morse::new_tap_hold(*tap_action, *hold_action)
```
HaoboGu: Do you think if it's better to change the default tap-hold mode to `HoldOnOtherKeyPress`?

rmk/src/config/mod.rs
```diff
@@ -61,6 +61,9 @@ pub struct MorseConfig {
     pub mode: MorseKeyMode,
     /// If the previous key is on the same "hand", the current key will be determined as a tap
     pub unilateral_tap: bool,
+
+    // for advanced morse:
+    pub action_sets: Vec<Morse<TAP_DANCE_MAX_TAP>, 8>, //TODO MORSE_MAX_NUM = 8
```
HaoboGu, line 66: Now we can rename `TAP_DANCE_MAX_TAP` to `Morse_MAX_TAP`?

HaoboGu:
Left a few more comments. I think this PR can be merged after the resolving the conflicts

HaoboGu:
> (This KeyState state machine is quite complicated... I'll need help for real morse pattern support.)

Yes, I agree🥲. There are lots of integrated tests in `src/tests`, you can run them, check the logs for tracking the state changes.

rmk/src/via/keycode_convert.rs
tib888: The quick fix is added

rmk/src/keymap.rs
tib888: I agree.

rmk/src/config/mod.rs
tib888: This constant will disappear soon anyway, since the morse max tap will be limited in 15 only by the u16 representation.
However there will be a MORSE_MAX_NUM and a MAX_MORSE_PATTERNS_PER_KEY:

```
pub struct MorseConfig {
    pub enable_hrm: bool,
    pub prior_idle_time: Duration,
    /// Default timeout time for tap or hold
    pub operation_timeout: Duration,
    /// Default mode
    pub mode: MorseKeyMode,
    /// If the previous key is on the same "hand", the current key will be determined as a tap
    pub unilateral_tap: bool,

    // for advanced morse:
    pub action_sets: Vec<Morse<MAX_MORSE_PATTERNS_PER_KEY>, MORSE_MAX_NUM>,
}

pub struct Morse<const N: usize> {
    /// The actions triggered by tapping the key
    pub(crate) actions: MorseActions<N>,
    /// The timeout time for each operation in milliseconds
    pub timeout_ms: u16,
    /// The decision mode of the morse key
    pub mode: MorseKeyMode,
    /// If the unilateral tap is enabled
    pub unilateral_tap: bool,
}

pub struct MorseActions<const N: usize> {
    /// The actions list, use `Action::No` to represent an empty slot
    actions: [(MorsePattern, Action); N],
    /// The number of saved actions
    len: u8,
}

pub struct MorsePattern(u16);
```

HaoboGu: Looks great :D

tib888:
I just noticed that a lot of test cases relies on per key full Morse config - all of those have to be fixed somehow... :(

tib888:
Can you fix the test_repeat_key_th unit test?

tib888:
The advanced part is mostly ready on this branch:
https://github.com/tib888/rmk/tree/advanced-morse

The tests say it is fine, but on HW (as expected) not works well yet. 

The state machine is full of "//FIXME" comments, where I need help. 

The trivial pseudo code for the state machine woud be something like this below.
Could you please help to achieve the working state?

```
enum KeyState
{
   Waiting,
   Pressed(MorsePattern),
   Released(MorsePattern),
}

fn process(current_state: KeyState, currently_pressed: bool, last_change_time: &mut instant) -> KeyState {
    match (current_state, currently_pressed) {
        (KeyState::Waiting, false) => KeyState::Waiting, //not expected
        (KeyState::Waiting, true) => KeyState::Pressed(MorsePattern::default()), //start morse pattern
        (KeyState::Pressed(history), true) => KeyState::Pressed(history), //wait for release
        (KeyState::Pressed(history), false) =>
        //just released -> add tap or hold to the pattern history
        {
            if now - last_change_time > hold_threshold {
                last_change_time = now;
                KeyState::Released(history.followed_by_hold())
            } else {
                last_change_time = now;
                KeyState::Released(history.followed_by_tap())
            }
        }
        (KeyState::Released(history), false) =>
        //just time passed
        {
            if now - last_change_time > gap_threshold {
                morse.ExecutePattern(history); //the gap was longer than timeout -> execute the recorded pattern, reset state
                KeyState::Waiting
            } else {
                Released(history) //wait for further keypress
            }
        }
        (KeyState::Released(history), true) =>
        //just pressed again
        {
            if now - last_change_time > gap_threshold {
                last_change_time = now;
                morse.ExecutePattern(history); //the gap was too long -> try to execute the recorded pattern
                KeyState::Pressed(MorsePattern::default()) //start new morse pattern
            } else {
                last_change_time = now;
                KeyState::Pressed(history) // continue the current morse pattern
            }
        }
    }
}
```

HaoboGu:
> Can you fix the test_repeat_key_th unit test?

let me check

HaoboGu:
> I just noticed that a lot of test cases relies on per key full Morse config - all of those have to be fixed somehow... :(

yes, they are important because many of them come from real cases

HaoboGu:
> Can you fix the test_repeat_key_th unit test?

The unit test seems correct? Did I miss something?

HaoboGu:
A few comments:

> ```rust
> enum KeyState
> {
>    Waiting,
>    Pressed(MorsePattern),
>    Released(MorsePattern),
> }
> 
> fn process(current_state: KeyState, currently_pressed: bool, last_change_time: &mut instant) -> KeyState {
>     match (current_state, currently_pressed) {
>         (KeyState::Waiting, false) => KeyState::Waiting, //not expected
>         (KeyState::Waiting, true) => KeyState::Pressed(MorsePattern::default()), //start morse pattern <- last_change_time should also be recorded here
>         (KeyState::Pressed(history), true) => KeyState::Pressed(history), //wait for release  <- when hold timeouts, will this event be triggered? 
>         (KeyState::Pressed(history), false) =>
>         //just released -> add tap or hold to the pattern history
>         // <- It worth to check whether this pattern is the last pattern. If so, the pattern can be triggered immediately
>         {
>             if now - last_change_time > hold_threshold {
>                 last_change_time = now;
>                 KeyState::Released(history.followed_by_hold())
>             } else {
>                 last_change_time = now;
>                 KeyState::Released(history.followed_by_tap())
>             }
>         }
>         (KeyState::Released(history), false) =>
>         //just time passed
>         {
>             if now - last_change_time > gap_threshold {
>                 morse.ExecutePattern(history); //the gap was longer than timeout -> execute the recorded pattern, reset state
>                 KeyState::Waiting
>             } else {
>                 Released(history) //wait for further keypress
>             }
>         }
>         (KeyState::Released(history), true) =>
>         //just pressed again
>         {
>             if now - last_change_time > gap_threshold {
>                 last_change_time = now;
>                 morse.ExecutePattern(history); //the gap was too long -> try to execute the recorded pattern
>                 KeyState::Pressed(MorsePattern::default()) //start new morse pattern
>             } else {
>                 last_change_time = now;
>                 KeyState::Pressed(history) // continue the current morse pattern
>             }
>         }
>     }
> }
> ```

This pseudo code looks correct for **a single key**, but in the practice, the most complex part is the multiple-key cases, with or without key rolling and permissive-hold/hold-on-other-key-press mode. That's very complex, which took me about 2 weeks to finish all the messy code(with bugs:/. 

It becomes even more complex when introducing the hold-tap pattern. So my suggestion is to write the tests first, use test + log to get how the state machine work and what's expected behavior, and then fix the code. Yeah it is very very difficult, so just take it easy.

tib888:
> > Can you fix the test_repeat_key_th unit test?
> 
> The unit test seems correct? Did I miss something?

that is the only failing test from the ones I was able to execute.

could you list the commands I should use for testing? (I do not want to miss any...)

I was running: 
`cargo test --no-default-features --features="log, std" --verbose`

tib888:
> This pseudo code looks correct for **a single key**, but in the practice, the most complex part is the multiple-key cases, with or without key rolling and permissive-hold/hold-on-other-key-press mode. That's very complex, which took me about 2 weeks to finish all the messy code(with bugs:/.
> 
> It becomes even more complex when introducing the hold-tap pattern. So my suggestion is to write the tests first, use test + log to get how the state machine work and what's expected behavior, and then fix the code. Yeah it is very very difficult, so just take it easy.

this is exacly why I asked your help: I did not want to reinvent the wheel, only do the minimally needed changes to the already working code. the problem is that it is hard to understand the current code, and I do not know, where to put the analog of these pseudo code fragments:

KeyState::Pressed(MorsePattern::default())
KeyState::Released(history.followed_by_hold())
KeyState::Released(history.followed_by_tap())

could you at least describe in detail, what is the exact meaning/intent of these:
Morse.unilateral_tap
MorseKeyMode::PermissiveHold

HaoboGu:
> could you list the commands I should use for testing? (I do not want to miss any...)

I'm using similar command: `cd rmk && cargo test --no-default-features`

> Morse.unilateral_tap

When this option is enabled, and tapping a morse key whose preceded key is on the same hand, trigger tapping immediately. It's like [Chordal Hold](https://docs.qmk.fm/tap_hold#chordal-hold) in QMK and [positional hold-tap](https://zmk.dev/docs/keymaps/behaviors/hold-tap#positional-hold-tap-and-hold-trigger-key-positions) in zmk.

> MorseKeyMode::PermissiveHold

When permissive hold is enabled, a morse key is held and another key is pressed & released, trigger morse key's hold immediately. It's [PermissiveHold](https://docs.qmk.fm/tap_hold#permissive-hold) in Qmk and [balanced behavior](https://zmk.dev/docs/keymaps/behaviors/hold-tap#interrupt-flavors) in zmk.

HaoboGu:
I reviewed the FIXME part in https://github.com/tib888/rmk/tree/advanced-morse carefully, most of them are correct, only a few places needs to be fixed:

- https://github.com/tib888/rmk/blob/890877b3a68ed2e65dca26bafce9347483650d09/rmk/src/keyboard.rs#L483-L485: needs followed_by_tap
- https://github.com/tib888/rmk/blob/890877b3a68ed2e65dca26bafce9347483650d09/rmk/src/keyboard.rs#L490: needs followed_by_tap

One thing I want to mention is that the `MorsePattern` is different from the previous `u8` in `KeyState::Held/IdlePostTap/PostTap/...`, because the `MorsePattern` records the action that's already happened, for example `Held("0b10")` means a tap is already issued, while u8 in `Held(u8)` means the index of current undecided action. If a tap is already issued, then the state should be `Held(1)`, indicates that the current key is held again after the first tap and the current available action should be "double_tap"(tap_actions[1]) or "hold_after_tap"(hold_actions[1]). 

I guess this subtle difference may lead to the wrong behavior?

tib888:
than you for the feedback, I'm not near to computers currently. I will check the details when I get back...

HaoboGu:
Oh I found why there's extra mem used, see this issue: https://github.com/rust-lang/rust/issues/62958

I added [this commit](https://github.com/HaoboGu/rmk/pull/521/commits/101c1f7c351d86246691ff408ae80975e2d3ae9b), the extra memory usage is gone.

image

tib888:
> Oh I found why there's extra mem used, see this issue: [rust-lang/rust#62958](https://github.com/rust-lang/rust/issues/62958)
> 
> I added [this commit](https://github.com/HaoboGu/rmk/pull/521/commits/101c1f7c351d86246691ff408ae80975e2d3ae9b), the extra memory usage is gone.

What should happen then with this branch?
Keep it to save even more memory or close it, and I should rebase somehow my `advanced-morse` branch (which I am working on currently) onto the main and finish it there?

HaoboGu:
> What should happen then with this branch?
> Keep it to save even more memory or close it, and I should rebase somehow my advanced-morse branch (which I am working on currently) onto the main and finish it there?

I prefer a new PR for the advanced-morse branch and merging anything you think should be added from this PR to the new PR

tib888:
https://github.com/HaoboGu/rmk/pull/527 will replace this.

PR #516, opened 2025-08-07
feat/vial-lock-and-matrix-tester

drindr:
- vial-lock (for matrix tester only now)
- allow specify unlock keys in the keyboard.toml
- matrix tester feature

### DIFF
--- a/rmk/src/matrix.rs
+++ b/rmk/src/matrix.rs
@@ -22,10 +22,8 @@ pub struct MatrixState<const ROW: usize, const COL: usize> {
 
 #[cfg(feature = "matrix_tester")]
 impl<const ROW: usize, const COL: usize> MatrixState<ROW, COL> {
-    const ROW: u8 = ROW as u8;
-    const COL: u8 = COL as u8;
-    const ROW_LEN: u8 = (COL as u8 + 8) / 8;
-    const OUT_OF_BOUNDARY: () = if Self::ROW * Self::ROW_LEN > 30 {
+    const ROW_LEN: usize = (COL + 8) / 8;
+    const OUT_OF_BOUNDARY: () = if ROW * Self::ROW_LEN > 30 {
         panic!(
             "Cannot use matrix tester because your keyboard has too many keys. \
             Consider disable the `matrix_tester` feature"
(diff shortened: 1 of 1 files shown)


rmk/src/matrix.rs
```diff
@@ -12,6 +12,55 @@ use crate::event::{Event, KeyboardEvent};
...
+        let slice = &self.state[..(row as usize * row_len as usize)];
+        let mut target_iter = target.iter_mut();
+        for row_bytes in slice.chunks(row_len.into()) {
+            for byte in row_bytes.iter().rev() {
+                *target_iter.next().unwrap() = *byte;
+            }
+        }
+    }
+    pub fn read(&self, row: u8, col: u8) -> bool {
+        let index = row * self.row_len * 8 + col;
```
HaoboGu: The index should to be checked, ensure that the input row/col are legal and the `self.state[]` won't panic

rmk/src/matrix.rs
```diff
@@ -12,6 +12,55 @@ use crate::event::{Event, KeyboardEvent};
...
+    state: [u8; 30],
+    row_len: u8,
+}
+
+#[cfg(feature = "matrix_tester")]
+impl<const ROW: usize, const COL: usize> MatrixState<ROW, COL> {
+    pub fn new() -> Self {
+        Self {
+            state: [0; 30],
+            row_len: (<usize as TryInto<u8>>::try_into(COL).unwrap() + 8) / 8,
```
HaoboGu: The row_len is calculated according from a generic const, maybe [associated const](https://doc.rust-lang.org/reference/items/associated-items.html#associated-constants) is better?

HaoboGu (changes requested):
The code looks good to me! Some documentation is needed for configuring unlock keys

rmk/src/matrix.rs
```diff
@@ -12,6 +12,70 @@ use crate::event::{Event, KeyboardEvent};
...
+pub struct MatrixState<const ROW: usize, const COL: usize> {
+    // 30 bytes is the limited by Vial and 240 keys is enough for
+    // most keyborad
+    state: [u8; 30],
+}
+
+#[cfg(feature = "matrix_tester")]
+impl<const ROW: usize, const COL: usize> MatrixState<ROW, COL> {
+    const ROW: u8 = ROW as u8;
+    const COL: u8 = COL as u8;
```
HaoboGu: is this redundant? I think ROW_LEN and OUT_OF_BOUNDARY are enough

HaoboGu:
Looks good to me now. I left one extra minor comment, I'm good with merging it as is

HaoboGu:
thanks a lot!

PR #520, opened 2025-08-10
Separate vial feature

### LINKED ISSUE #272: Option to disable Vial configuration storage
A configuration option should exist that lets users configure their keymaps only via the keyboard.toml configuration. Ideally Vial would still be available for configuration testing, but the configuration would get reset on reboot. This would be different than `clear_storage = true` in that only the keymaps would get reset, not other irrelevant storage such as bluetooth configuration etc.

something like `disable_vial = true` or alternatively `use_vial = false`, or maybe even `save_vial_layout = false`

### LINKED ISSUE #470: Non-modifier tap-hold support
Currently only `KeyAction::TapHold(Action::Key(...), {Action::Modifier(...), Action::LayerOn(...)})` are deserialized to/from via keycodes
```rust
//keycode_convert.rs:. 66
        KeyAction::TapHold(tap, hold) => match hold {
            Action::LayerOn(l) => {
                if l > 16 {
                    0
                } else {
                    let keycode = match tap {
                        Action::Key(k) => k as u16,
                        _ => 0,
                    };
                    0x4000 | ((l as u16) << 8) | keycode
                }
            }
            Action::Modifier(m) => {
                let keycode = match tap {
                    Action::Key(k) => k as u16,
                    _ => 0,
                };
                0x2000 | ((m.into_bits() as u16) << 8) | keycode
            }
            _ => 0x0000,
        },

//keycode_convert.rs: 106
        0x2000..=0x3FFF => {
            // Modifier tap/hold
            // The via equivalent of Modifier tap/hold is called Mod-tap, whose keycode representation is same with RMK
            let keycode = KeyCode::from_primitive(via_keycode & 0x00FF);
            let modifier = ModifierCombination::from_bits(((via_keycode >> 8) & 0b11111) as u8);
            KeyAction::TapHold(Action::Key(keycode), Action::Modifier(modifier))
        }
        0x4000..=0x4FFF => {
            // Layer tap/hold
            // The via equivalent of Modifier tap/hold is called Mod-tap,
            let layer = (via_keycode >> 8) & 0xF;
            let keycode = KeyCode::from_primitive(via_keycode & 0x00FF);
            KeyAction::TapHold(Action::Key(keycode), Action::LayerOn(layer as u8))
        }

```

This means something like `th!(A, B)` does not work, since this will end up being translated to `KeyAction::No` .

Having this possibility would be useful e.g. for autoshift on hold (`th!(A, shifted!(B))`. In my zmk layout, I use something like this for emitting e.g. `(`  on tap, and a `()` macro on hold.

alex-tdrn:
I've tested these changes by running the `cargo test` + `cargo check` with and without the `vial` feature.

I haven't tested directly that a firmware built with vial in mind still works as expected, but I've tested that some of the examples build and flash. Would need some help here from other people to test these changes, if they use vial.

Made it so we test multiple combinations of features on CI, since the storage tests have not been working for a while. 
Also add `cargo check` with different feature sets. This way, any PR will be alerted if they have leftover code that's not needed for one of the feature combinations.

I've tested that I can use non-modifier tap-hold with the "vial" feature removed (although there seem to be some key repeating bugs, which I'm still investigating. I don't think they are related to this PR though). So this closes #470 .

Also added a `.gitattributes` and `.editorconfig` file to force everything to have LF endings. I'm on windows, and neovim creates `CRLF` by default, if there's no `.editorconfig` file telling it otherwise.

NOTE: this is a breaking change for users who have `no-default-features` in their `Cargo.toml`. They will have to add back the `vial` feature, unless they use the `matrix_tester` or `vial_lock` features, since they automatically bring in the `vial` feature.

EDIT: would also close #272


alex-tdrn:
Ah, didn't take the toml config into account. I'll fix it tomorrow.

rmk/src/via/vial.rs
```diff
@@ -1,3 +1,5 @@
+#![cfg(feature = "vial")]
```
HaoboGu: It's unnecessary because this file is already behind `vial` feature gate in `via/mod.rs`

rmk/src/via/protocol.rs
```diff
@@ -1,3 +1,5 @@
+#![cfg(feature = "vial")]
```
HaoboGu: It's unnecessary because this file is already behind vial feature gate in `via/mod.rs`

HaoboGu:
Thanks for the work!

Is it possible to move vial related code to separated sub-mods? For example, adding a `vial.rs` in `src/ble/trouble` and moving vial code to `src/ble/trouble/vial.rs`. This way could reduce the usage of `#[cfg(not(feature = "vial"))]`.

And, I have to admit the current implementation is not very clear, all those features aren't decoupled very well, making it hard to extract/disable a single functionality. Some refactoring is needed 🥲

alex-tdrn:
> Thanks for the work!
> 
> Is it possible to move vial related code to separated sub-mods? For example, adding a `vial.rs` in `src/ble/trouble` and moving vial code to `src/ble/trouble/vial.rs`. This way could reduce the usage of `#[cfg(not(feature = "vial"))]`.
> 
> And, I have to admit the current implementation is not very clear, all those features aren't decoupled very well, making it hard to extract/disable a single functionality. Some refactoring is needed 🥲

True, I'll try to isolate vial a bit better this week, and I'll see what I can come up with ✌️. Marking this as draft again until I fix CI and do some refactoring to reduce the `cfg` soup a bit.

HaoboGu:
successor: #554

PR #522, opened 2025-08-10
Macros config support

patricios-space:
# Macro config support

This PR enables users to map keys to macro triggers directly through keyboard.toml keymap

## Tasks

- [x] Add m! rust macro for creating a macro trigger action
- [x] Add M(n) to keymap parser to trigger macros
- [ ] Update documentation

## Example

```
[[behavior.macro.macros]]
operations = [
  { operation = "down", keycode = "RAlt" }, 
  { operation = "tap", keycode = "Quote" },
  { operation = "up", keycode = "RAlt" },
]


[[layer]]
name = "base_layer"
keys = """
M(0)
"""
```

### DIFF
--- a/rmk-config/src/keymap.pest
+++ b/rmk-config/src/keymap.pest
@@ -98,12 +98,15 @@ shifted_action = { ^"SHIFTED" ~ "(" ~ keycode_name ~ ")" }
 // Rule 8: TD(n) - Tap Dance
 tap_dance_action = { ^"TD" ~ "(" ~ number ~ ")" }
 
+// Rule 9: M(n) - Trigger Macro
+trigger_macro_action = { ^"M" ~ "(" ~ number ~ ")" }
+
 // --- Top Level Rules ---
 
 // A single key action entry in the map
 // Order is important: more specific function-like rules first, then aliases/specials, then simple keycodes.
 key_action = _{ // Consume surrounding whitespace/comments implicitly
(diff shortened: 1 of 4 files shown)


rmk-config/src/keymap.pest
```diff
@@ -98,12 +98,15 @@ shifted_action = { ^"SHIFTED" ~ "(" ~ keycode_name ~ ")" }
 // Rule 8: TD(n) - Tap Dance
 tap_dance_action = { ^"TD" ~ "(" ~ number ~ ")" }
 
+// Rule 9: M(n) - Trigger Macro
```
HaoboGu, line 101: I think users might be confused by `M(n)` and `MO(n)`, they looks similar but they're totally different. Use `Macro(n)` maybe?

duck-types: I see your point and agree that `M(n)` can be a bit confusing. I thought the same of all layer triggers though. I imagine that the project kept this abbreviations to be as consistent as possible with QMK/Vial users. As in Vial macros triggers are written as `Mn`, I thought using `M(n)` was the most similar option. If you don't care that much about familiarity, `Macro(n)` sounds good to me. Let me know and I can rename it or complete the documentation for triggering macros.

HaoboGu: Yes, rename it please

HaoboGu:
@patricios-space Would you have time to do the renaming? Or I can just merge it as is.

HaoboGu:
Okay I'm merging it and do the rename later

PR #524, opened 2025-08-12
Advanced morse

tib888:
WIP - ability to handle real morse codes

The tap_hold, tap_dance kept unchanged in the config to be recognizable by forks, non-positional combos, etc., but unified in the held buffer into morse.

Fixes are still needed because 5 tests fails...


tib888:
https://github.com/HaoboGu/rmk/pull/527 will replace this.

PR #527, opened 2025-08-14
real morse implementation

tib888:
Ability to handle real morse codes

The tap_hold, tap_dance kept unchanged in the config to be recognizable by forks, non-positional combos, etc.
Morse configs can be indexed the same way as tap dances, but they do not have vial support.

Remaining work:
Config from keyboard.toml, add some unit tests

Example:
(".", E) //tap
("-", T) //hold
("..", I) //double tap
(".-", A) //tap hold
("-...", B) //hold tap tap tap


tib888:
https://github.com/HaoboGu/rmk/pull/530 replaced this

PR #530, opened 2025-08-15
real morse implementation

tib888:
Ability to handle real morse codes

The tap_hold, tap_dance kept unchanged in the config to be recognizable by forks, non-positional combos, etc.
Morse configs can be indexed the same way as tap dances, but they do not have vial support.

Examples:

```
[[behavior.morse]]
timeout = "200ms"
permissive_hold = false
unilateral_tap = false
hold_on_other_press = false
actions = [
   {pattern = ".-", action = "A"}, 
   {pattern = "-...", action = "B"}, 
   {pattern = "-.-.", action = "C"}, 
   {pattern = "-..", action = "D"}, 
   {pattern = ".", action = "E"}, 
   {pattern = "..-.", action = "F"}, 
   {pattern = "--.", action = "G"}, 
   {pattern = "....", action = "H"}, 
   {pattern = "..", action = "I"}, 
   {pattern = ".---", action = "J"}, 
   {pattern = "-.-", action = "K"}, 
   {pattern = ".-..", action = "L"}, 
   {pattern = "--", action = "M"}, 
   {pattern = "-.", action = "N"}, 
   {pattern = "---", action = "O"}, 
   {pattern = ".--.", action = "P"}, 
   {pattern = "--.-", action = "Q"}, 
   {pattern = ".-.", action = "R"}, 
   {pattern = "...", action = "S"}, 
   {pattern = "-", action = "T"}, 
   {pattern = "..-", action = "U"}, 
   {pattern = "...-", action = "V"}, 
   {pattern = ".--", action = "W"}, 
   {pattern = "-..-", action = "X"}, 
   {pattern = "-.--", action = "Y"}, 
   {pattern = "--..", action = "Z"}, 
   {pattern = ".----", action = "Kc1"}, 
   {pattern = "..---", action = "Kc2"}, 
   {pattern = "...--", action = "Kc3"}, 
   {pattern = "....-", action = "Kc4"}, 
   {pattern = ".....", action = "Kc5"}, 
   {pattern = "-....", action = "Kc6"}, 
   {pattern = "--...", action = "Kc7"}, 
   {pattern = "---..", action = "Kc8"}, 
   {pattern = "----.", action = "Kc9"}, 
   {pattern = "-----", action = "Kc0"}
]
```

```
[[behavior.morse]]
timeout = "300ms"
permissive_hold = false
unilateral_tap = false
hold_on_other_press = false
actions = [ 
  # Extended many function keys with morse
  {pattern = ".", action = "Kp1"}, {pattern = "..", action = "Kp2"}, {pattern = "...", action = "Kp3"}, {pattern = "....", action = "Kp4"}, {pattern = ".....", action = "Kp5"}
]
```

### DIFF
--- a/docs/en/docs/features/configuration/behavior.md
+++ b/docs/en/docs/features/configuration/behavior.md
@@ -154,8 +154,7 @@ Tap dance configuration includes the following parameters:
   - `tap_actions`: An array of actions, each corresponding to the number of taps. For example, `tap_actions = ["F1", "F2", "F3"]` means a single tap triggers "F1", double tap triggers "F2", triple tap triggers "F3", and so on. If the tap count exceeds the length of the array, the last action is used.
   - `hold_actions`: An array of actions, each corresponding to holding the key after a certain number of taps. For example, `hold_actions = ["MO(1)", "MO(2)", "MO(3)"]` means holding after one tap triggers "MO(1)", holding after two taps triggers "MO(2)", and so on. If the tap count exceeds the length of the array, the last action is used.
   - `morse_actions`: list of patten -> action pairs. The pattern is a tap/hold sequence, its length is limited in 15. The morse pattern of `C` for example can be described like this: `"-.-."` or `"_._."` or `"1010"`. See the examples below!
-  - `strict`: The pattern checking should be strict, no predictions allowed before the user finishes the pattern. Defaults to `false` except when morse_actions used.
-
+  - `strict`: Means that the pattern checking should be strict, no predictions are allowed before the user finishes the pattern. Defaults to `true`.
 
 ::: warning
 
--- a/rmk-macro/src/behavior.rs
+++ b/rmk-macro/src/behavior.rs
@@ -202,6 +202,8 @@ fn expand_tap_dance(tap_dance: &Option<TapDancesConfig>) -> proc_macro2::TokenSt
                     }
                     None => quote! { 200u16 },
                 };
+
+                let strict = td.strict.unwrap_or(true);//the default is strict = true
                 
                 if let Some(morse_actions) = &td.morse_actions {
                     if td.tap.is_some() || td.hold.is_some() || td.hold_after_tap.is_some() || td.double_tap.is_some() || td.tap_actions.is_some() || td.hold_actions.is_some() {
@@ -209,8 +211,6 @@ fn expand_tap_dance(tap_dance: &Option<TapDancesConfig>) -> proc_macro2::TokenSt
                     }
                     
                     let actions_def = expand_morse_actions(&morse_actions);
-
-                    let strict = td.strict.unwrap_or(true);//for morse the default is strict = true
 
                     quote! {                        
                         ::rmk::tap_dance::TapDance {
@@ -249,8 +249,6 @@ fn expand_tap_dance(tap_dance: &Option<TapDancesConfig>) -> proc_macro2::TokenSt
                         None => quote! { ::rmk::heapless::Vec::new() },
                     };
 
-                    let strict = td.strict.unwrap_or(false);
-
                     quote! {
                         ::rmk::tap_dance::TapDance::new_with_actions(
                             #tap_actions_def,
@@ -264,7 +262,6 @@ fn expand_tap_dance(tap_dance: &Option<TapDancesConfig>) -> proc_macro2::TokenSt
                     let hold = parse_key(td.hold.clone().unwrap_or_else(|| "No".to_string()));
                     let hold_after_tap = parse_key(td.hold_after_tap.clone().unwrap_or_else(|| "No".to_string()));
                     let double_tap = parse_key(td.double_tap.clone().unwrap_or_else(|| "No".to_string()));
-                    let strict = td.strict.unwrap_or(false);
 
                     quote! {
                         ::rmk::tap_dance::TapDance::new_from_vial(
--- a/rmk/src/tap_dance.rs
+++ b/rmk/src/tap_dance.rs
@@ -97,7 +97,7 @@ impl Default for TapDance {
             timeout_ms: 250,
             mode: TapHoldMode::HoldOnOtherPress,
             unilateral_tap: false,
-            strict_pattern_checking: false,
+            strict_pattern_checking: true,
             actions: Vec::default(),
         }
     }


tib888:
Theoretically the PR is ready, everything works, including morse unit tests.
But when config used to configure morse keys, it does not work on my HW. (Meanwhile th, td, work well.)
So there must be something configuration or storage related little thing, which is missing or wrong.

Please review / merge!

HaoboGu:
Nice work!! I'll take a look asap

tib888:
Finally it works: I have hacked vial support by mapping the morse keys after the indexes of tap dance keys.

tib888:
Please note, that some unit tests are randomly failing on slow/heavily loaded testing HW, likely because in some checks the Instant::now() is used instead of the real event time stamp. This should be improved!

rmk/src/storage/mod.rs
```diff
@@ -534,28 +567,46 @@ impl Value<'_> for StorageData {
                     if buffer.len() < 3 + TAP_DANCE_MAX_TAP * 4 {
                         return Err(SerializationError::InvalidData);
                     }
-                    let timeout = BigEndian::read_u16(&buffer[1..3]);
+                    let timeout_ms = BigEndian::read_u16(&buffer[1..3]);
```
HaoboGu: Can the storage of `TapDanceData` be merged into the `MorseKeyData` storage?

HaoboGu:
> Finally it works: I have hacked vial support by mapping the morse keys after the indexes of tap dance keys.

How did you achieve it, could you provide an example for this setting?

rmk/src/storage/mod.rs
tib888: I want to keep TapDance and Morse separate because of one important diference: (see: is_pattern_completed() )
The tap dance is able to make its decision immediately when a hold occures -> so you can have hold actions, like shift, alt, ctrl. While morse may do this only if you reach the longest pattern.

Due to the same reason, hrm decisions are important and valid for th, td, but not much for morse.

However for morse a separate gap and hold timeout config may be needed - the tests will show.

tib888:
> > Finally it works: I have hacked vial support by mapping the morse keys after the indexes of tap dance keys.
> 
> How did you achieve it, could you provide an example for this setting?

Just put this into your config:

``` toml
[[behavior.morse]]
timeout = "200ms"
permissive_hold = false
unilateral_tap = false
hold_on_other_press = false
actions = [
   {pattern = ".-", action = "A"}, 
   {pattern = "-...", action = "B"}, 
   {pattern = "-.-.", action = "C"}, 
   {pattern = "-..", action = "D"}, 
   {pattern = ".", action = "E"}, 
   {pattern = "..-.", action = "F"}, 
   {pattern = "--.", action = "G"}, 
   {pattern = "....", action = "H"}, 
   {pattern = "..", action = "I"}, 
   {pattern = ".---", action = "J"}, 
   {pattern = "-.-", action = "K"}, 
   {pattern = ".-..", action = "L"}, 
   {pattern = "--", action = "M"}, 
   {pattern = "-.", action = "N"}, 
   {pattern = "---", action = "O"}, 
   {pattern = ".--.", action = "P"}, 
   {pattern = "--.-", action = "Q"}, 
   {pattern = ".-.", action = "R"}, 
   {pattern = "...", action = "S"}, 
   {pattern = "-", action = "T"}, 
   {pattern = "..-", action = "U"}, 
   {pattern = "...-", action = "V"}, 
   {pattern = ".--", action = "W"}, 
   {pattern = "-..-", action = "X"}, 
   {pattern = "-.--", action = "Y"}, 
   {pattern = "--..", action = "Z"}, 
   {pattern = ".----", action = "Kc1"}, 
   {pattern = "..---", action = "Kc2"}, 
   {pattern = "...--", action = "Kc3"}, 
   {pattern = "....-", action = "Kc4"}, 
   {pattern = ".....", action = "Kc5"}, 
   {pattern = "-....", action = "Kc6"}, 
   {pattern = "--...", action = "Kc7"}, 
   {pattern = "---..", action = "Kc8"}, 
   {pattern = "----.", action = "Kc9"}, 
   {pattern = "-----", action = "Kc0"}
]
```

Then in your keymap use " MORSE(0) ", just like you would do with " TD(0) ".
In rust code, use the mrs!(0) macro instead.

If you add a second one it will become be MORSE(1):
``` toml
[[behavior.morse]]
timeout = "300ms"
permissive_hold = false
unilateral_tap = false
hold_on_other_press = false
actions = [ 
  {pattern = ".", action = "Kp1"}, {pattern = "..", action = "Kp2"}, {pattern = "...", action = "Kp3"}, {pattern = "....", action = "Kp4"}, {pattern = ".....", action = "Kp5"}
]
```

HaoboGu:
It seems that in this way you still cannot set it in vial? what did you mean by "hacked the vial support"

tib888:
> It seems that in this way you still cannot set it in vial? what did you mean by "hacked the vial support"

Morse will appear in vial as tap dance, starting from the index tap_dance_max_num. And you can see/set there the '.', '..', '-', '.-'  actions only.

So if tap_dance_max_num is 10, then morse(2) will appear as tap dance 12 in vial.

HaoboGu:
If I understand it correctly, the full Morse operation sequence cannot be displayed/changed, only the first few operations like tap-dance in vial can be changed?

tib888:
> If I understand it correctly, the full Morse operation sequence cannot be displayed/changed, only the first few operations like tap-dance in vial can be changed?

Correct. But at least you can assign your pre-configured morse to keys as they would be tap dances. Without this even if I put the MORSE(0) in the keymap in the config, it did not work. (Making Vial optional would be nice!)

tib888:
I think, this is ready to merge, so it only depends on you now....

HaoboGu:
I haven't got enough time for reviewing. It may take a few days

rmk/src/keyboard/morse.rs
```diff
@@ -160,4 +154,115 @@ impl<'a, const ROW: usize, const COL: usize, const NUM_LAYER: usize, const NUM_E
...
+                HOLD => *hold_action,
+                _ => Action::No,
+            },
+            KeyAction::TapDance(idx) => {
+                behavior_config
+                    .tap_dance
+                    .tap_dances
+                    .get(*idx as usize)
+                    .map_or(Action::No, |td| {
+                        let l = pattern.pattern_length();
```
HaoboGu: Can `td.0.get(pattern)` be used here? Because the underlying representation of tap-dance is morse, too.

EDIT: oh sorry I forgot that the tap-dance's struct has been changed. So please address the comment at `is_pattern_completed`

rmk/src/keyboard/morse.rs
```diff
@@ -160,4 +154,115 @@ impl<'a, const ROW: usize, const COL: usize, const NUM_LAYER: usize, const NUM_E
...
+                .morse_keys
+                .get(*idx as usize)
+                .map(|morse| Duration::from_millis(morse.timeout_ms as u64)),
+
+            _ => None,
+        }
+        .unwrap_or_else(|| behavior_config.morse.operation_timeout)
+    }
+
+    pub fn morse_mode(behavior_config: &BehaviorConfig, keyAction: &KeyAction) -> (MorseKeyMode, bool) {
```
HaoboGu: A comment is needed here, for explaining what's the returned `bool`

rmk/src/keyboard/morse.rs
```diff
@@ -160,4 +154,115 @@ impl<'a, const ROW: usize, const COL: usize, const NUM_LAYER: usize, const NUM_E
...
+        }
+    }
+
+    pub fn is_pattern_completed(
+        behavior_config: &BehaviorConfig,
+        keyAction: &KeyAction,
+        pattern: MorsePattern,
+    ) -> bool {
+        pattern.is_full()
+        || (pattern.last_is_hold() && matches!(keyAction, KeyAction::TapDance(_) | KeyAction::TapHold(_, _))) //allow tap hold, tap dance to finish with the first long press (this is not true for real morse!)
```
HaoboGu: I think there's better way to calculate it by checking whether there is longer patterns that **starts with** current finished pattern.

Take td as the example, typically td has for action seqs: `.`, `_`, `..`, `._`.

- When the first action is `_`, the only pattern that starts with `_` is `_` itself, so when `_` comes as the first action, the completed decision can be made immediately
- When the first action is `.`, the there are two longer actions are `.*`, aka `..` and `._`, so it is determined as not finished

With this I believe that the difference between normal morse keys and tap-dance/tap-hold can be eliminated, the storage code could also be simplified.

rmk/tests/common/morse.rs
```diff
@@ -1,35 +1,76 @@
...
+    let morse0 = MorseKey {
+        actions: Vec::from_slice(&[
+            (MorsePattern::from_u16(0b1_01), Action::Key(KeyCode::A)),
+            (MorsePattern::from_u16(0b1_1000), Action::Key(KeyCode::B)),
+            (MorsePattern::from_u16(0b1_1010), Action::Key(KeyCode::C)),
+            //(MorsePattern::from_u16(0b1_100), Action::Key(KeyCode::D)),
+            //(MorsePattern::from_u16(0b1_0), Action::Key(KeyCode::E)),
+            //(MorsePattern::from_u16(0b1_0010), Action::Key(KeyCode::F)),
+            // (MorsePattern::from_u16(0b1_110), Action::Key(KeyCode::G)),
+            // (MorsePattern::from_u16(0b1_0000), Action::Key(KeyCode::H)),
```
HaoboGu: Remove the unused code?

rmk/tests/keyboard_morse_tap_dance_test.rs
```diff
@@ -261,6 +261,27 @@ rusty_fork_test! {
...
+    fn test_rolling_3() {
+        key_sequence_test! {
+            keyboard: create_tap_dance_test_keyboard(),
+            sequence: [
+                [0, 0, true, 150], // Press td!(0)
+                [0, 0, false, 10], // Release td!(0)
+                [0, 0, true, 150], // Press td!(0)
+                [0, 1, true, 260], // Press td!(1) -> td!(0) timeout (tap-hold)
+                [0, 1, false, 260], // Release td!(1)
+                [0, 0, false, 260], // Release td!(0) -> td!(1) timeout (hold)
```
HaoboGu: Is this comment incorrect? td!(1) is already released at this time

rmk/tests/keyboard_morse_test.rs
```diff
@@ -781,4 +782,106 @@ rusty_fork_test! {
...
+                [0, [kc_to_u8!(O), 0, 0, 0, 0, 0]],
+                [0, [0, 0, 0, 0, 0, 0]],
+                [0, [kc_to_u8!(S), 0, 0, 0, 0, 0]],
+                [0, [0, 0, 0, 0, 0, 0]],
+            ]
+        };
+    }
+
+    #[test]
+    //#[ignore]
```
HaoboGu: #[ignore] can be removed

rmk/src/keyboard/morse.rs
tib888: The idea is good. But what should happen if there is no match at all for the pattern due to some user mistake? 

For example the possibilities are:
```
.      - A
-      - B
.-..   - C
--..   - D

```
and the user input is: `---` should this terminate immediately and accept the last "-" as a "B" or since it is a worong pattern, so wait until there is a long gap and do nothing since the user made a mistake? (I would vote for the later.)

An other example: the possibilities are:
```
.     - A
.-..  - B
--..  - C
```

and the user input is: ` ..`  should this finish with no action or with "A", "A" ? (I would vote for no action.)

And one last example:
```
.         - Kp1
..        - Kp2
...       - Kp3
```

and the user input is accidentally four tap with no long gap: ` .... ` Would you expect no result or "3", "1" ? (I would vote for no action - but that would contradict the idea!)

rmk/src/keyboard/morse.rs
tib888: well, actually I changed that, it is not based on morse anymore:

``` rust
pub struct TapDance {
    /// array of (tap, hold) action pairs
    pub(crate) actions: [(Action, Action); TAP_DANCE_MAX_TAP],
    /// The timeout time for each operation in milliseconds
    pub timeout_ms: u16,
    /// The decision mode of the morse key
    pub mode: MorseKeyMode,
    /// If the unilateral tap is enabled
    pub unilateral_tap: bool,
}
```

I kept this because I did not want to break the compatibility of existing toml configurations.

But if we go with your idea of unifying TD+Morse, then morse config could go into the tap dances, like this:
what do you think?

``` toml
[behavior.tap_dance]
tap_dances = [

# vial like:
{ tap = "U", hold = "RightBracket", double_tap = "Minus", hold_after_tap = "BackSlash"},#u ú ü ű

# advanced tap dance:
{ 
    tap_actions = ["F1", "F2", "F3", "F4", "F5"], 
    hold_actions = ["MO(1)", "MO(2)", "MO(3)", "MO(4)", "MO(5)"],
    tapping_term = "300ms" 
},

# real morse:
{
    tapping_term = "250ms"
    gap_time = "250ms"
    morse_actions = [
       {pattern = ".-", action = "A"}, 
       {pattern = "-...", action = "B"}, 
       {pattern = "-.-.", action = "C"}, 
       {pattern = "-..", action = "D"}, 
       {pattern = ".", action = "E"}, 
       {pattern = "..-.", action = "F"}
    ]
}
]
```

rmk/src/keyboard/morse.rs
tib888: done

rmk/tests/keyboard_morse_tap_dance_test.rs
tib888: fixed

rmk/tests/keyboard_morse_test.rs
tib888: done

rmk/src/keyboard/morse.rs
HaoboGu: 🤔 I would prefer "DB", "AA" and "31" for the three cases, because the keyboard never knows whether the user is intentional or not. In my opinion, if users are making mistakes, the wrong output is better than silent fail. 

Maybe an option can be added later: when it is enabled, all wrong actions in the middle is ignored until the user stops input(timeout); when it's disabled, the early decision is made when user inputs the wrong action.

tib888: the refactor is almost ready - just some runtime borrowing error occured on some tests...

HaoboGu: Check the `.await` usage when the `RefCell<KeyMap>` is borrowed.

rmk/src/keyboard/morse.rs
HaoboGu: How about moving tap-dance config to morse? aka 
```toml
[behavior.morse]
morses = [
    
  # vial like:
  { tap = "U", hold = "RightBracket", double_tap = "Minus", hold_after_tap = "BackSlash"},#u ú ü ű
  
  # advanced tap dance:
  { 
      tap_actions = ["F1", "F2", "F3", "F4", "F5"], 
      hold_actions = ["MO(1)", "MO(2)", "MO(3)", "MO(4)", "MO(5)"],
      tapping_term = "300ms" 
  },
  
  # real morse:
  {
      tapping_term = "250ms"
      gap_time = "250ms"
      morse_actions = [
         {pattern = ".-", action = "A"}, 
         {pattern = "-...", action = "B"}, 
         {pattern = "-.-.", action = "C"}, 
         {pattern = "-..", action = "D"}, 
         {pattern = ".", action = "E"}, 
         {pattern = "..-.", action = "F"}
      ]
  }
]
```

Adding a documentation in tap-dance section to explaining this migration is enough then.

And in vial, the morse keys are loaded as tap-dance, but only `.`, `..`, `_`, `._` are configurable.

tib888: I kept the old names TapHold, TapDance, TapHoldConfig since they are already known by everyone, and also used in the configs, macros with the same name. However TapDance now stores (pattern, action) pairs, to support morse.

'morse' name is used only as the super set of TapHold, TapDance in the rmk code around the pattern processing state machine.

rmk/src/keyboard/morse.rs
tib888: I have implemented `try_predict_final_action` instead of `is_pattern_completed`.
Now only the real morse unit test fail.
I am working to solve that, but likely a "strict" mode should be introduced anyway, and morse_actions should set it to true by default.

One example why it is needed:
```
.         - Esc
..........  - Reboot
```

try_predict_final_action would predict and execute Reboot immediately at the second tap.

HaoboGu: Yeah, this case should be definitely considered, I agree with you, the reboot should not be triggered in the `..` case.

How about adding a condition to check whether current action sequence triggers a `KeyAction` or not?

Reconsidering the first tap-dance case: `.`, `_`, `..`, `._`, the `_` could be triggered immediately when the action seq is exactly `_`.

If we only do `predict_final_action` when current action sequence matches exactly an output `KeyAction`, the issue can be solved imo.

For example, if we have this:

```
..   - Kp1
...  - Kp2
.._  - Kp3
..._ - Kp4
__   - Kp5
```

`_` should not trigger Kp5, but `__` triggers.


Back to your 3 example, the output should become "no", "no" and "3,1".

tib888: I introduced a "strict" option, which was true only for morse_actions, but now I set it to true in all cases.

tib888:
I consider this PR ready, please try it out, merge!

These are the related final configs: 
(the tap_dance_max_num, max_patterns_per_key are automatically calculated if you do not set them) 
(in the keymapp use TD(0), TD(1), TD(2), ... )

```
[rmk]
# Maximum number of tap dances
tap_dance_max_num = 8
# Maximum number of morse patterns a morse key can handle (min 4 due to tap dance, max 255)
max_patterns_per_key = 36

[behavior.tap_dance]
tap_dances = [
  { tap = "A", hold = "B", double_tap = "C", hold_after_tap = "D"},

  { tap_actions = ["F1", "F2", "F3", "F4", "F5"], hold_actions = ["Kp1", "Kp2", "Kp3", "Kp4", "Kp5"] },

  { timeout = "200ms", strict = true, morse_actions = [
      {pattern = ".-", action = "A"}, 
      {pattern = "-...", action = "B"}, 
      {pattern = "-.-.", action = "C"}, 
      {pattern = "-..", action = "D"}, 
      {pattern = ".", action = "E"}, 
      {pattern = "..-.", action = "F"}, 
      {pattern = "--.", action = "G"}, 
      {pattern = "....", action = "H"}, 
      {pattern = "..", action = "I"}, 
      {pattern = ".---", action = "J"}, 
      {pattern = "-.-", action = "K"}, 
      {pattern = ".-..", action = "L"}, 
      {pattern = "--", action = "M"}, 
      {pattern = "-.", action = "N"}, 
      {pattern = "---", action = "O"}, 
      {pattern = ".--.", action = "P"}, 
      {pattern = "--.-", action = "Q"}, 
      {pattern = ".-.", action = "R"}, 
      {pattern = "...", action = "S"}, 
      {pattern = "-", action = "T"}, 
      {pattern = "..-", action = "U"}, 
      {pattern = "...-", action = "V"}, 
      {pattern = ".--", action = "W"}, 
      {pattern = "-..-", action = "X"}, 
      {pattern = "-.--", action = "Y"}, 
      {pattern = "--..", action = "Z"}, 
      {pattern = ".----", action = "Kc1"}, 
      {pattern = "..---", action = "Kc2"}, 
      {pattern = "...--", action = "Kc3"}, 
      {pattern = "....-", action = "Kc4"}, 
      {pattern = ".....", action = "Kc5"}, 
      {pattern = "-....", action = "Kc6"}, 
      {pattern = "--...", action = "Kc7"}, 
      {pattern = "---..", action = "Kc8"}, 
      {pattern = "----.", action = "Kc9"}, 
      {pattern = "-----", action = "Kc0"}
    ] }
]
```

PR #532, opened 2025-08-19
Remove strict mode for tap-dances, use exactly match instead

HaoboGu:
with some code refactors

### DIFF
--- a/rmk/src/tap_dance.rs
+++ b/rmk/src/tap_dance.rs
@@ -81,9 +81,6 @@ pub struct TapDance {
     /// If the unilateral tap is enabled
     pub unilateral_tap: bool,
 
-    // real morse codes need stricter pattern checking
-    pub strict_pattern_checking: bool,
-
     /// The list of pattern -> action pairs, which can be triggered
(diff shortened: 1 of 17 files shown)


rmk/src/tap_dance.rs
```diff
@@ -166,29 +150,30 @@ impl TapDance {
...
 
-        if let Some((pattern, action)) = first {
-            if !self.strict_pattern_checking || pattern_start.pattern_length() == pattern.pattern_length() {
-                // in strict case we want to have full length match (prediction too early may confuse the user)
-                Some(*action)
-            } else {
-                None
-            }
+        if let Some(action) = first {
+            Some(*action)
```
tib888, line 176: This should have left here to be always stritc!
```
          // we want to have full length match (prediction too early may confuse the user)
          if pattern_start.pattern_length() == pattern.pattern_length() { 
                Some(*action)
            } else {
                None
            }
```

HaoboGu: the current way is always strict, the only difference is that for

. kp1
... kp2

when you tap ...., the output is "21"

PR #533, opened 2025-08-19
Rename `TapDance` to `Morse`

HaoboGu:
Morse key was introduced in #530, but now the configuration & processing uses tap dance, which is actually different from what the original tap dance means. 

This PR rename `TapDance` to `Morse`, adds documentation that explains the difference between `Morse` and `TapDance`.

`TapHold` is kept as is now, but in the future `TapHold` can also be merged into `Morse` action. That means there's only one kind of multi-role key in RMK world, definitions like `TapHold` and `TapDance` are only a subset of `Morse`. To help the new users, the documentation needs to be polished.

### DIFF
--- a/rmk-config/src/keymap.pest
+++ b/rmk-config/src/keymap.pest
@@ -95,15 +95,15 @@ tap_hold_action = _{ mt_action | th_action }
 // Rule 7: SHIFTED(key)
 shifted_action = { ^"SHIFTED" ~ "(" ~ keycode_name ~ ")" }
 
-// Rule 8: TD(n) - Tap Dance index
-tap_dance_action = { ^"TD" ~ "(" ~ number ~ ")" }
+// Rule 8: TD(n) - Morse(Tap Dance) index
+morse_action = { ^"TD" ~ "(" ~ number ~ ")" }
 
 // --- Top Level Rules ---
 
 // A single key action entry in the map
 // Order is important: more specific function-like rules first, then aliases/specials, then simple keycodes.
 key_action = _{ // Consume surrounding whitespace/comments implicitly
-    wm_action | osm_action | layer_action | tap_hold_action | shifted_action | tap_dance_action | no_action | transparent_action | simple_keycode
+    wm_action | osm_action | layer_action | tap_hold_action | shifted_action | morse_action | no_action | transparent_action | simple_keycode
 }
 
 // The entire key map string: Start, zero or more key actions, End.
--- a/rmk-config/src/keymap.pest
+++ b/rmk-config/src/keymap.pest
@@ -22,7 +22,7 @@ number = @{ ASCII_DIGIT+ }
 layer_number = @{ number }
 layer_name = @{ loose_identifier }
 
-//the order is important here, as we want to match the number first 
+// The order is important here, as we want to match the number first 
 layer_reference = _{ layer_number | layer_name }
 
 // Modifier Names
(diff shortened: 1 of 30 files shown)


rmk-config/src/keymap.pest
```diff
@@ -95,15 +95,15 @@ tap_hold_action = _{ mt_action | th_action }
 // Rule 7: SHIFTED(key)
 shifted_action = { ^"SHIFTED" ~ "(" ~ keycode_name ~ ")" }
 
-// Rule 8: TD(n) - Tap Dance index
-tap_dance_action = { ^"TD" ~ "(" ~ number ~ ")" }
+// Rule 8: TD(n) - Morse(Tap Dance) index
```
tib888: Lets write instead: 
// Rule 8: TD(n) - Morse index (in Vial its simplest form is known as "Tap Dance", hence the TD name)

rmk/src/layout_macro.rs
```diff
@@ -163,10 +163,18 @@ macro_rules! encoder {
     };
 }
 
-/// Create a tap dance action
+/// Create a morse(tap dance) action
```
tib888: Lets write instead: 
/// Create a Morse(index) action (in Vial its simplest form is known as "Tap Dance", hence the td name)

rmk/src/layout_macro.rs
```diff
@@ -163,10 +163,18 @@ macro_rules! encoder {
...
 #[macro_export]
 macro_rules! td {
     ($index: literal) => {
-        $crate::action::KeyAction::TapDance($index)
+        $crate::action::KeyAction::Morse($index)
+    };
+}
+
+/// Create a morse(tap dance) action
```
tib888: Lets write instead:
/// Create a Morse(index) action (in Vial it will appear as "Tap Dance")

rmk/src/via/vial.rs
```diff
@@ -295,80 +295,73 @@ pub(crate) async fn process_vial<
...
+                    let morses = &mut keymap.borrow_mut().behavior.morse.morses;
 
-                    if tap_dance_idx < tap_dances.len() {
-                        // Update the tap dance in keymap
-                        if let Some(tap_dance) = tap_dances.get_mut(tap_dance_idx) {
-                            // Extract tap dance data from report
+                    if morse_idx < morses.len() {
+                        // Update the morse in keymap
+                        if let Some(morse) = morses.get_mut(morse_idx) {
+                            // Extract morse(tap dance) data from report
```
tib888: // Extract morse (also known as "tap dance" in vial)

tib888:
> Morse key was introduced in #530, but now the configuration & processing uses tap dance, which is actually different from what the original tap dance means.
> 
> This PR rename `TapDance` to `Morse`, adds documentation that explains the difference between `Morse` and `TapDance`.
> 
> `TapHold` is kept as is now, but in the future `TapHold` can also be merged into `Morse` action. That means there's only one kind of multi-role key in RMK world, definitions like `TapHold` and `TapDance` are only a subset of `Morse`. To help the new users, the documentation needs to be polished.

I think mergeing TapHold into Morse is not practical due to the difference in their memory usage: TH const amout 2*sizeof(Action), while Morse has an array settings + an array which costs sizeof((MorsePattern, Action)) * max pattern length -the solution to refer these by index saves a lots of memory in keymaps.

On the other hand TapHold currently support such combination of 2 actions, which is not vial compatible.
Would it be possible to create such a config macro, which convert those to Morse automatically?
(Of course this would makes sense only if the vial compatibility is important...)

rmk-config/src/keymap.pest
HaoboGu: I'm thinking accepting both "TD" and "Morse", which are actually same in the keymap

HaoboGu:
> I think mergeing TapHold into Morse is not practical due to the difference in their memory usage: TH const amout 2*sizeof(Action), while Morse has an array settings + an array which costs sizeof((MorsePattern, Action)) * max pattern length -the solution to refer these by index saves a lots of memory in keymaps.
> 
> 

yes, I agree with you, that why I keep it for now. TapHold will be merged only if:

1. we have a great alternative for Vial, which allows us do whatever we want, then the Taphold can be made global indexed, just like TapDance 
2. Or per-key configuration of Taphold is needed and Vial supports it. In this case the Morse struct should have MaxPatternLength = 2 for Taphold, which will also increase the memory usage but it's acceptable.

PR #537, opened 2025-08-26
per key position configuration

tib888:
For issue https://github.com/HaoboGu/rmk/issues/512

`matrix_map` extended with optional L/R hand information and home row information - `is_same_hand`, `is_home_row` functions can be improved with this.

In the future other info, like per key timeout or other similar info could be added....

Example:
```
matrix_map = """
(2, 0, L) (2, 1, L)      (1, 2, L)      (1, 3, L)      (1, 4, L)      (1, 5, L)      (1, 6, R) (1, 7, R)      (1, 8, R)      (1, 9, R)      (2, 10, R)      (2, 11, R)
(3, 0, L) (3, 1, L, HRM) (2, 2, L, HRM) (2, 3, L, HRM) (2, 4, L, HRM) (2, 5, L)      (2, 6, R) (2, 7, R, HRM) (2, 8, R, HRM) (2, 9, R, HRM) (3, 10, R, HRM) (3, 11, R)
(4, 0, L) (4, 1, L)      (3, 2, L)      (3, 3, L)      (3, 4, L)      (3, 5, L)      (3, 6, R) (3, 7, R)      (3, 8, R)      (3, 9, R)      (4, 10, R)      (4, 11, R)
                                                    (4, 2, L) (4, 3, L) (4, 4, L)  (4, 7, R) (4, 8, R) (4, 9, R)
"""
```


HaoboGu:
Because in the future more configuration of a key will be added, I'm considering that maybe we need a separate key config, like 

```toml
[[key_config]]
name = "hrm"
positions = "(0, 1), (0, 2)" # Positions in `matrix_map`
tapping_term = "250ms"
unilateral_tap = true
morse_mode = "permissive_hold"
# Other configurations ...
```
?

Keep `(x, y, hand)` is fine in matrix map I think

tib888:
What about this idea:

In the matrix map refer named key profiles:

```toml
# split ortho example for matrix map, with hand and home row information filled:
matrix_map = """
(0, 0, L)     (0, 1, L)     (0, 2, L)     (0, 3, L)     (0, 4, L)         (0, 5, R)   (0, 6, R)      (0, 7, R)     (0, 8, R)     (0, 9, R)   
(1, 0, L, H2) (1, 1, L, H2) (1, 2, L, H)  (1, 3, L, H1) (1, 4, L)         (1, 5, R)   (1, 6, R, H1)  (1, 7, R, H)  (1, 8, R, H2) (1, 9, R, H2)
(2, 0, L)     (2, 1, L)     (2, 2, L)     (2, 3, L)     (2, 4, L)         (2, 5, R)   (2, 6, R)      (2, 7, R)     (2, 8, R)     (2, 9, R)   
                                           (3, 3, L)    (3, 4, L)         (3, 5, R)   (3, 6, R) 
"""

[key_profiles]
"H" = { unilateral_tap = true, morse_mode = "permissive_hold" }
"H1" = { based_on = "H", timeout_adjust = "-10ms"  }
"H2" = { based_on = "H", timeout_adjust = "+10ms"  }
```

The key profiles could "inherit" and modify each other (if `based_on` is missing, they start from the default morse config).
The keys, which do not refer a profile would be also based on default morse config.

`enable_hrm`, per morse timing, mode config could be removed, since this could replace those.

Anyway the goal of this PR is just to provide the infrastructure to store per key configuration in BehaviorConfig.key_info.
The KeyInfo struct later can be extended to hold more/less info, starting with Hand.

HaoboGu:
Oh I like the `(1, 1, L, H2)` idea, maybe the hand can be also merged, for example `(1, 1, LH)` for lefthand, and `(1, 10, RH)` for right hand. 

In `[key_profile]` I prefer to specify each profile without inheritance. The inheritance might be hard to implement maybe? Because you have to deal with a typology of the inherit graph. And it doesn't make the configuration much more clearer than specifying all key profiles directly IMO.

> Anyway the goal of this PR is just to provide the infrastructure to store per key configuration in BehaviorConfig.key_info.
The KeyInfo struct later can be extended to hold more/less info, starting with Hand.

Yeah, maybe an issue for more discussion is great.

tib888:
This version is ready:

```toml
# split ortho example for matrix map, with hand and home row information filled:
matrix_map = """
(0, 0, L)    (0, 1, L)    (0, 2, L)    (0, 3, L)    (0, 4, L)         (0, 5, R)   (0, 6, R)    (0, 7, R)    (0, 8, R)    (0, 9, R)   
(1, 0, L:H2) (1, 1, L:H2) (1, 2, L:H1) (1, 3, L:H1) (1, 4, L)         (1, 5, R)   (1, 6, R:H1) (1, 7, R:H1) (1, 8, R:H2) (1, 9, R:H2)
(2, 0, L)    (2, 1, L)    (2, 2, L)    (2, 3, L)    (2, 4, L)         (2, 5, R)   (2, 6, R)    (2, 7, R)    (2, 8, R)    (2, 9, R)   
                                       (3, 3, L:T)  (3, 4, L:T)       (3, 5, R:T) (3, 6, R:T) 
"""

[behavior]
tap_hold = { enable_flow_tap = true, prior_idle_time = "120ms", hold_on_other_press = true, hold_timeout = "250ms", gap_timeout = "250ms" }

[behavior.key_profiles]
# matrix_map may refer these to override the defaults given in tap_hold for some key positions - this example is a home row mod
H1 = { permissive_hold = true, unilateral_tap = true, hold_timeout = "200ms", gap_timeout = "200ms" }
H2 = { permissive_hold = true, unilateral_tap = true, hold_timeout = "250ms", gap_timeout = "250ms" }
# thumb tap-hold example
T = { hold_on_other_press = true, unilateral_tap = false, hold_timeout = "250ms", gap_timeout = "250ms" }
```

tib888:
Please review/merge!

HaoboGu:
Thanks for the work, I'll have time to review it next week

rmk/src/config/mod.rs
```diff
@@ -24,18 +24,28 @@ pub struct RmkConfig<'a> {
...
     pub one_shot: OneShotConfig,
     pub combo: CombosConfig,
     pub fork: ForksConfig,
     pub morse: MorsesConfig,
     pub keyboard_macros: KeyboardMacrosConfig,
     pub mouse_key: MouseKeyConfig,
+
+    // if a key_info.profile may override the defaults defined in tap_hold
+    pub tap_hold: TapHoldConfig,
+    pub key_info: Option<[[KeyInfo; COL_NUM]; ROW_NUM]>,
```
HaoboGu, line 51: Adding `KeyInfo` matrix to `BehaviorConfig` feels like incorrect: now the number of morse keys are limited, but `key_info` costs `size_of(KeyInfo) * COL_NUM * ROW_NUM`, which cannot really reduce the overhead compared with

```rust
pub struct Morse<const MAX_PATTERN_NUM: usize> {
    /// Configuration for tap-hold and morse
    pub key_profile: KeyProfile,
    /// The list of pattern -> action pairs, which can be triggered
    pub actions: Vec<(MorsePattern, Action), MAX_PATTERN_NUM>,
}

// KeyAction
pub enum KeyAction {
    /// No action. Serialized as 0x0000.
    No,
    /// Transparent action, next layer will be checked. Serialized as 0x0001.
    Transparent,
    /// A single action, such as triggering a key, or activating a layer. Action is triggered when pressed and cancelled when released.
    Single(Action),
    /// Don't wait the release of the key, auto-release after a time threshold.
    Tap(Action),
    /// Tap hold action, use morse struct with the configuration
    TapHold(Morse<2>),
    /// Morse action, references a morse configuration by index.
    Morse(u8),
}
```

and introduced lots of checking of whether a configuration for a key exists.

From my understanding, the `key_info: KeyInfo` is like a unified configuration struct of keys, that means you can change the config of a `Single` key to `Some(config_for_morse)` by accident in the code. I prefer to move the configuration like `KeyProfile` into the specific types, like

```rust
pub struct Morse<const MAX_PATTERN_NUM: usize> {
    /// Configuration for tap-hold and morse
    pub key_profile: KeyProfile,
    /// The list of pattern -> action pairs, which can be triggered
    pub actions: Vec<(MorsePattern, Action), MAX_PATTERN_NUM>,
}
```

tib888: > Adding KeyInfo matrix to BehaviorConfig feels like incorrect: now the number of morse keys are limited, but key_info costs size_of(KeyInfo) * COL_NUM * ROW_NUM, which cannot really reduce the overhead compared with

Yes, I was not sure either. This way at least the *LAYER_NUM does not apply on the size.
The other solution would be to have an array for key profiles, and use indices only in KeyInfo -> in that case KEY_PROFILE_MAX_NUM has to be introduced.

To simplify the overview, I have collected the most important part here:

``` rust
pub struct KeyProfile {
    /// If the previous key is on the same "hand", the current key will be determined as a tap
    pub unilateral_tap: bool,
    /// The decision mode of the morse/tap-hold key
    pub mode: MorseMode,

    /// If the key is pressed longer than this, it is accepted as `hold` (in milliseconds)
    pub hold_timeout_ms: u16,
    /// The time elapsed from the last release of a key is longer than this, it will break the morse pattern (in milliseconds)
    pub gap_timeout_ms: u16,
}

//this contains the defaults
pub struct TapHoldConfig {
    pub enable_flow_tap: bool,     //default: false
    pub prior_idle_time: Duration, //used for flow tap detection, default: 120ms,
    pub default_profile: KeyProfile,
}

//this is stored for every key position
pub struct KeyInfo {
    pub hand: Hand,
    pub profile: Option<KeyProfile>,
}

pub struct BehaviorConfig<const ROW_NUM: usize, const COL_NUM: usize> {
    //...
    pub key_info: Option<[[KeyInfo; COL_NUM]; ROW_NUM]>,
}
```

If we want to save memory, I can refactor the above into to:
``` rust
//this is stored for every key position
pub struct KeyInfo {
    pub hand: Hand,
    pub profile: Option<u8>, //index into BehaviroConfig.key_profiles
}

pub struct BehaviorConfig<const ROW_NUM: usize, const COL_NUM: usize> {
    //...
    pub key_info: Option<[[KeyInfo; COL_NUM]; ROW_NUM]>;
    pub key_profiles: [KeyProfile; KEY_PROFILE_MAX_NUM];
}
```

What do you think? (Also maybe KeyProfile could be renamed to MorseProfile?)

> From my understanding, the key_info: KeyInfo is like a unified configuration struct of keys, that means you can change the config of a Single key to Some(config_for_morse) by accident in the code. I prefer to move the configuration like KeyProfile into the specific types, like

Currently KeyInfo is a place holder for any position specific information. 
Currently contains only L/R, morse key profile config, but later can be extended with (just ideas):
- phisycal coordinates, key size/shape for the future editing UI,
- per key back light sequence number
- etc.

HaoboGu: It doesn't really address my concern: in current impl when a key is not a `Morse` key, it's still possible to set the `profile` to `Some(x)`. In other word, you have to align `profile: Option<u8>` to the key set at the position **manually**: 

```
profile = None - no morse key at the position
profile = Some(x) - there exists morse key at the position
```

There's no guarantee. That's why I prefer 

```rust
pub struct Morse<const MAX_PATTERN_NUM: usize> {
    /// Configuration for tap-hold and morse
    pub key_profile: KeyProfile,
    /// The list of pattern -> action pairs, which can be triggered
    pub actions: Vec<(MorsePattern, Action), MAX_PATTERN_NUM>,
}
```

when it costs even more memory(by forcing the configuration for each-key, not each position).


As for the global configuration like left/right hand, backlight sequence number, etc, I'm fine with the current `KeyInfo`

tib888: My point is that if you have different morse/taphold keys (like I do) at the same position, but on different layer, you want them to work with the same timing, etc. profile. That is why I do not want to store this in the Morse. 
Furthermore, the TapHold keys do not have field so store config currently, but with this solution if they are used in a certain key position, they will get the config.

HaoboGu: Yeah, I got your point. For me I prefer the real per-key configuration, but I'm also fine with position based config. What's unacceptable for me is the unified configuration struct, that might be a rabbit hole for future developers. My opinion is that the configuration for a `Morse` key should be set ONLY when the key is in the keymap, it should be **guaranteed**.

HaoboGu: I propose two solutions:

1. Like what I proposed in the previous comment, add the configuration to `Morse` and change `TapHold` to `Morse<2>`. 
```rust
pub enum KeyAction {
    /// ..
    /// Tap hold action, use morse struct with the configuration
    TapHold(Morse<2>),
    /// Morse action, references a morse configuration by index.
    Morse(u8),
}
```
  - pros: fully customizable keys, "unlimited" tap-hold keys 
  - cons: consumes more memory
2. Index all TapHold and Morse in a list, and add a configuration list, too
```rust
pub enum KeyAction {
    /// ..
    /// Tap hold action, references a tap hold configuration by index
    TapHold(u8),
    /// Morse action, references a morse configuration by index.
    Morse(u8),
}
```
  - pros: memory efficient, fully customizable keys, consistent implementation of TapHold/Morse
  - cons: the maximum number of TapHold/Morse is limited at the compile-time

tib888: This later one practically breaks the vial compatibility:  in all layers and all key positions you may have a different tap-hold combination. So the tap hold array size should prepare for this worst case so the  size_of(TapHold/Morse) * COL_NUM * ROW_NUM * LAYER_NUM memory requirement is the worst possible. 
Also you will have no clue when using vial, what settings are associated with the newly configured TH / TD keys.

While, if we go with positional profile config, it is perfectly clear what you need to fine tune...

tib888: > Yeah, I got your point. For me I prefer the real per-key configuration, but I'm also fine with position based config. What's unacceptable for me is the unified configuration struct, that might be a rabbit hole for future developers. My opinion is that the configuration for a Morse key should be set ONLY when the key is in the keymap, it should be guaranteed.

I really do not understand what is the problem. 
```
matrix_map = """
(2, 0, L) (2, 1, L)      (1, 2, L)      (1, 3, L)      (1, 4, L)      (1, 5, L)      (1, 6, R) (1, 7, R)      (1, 8, R)      (1, 9, R)      (2, 10, R)      (2, 11, R)
(3, 0, L) (3, 1, L, HRM) (2, 2, L, HRM) (2, 3, L, HRM) (2, 4, L, HRM) (2, 5, L)      (2, 6, R) (2, 7, R, HRM) (2, 8, R, HRM) (2, 9, R, HRM) (3, 10, R, HRM) (3, 11, R)
(4, 0, L) (4, 1, L)      (3, 2, L)      (3, 3, L)      (3, 4, L)      (3, 5, L)      (3, 6, R) (3, 7, R)      (3, 8, R)      (3, 9, R)      (4, 10, R)      (4, 11, R)
                                                    (4, 2, L) (4, 3, L) (4, 4, L)  (4, 7, R) (4, 8, R) (4, 9, R)
"""
```

In some key positions you have the option to override the defaults, but this does not mean, that we unify everything.
It means only that if a tap-hold/morse key is procesed, it looks for its position and if there is an config override it appiles that. Other kind of keys of course do not need that info, so will not use it...

HaoboGu: > It means only that if a tap-hold/morse key is procesed, it looks for its position and if there is an config override it appiles that. Other kind of keys of course do not need that info, so will not use it...

Yeah that's true, we know it, but other RMK developers/contributors might not know. That alignment

- Morse key: profile: Some(x)
- Non-Morse key: profile: None
 
is guaranteed by **developers**, not **code**, which *allows* people to make mistakes.

If you put profile in `Morse`, there's just no way to set the profile when the key is not `Morse`

tib888: But the tap hold config can not be solved (currently with Vial GUI) this way -> for that only the positional config works (currently). 
I'm not completely against your proposal, but I need a solution now because my home row is not "asdf".

I think the users will understand this concept easily if the documentation/naming is good enough (would it be better to name "key_profile" to "tap_hold_profile" ?).

(Also, the memory usage of KeyInfo (including KeyProfile) struct can be decreased to just 32 bits per key with bitfield structs.)

HaoboGu: > But the tap hold config can not be solved (currently with Vial GUI) this way -> for that only the positional config works (currently). 
> I'm not completely against your proposal, but I need a solution now because my home row is not "asdf".

I don't see any difference between positional config and per-key config? IMO both of them cannot be set with Vial GUI, and both work for non-"asdf" home row. I am also fine with the position based config(while I prefer real per-key config), so I guess we get agreement here?

For me the problem is the implementation, position based config has no 100% guarantee for correct config after changing keys many times at a position in different layers. Considering that RMK might have its own GUI in the future, and the available memory on MCUs will keep increasing, I think the per-key configuration is the way. 

>I think the users will understand this concept easily if the documentation/naming is good enough (would it be better to name "key_profile" to "tap_hold_profile" ?).
> 
> (Also, the memory usage of KeyInfo (including KeyProfile) struct can be decreased to just 32 bits per key with bitfield structs.)

I agree, that would be better

tib888: OK, then one last question: if you would have a keymap like this for example:
```rust
[[layer]]
name = "eng"
keys ="""
G  Y  M  R  B                                     W  C  O  U @; 
LT(sym, K) S  MT(T, LCtrl) MT(L, LShift)  V       J MT(N, RShift)  MT(A, LCtrl)  E LT(sym, I)
F  D  P  H  Q                                     X  Z @/ @, @. 
"""
```
How would you configure "per-key" the tap-holds used in the middle row? (With the indexed morse it would be easy easy, I know, but this is tap-hold.)
Where would that config end up in the toml config? 
How would those configs reference the keymap (or vice versa)? 
What will happen if you move one of them to a different position (with vial)?
What will happen if you switch the Shift with Ctrl or change "sym" to "num" layer (with vial), so "new" tap-holds appear at the same position?

HaoboGu: If we do this, the `tap_hold_profile` or `morse_profile` only appears in tap-hold keys, then the `LT`/`MT`s may take an optional configuration param:

```toml
[[layer]]
name = "eng"
keys ="""
G  Y  M  R  B                                               W  C  O  U @; 
LT(sym, K) S  MT(T, LCtrl, hrm) MT(L, LShift, hrm)  V       J MT(N, RShift, hrm)  MT(A, LCtrl, hrm)  E LT(sym, I)
F  D  P  H  Q                                               X  Z @/ @, @. 
"""

[[morse_profile]]
name = "hrm"
mode = "permissive_hold"
unilateral_tap = true
```

> What will happen if you move one of them to a different position (with vial)?
> What will happen if you switch the Shift with Ctrl or change "sym" to "num" layer (with vial), so "new" tap-holds appear at the same position?

Yeah, that kind of configuration cannot be set in Vial. That's the limitation of Vial. But if we do our own GUI, this should not be the limitation, the configuration could be more intuitive.

tib888: I have done the renaming, reduced the memory usage of TapHoldProfile to 32 bits.

My idea is to put the "per-key" config you proposed behind a feature gate, so if one needs it this and has a capable MCU, then he may enable the feature. 
(In that case the keymap will use an extra NUM_ROW * NUM_COL * NUM_LAYER * 32bit of ram)

What do you think?

HaoboGu: You mean support both per-position & per-key configuration? I'm fine with that, but I'm afraid that the feature gate might become very complex.. Maybe this implementation can be used?

```
pub enum KeyAction {
    /// ..
    /// Tap hold action, references a tap hold configuration by index
    TapHold(u8),
    /// Morse action, references a morse configuration by index.
    Morse(u8),
}


```

The list size is `ROW*COL*NUM_LAYER` when per-key configuration is enabled and `ROW*COL` when per-position configuration is used.

tib888: The simplest solution would be to put back `profile: TapHoldProfile,` into the Morse: 
```rust
pub struct Morse<const MAX_PATTERN_NUM: usize> { 
     pub profile: TapHoldProfile,
     pub actions: Vec<(MorsePattern, Action), MAX_PATTERN_NUM>, 
}
```
Then if the user needs per key tuning, he should simply port those tap-hold keys to morse and live with that.

----------
> You mean support both per-position & per-key configuration? I'm fine with that, but I'm afraid that the feature gate might become very complex..

I'll give a try to this feature gate idea:

```rust
pub enum KeyAction {
    ...
    TapHold(Action, Action, TapHoldProfile), //use TapHoldProfile here only if feature gate enabled
    Morse(u8),
}
pub struct Morse<const MAX_PATTERN_NUM: usize> { 
     pub profile: TapHoldProfile, //this could also depend on feature gate, but not important
     pub actions: Vec<(MorsePattern, Action), MAX_PATTERN_NUM>, 
}
```

If I do not succeed, then I'll finish with the above solution.

----------

>pub enum KeyAction {
>    /// ..
>    /// Tap hold action, references a tap hold configuration by index
>    TapHold(u8),
>    /// Morse action, references a morse configuration by index.
>    Morse(u8),
>}

Using TapHold(u8) would mean that if vial is used, depending on the current keymap content the tap-hold array has to be repopulated, and if the user inserts more than max_taphold_num, then fails. Also, the management of this would be quite complex... so I will not do this for sure.
(BTW, in this case TapHolds should be just deleted, since TapHold(u8) and Morse(u8) are (almost) the same, and would be redundant.)

HaoboGu: > Using TapHold(u8) would mean that if vial is used, depending on the current keymap content the tap-hold array has to be repopulated, and if the user inserts more than max_taphold_num, then fails. Also, the management of this would be quite complex... so I will not do this for sure.
(BTW, in this case TapHolds should be just deleted, since TapHold(u8) and Morse(u8) are (almost) the same, and would be redundant.)

Oh yes, the management is complex, maybe a list is not enough, a map like `FnvHashMap<(row,col,layer), Morse>` seems better(but another 16bit overhead here LOL). 

I'm not worry about the Vial insert failure, because the default max_taphold_num should be set to `ROW*COL*NUM_LAYER` for per-key config and `ROW*COL` for per-position, so the space is guaranteed at the compile-time(it's actually same as the other implementations, you cannot even change the needed space). And for smaller MCUs, the user can choose to decrease the max_taphold_num to same memory, then the Vial insert failure is expected by the user.

Now let's see will the feature gate work, I really appreciate your effort ❤️

tib888:
I have added the per key profile setting possibility to Morse.
I think it is ready to review/test/merge, however I have not tested it on HW yet.

I'll open a new PR if I progress with the feature gated per taphold key profile implementation.

docs/en/docs/features/configuration/behavior.md
```diff
@@ -8,32 +8,52 @@ tri_layer = { upper = 1, lower = 2, adjust = 3 }
...
-- `prior_idle_time`: If the previous non-modifier key is released within this period before pressing the current tap-hold key, the tap action for the tap-hold behavior will be triggered. This parameter is effective only when enable_hrm is set to `true`. Defaults to 120ms.
 - `hold_timeout`: Defines the duration a tap-hold key must be pressed to determine hold behavior. If tap-hold key is released within this time, the key is recognized as a "tap". Holding it beyond this duration triggers the "hold" action. Defaults to 250ms.
-- `post_wait_time`: Adds an additional delay after releasing a tap-hold key to check if any keys pressed during the `hold_timeout` are released. This helps accommodate fast typing scenarios where some keys may not be fully released during a hold. Defaults to 50ms
+- `gap_timeout`: Defines the duration a tap-hold key must be released to terminate a morse sequence. Defaults to 250ms.
```
HaoboGu: Move this section under or after `Morse` configuration section? Because for TapHold, the `gap_timeout` will never be used. And the morse behavior is mentioned at line 13, which might confuse reader.

docs/en/docs/features/configuration/behavior.md
```diff
@@ -8,32 +8,52 @@ tri_layer = { upper = 1, lower = 2, adjust = 3 }
...
+# This example enables HRM with all tap-hold features; additionally for home row keys, use the "HRM" profile below,
+tap_hold = { enable_flow_tap = true, prior_idle_time = "120ms",  hold_on_other_press = true, hold_timeout = "250ms", gap_timeout = "250ms" }
+
+# This example enables fast modifiers without HRM
+tap_hold = { enable_flow_tap = false, hold_on_other_press = true, hold_timeout = "200ms", gap_timeout = "200ms" }
 
-# Fast modifiers without HRM
-tap_hold = { enable_hrm = false, hold_on_other_press = true, hold_timeout = "200ms" }
+# This example is the most basic configuration
+tap_hold = { enable_flow_tap = false, hold_timeout = "200ms", gap_timeout = "200ms" }
```
HaoboGu: remove `gap_timeout` for tap_hold?

docs/en/docs/features/configuration/behavior.md
HaoboGu: And finally I tend to remove `tap_hold` section completely, all the related settings should be set in `[morse]` subtable. And in the document, the title could be `Morse(Tap-Hold/TapDance)`

docs/en/docs/features/configuration/behavior.md
```diff
@@ -8,32 +8,52 @@ tri_layer = { upper = 1, lower = 2, adjust = 3 }
 one_shot = { timeout = "1s" }
 ```
 
-## Tap Hold
+## Tap Hold, Key profiles
 
-In the `tap_hold` sub-table you can configure tap-hold behavior which performs one action when tapped and another action when held.
+In the `tap_hold` sub-table you can configure the default tap-hold/morse behavior.
```
HaoboGu: Add a reference or an explanation for morse behavior here would be fine, here is the first time the reader knows "morse"

docs/en/docs/features/configuration/behavior.md
```diff
@@ -196,7 +216,7 @@ morses = [
         { pattern = "-...", action = "B" },
         { pattern = "-.-.", action = "C" },
         { pattern = "-..", action = "D" },
-    ] },
+    ], profile = "MRZ" },
```
HaoboGu, line 171: MRZ doesn't appear in this file, maybe `M`?

docs/en/docs/features/configuration/behavior.md
```diff
@@ -212,11 +232,13 @@ Mixing fields from different methods in the same definition is not allowed.
...
-### Common configuration
+### Fine tuning
 
-The following setting applies to all three definition methods:
+To fine tune the timing, tap hold behavior, it is possible to set the tap hold profile of any morse key regardless of its configuration style:
+ - `profile` : refers a profile name from `[behavior.tap_hold_profiles]`
 
-  - `timeout`: The time window (in milliseconds or seconds) within which taps are considered part of the same morse sequence. Defaults to 200ms if not specified.
+ If this is not set then we look for positional tap hold profile settings (see `matrix_map`).
+ If that is also missing at that kay position, then the defaults will be applied from `[behavior.tap_hold]`.
```
HaoboGu:
```suggestion
 If that is also missing at that key position, then the defaults will be applied from `[behavior.tap_hold]`.
```

docs/en/docs/features/configuration/layout.md
```diff
@@ -15,9 +15,39 @@ matrix_map = """
...
+The `matrix_map` is a string built from `(row, col, <hand> : <profile_name>)` tuples, listed in the same order as you want to define your keys in your key map. The `(row, col)` coordinates are using zero based indexing and referring to the position in the "electronic matrix" of your keyboard. As you can see in [matrix configuration](keyboard_matrix.md), even the direct pin based keyboards are represented with a matrix. In case of split keyboards, the positions refer to the position in the "big unified matrix" of all split parts. 
+With the help of this matrix map, the configuration of non-regular key matrices can be intuitively arranged in your key maps. (Triple quote mark `"""` is used to limit multi-line strings)
+
+The folowing optional fields are used mainly if `enable_flow_tap = true`:
```
HaoboGu:
```suggestion
The folowing optional fields are used mainly if `enable_flow_tap = true` and for keys have `unilateral_tap = true` setting:
```

docs/en/docs/features/configuration/layout.md
```diff
@@ -15,9 +15,39 @@ matrix_map = """
...
+The `<hand>` is optional, can be used to associate each key to the left or right hand using the `L` or `R` values (this info is used during unilateral tap processing).
+The `<profile_name>` is also optional, if given, selects a key profile from `behavior.tap_hold_profiles` to override the defaults set in `behavior.tap_hold`.
```
HaoboGu: There is a limitation that `<profile_name` cannot be `L` or `R`, when the users don't want set `<hand>`, it might lead to misinterpreting of the matrix_map string.

The following is an example:

```toml
matrix_map = """
(0, 0, L) (0, 1) (0, 2, R)
"""

[behavior.tap_hold_profiles]
L = { .. }
R = { .. }
```

rmk/src/config/mod.rs
```diff
@@ -67,27 +84,75 @@ impl Default for MorsesConfig {
...
+    /// - If permissive_hold is set, same as QMK's permissive hold:
+    ///   When another key is pressed and released while the current morse key is held,
+    ///   the hold action of current morse key will be triggered
+    ///   https://docs.qmk.fm/tap_hold#tap-or-hold-decision-modes
+    /// - if hold_on_other_press is set - triggers hold immediately if any other non-morse
+    ///   key is pressed while the current morse key is held
+    /// - If permissive_hold and hold_on_other_press are mutually exclusive,
+    ///   you should not set both!
+    #[bits(1)]
+    pub permissive_hold: bool,
```
HaoboGu: Does  

```rust
#[bits(2)]
pub mode: MorseMode
```
work?

if not, add `repr[u8]` to MorseMode and use

```rust
#[bits(2)]
pub mode: u8
```
?

That could avoid that `permissive_hold` and `hold_on_other_press` are set to true at the same time.

rmk/src/config/mod.rs
```diff
@@ -67,27 +84,75 @@ impl Default for MorsesConfig {
...
-    pub prior_idle_time: Duration,
-    /// Default timeout time for tap or hold
-    pub timeout: Duration,
-    /// Default mode
-    pub mode: MorseMode,
+/// configuration for tap-hold and morse (to save some ram, bitfield is used)
+#[bitfield(u32, order = Lsb, defmt = cfg(feature = "defmt"))]
+#[derive(PartialEq, Eq)]
+pub struct TapHoldProfile {
+    /// Whether this profile is filled with valid data, saves an Option, so saves memory
```
HaoboGu: How does it work? `Option` itself doesn't save any memory

rmk/src/config/mod.rs
```diff
@@ -24,18 +25,34 @@ pub struct RmkConfig<'a> {
...
+impl Default for Hand {
+    fn default() -> Self {
+        Self::Unknown
+    }
+}
+
 /// Config for configurable action behavior
 #[derive(Debug, Default)]
-pub struct BehaviorConfig {
+pub struct BehaviorConfig<const ROW_NUM: usize, const COL_NUM: usize> {
```
HaoboGu, line 42: Is it possible to move the `key_info` to `KeyMap`? In this way the generics of `BehaviorConfig` can be removed, and `key_info` in `KeyMap` seems more reasonable than in `BehaviorConfig`.

docs/en/docs/features/configuration/layout.md
tib888: that is why I replaced the `,` to `:` as separator for the profile name.
So, both (0, 2, R: R) and (0, 2: R) is working.

rmk/src/config/mod.rs
tib888: `Option<TapHoldProfile>` eats more memory than `TapHoldProfile` itself, that is why the is_filled field  was added.
Manual packing instead of bitfields solved this (and a few more things) - now every field in MorseProfile is optional. So the user can override each one separately.

rmk/src/config/mod.rs
tib888: First I also wanted to do that to save the generic parameters, but something has blocked it. Finally decided to put it where it belongs (beside the behavior configuration). This way the implementation was easier.

Anyway, if you are able to move it beside the keymaps, I am not against that.

docs/en/docs/features/configuration/behavior.md
tib888: I have moved behavior.tap_hold into behavior.morse, and behavior.tap_hold_profiles into behavior.morse.profiles.

docs/en/docs/features/configuration/behavior.md
tib888: fixed

rmk/src/config/mod.rs
tib888: I solved this by manually implememting the bit packing.

tib888:
The review fixes are ready!

docs/en/docs/features/configuration/layout.md
tib888: fixed

docs/en/docs/features/configuration/behavior.md
tib888: fixed

docs/en/docs/features/configuration/behavior.md
tib888: since this has moved to [behavior.morse], there makes sense

docs/en/docs/features/configuration/behavior.md
tib888: the explanation of these "fine tuning" settings has been moved down, under the morse explanation

rmk/src/config/mod.rs
HaoboGu: okay, let's get it merged first. That's not blocking issue

HaoboGu:
great! I will merge it after testing it, thanks for your work!

tib888:
> great! I will merge it after testing it, thanks for your work!

It is always a pleasure when I get chance to to work in Rust.
Thank you for the great infrastructure and the great support!

docs/en/docs/features/configuration/behavior.md
```diff
@@ -237,6 +203,82 @@ Note that the Vial-style method (using `tap`, `hold`, `hold_after_tap`, `double_
...
+Each key profile has an associated name, which may be referred 
+- from the layout.matrix_map (the name is case sensitive), to override the defaults in certain key positions
+- the Morse keys may also have their per key profile overrides (which is stronger than the positional override) by setting the `profile` field.
+
+The following examples are the typical default configurations:
+
+```toml
+# This example enables HRM with all tap-hold features; additionally for home row keys, use the "HRM" profile below,
+[behavior.morse]
+enable_flow_tap = true,
```
HaoboGu, line 252: Remove the ","s?

HaoboGu: I'll do it

HaoboGu:
It works well for me. Thanks again!

PR #538, opened 2025-08-26
feat: Clear Layout Flag

### LINKED ISSUE #536: Feature Request: Clear Layout Flag
When debugging you have to set `clear_storage = true` to update the layout by flashing.

However this is very cumbersome in a bluetooth split setup as the pairing information is cleared too.
It would be nice if there was a separate flag for clearing the layout only.

Taym95:
fix #536

### DIFF
--- a/docs/en/docs/features/configuration/storage.md
+++ b/docs/en/docs/features/configuration/storage.md
@@ -15,4 +15,8 @@ num_sectors = 2
 # Set it to true will reset the storage(including keymap, BLE bond info, etc.) at each reboot.
 # This option is useful when testing the firmware.
 clear_storage = false
+# Clear layout at keyboard boot.
+# Set it to true will reset the layout(including keymap, encoder map, behavior config, etc.) at each reboot.
+# This option is useful when testing the firmware without losing bluetooth pairing informations.
+clear_layout = false
 ```


rmk/src/storage/mod.rs
```diff
@@ -782,6 +784,10 @@ pub struct Storage<
     pub(crate) flash: F,
     pub(crate) storage_range: Range<u32>,
     pub(crate) buffer: [u8; get_buffer_size()],
+
+    factory_keymap: [[[KeyAction; COL]; ROW]; NUM_LAYER],
```
HaoboGu: I prefer to not persist the `factory_*` in RAM because it increases memory usage [a lot](https://github.com/HaoboGu/rmk/pull/538#issuecomment-3226020815)(see the report of use_config/nrf52840_ble). Instead, I think the better way is check the `clear_layout` at the startup and save the keymap if `clear_layout` is true. Just like what `clear_storage` does.

For the run-time reset, I think a way loads the keymap from flash(not ram) is possible, but now checking it at the startup is enough imo.

Taym95: I also thought about this but I was not sure, I ignored ResetLayout at runtime, thanks!

HaoboGu (approved):
lgtm now, thanks a lot! I'll test it and merge it soon.

HaoboGu:
Could you also add some doc at `docs/en/docs/features/configuration/storage.md`?

HaoboGu:
It works great! Thanks for your effort! Would you be willing to add some docs, or I can merge it just as is?

Taym95:
> docs/en/docs/features/configuration/storage.md

Added doc in `docs/en/docs/features/configuration/storage.md`

HaoboGu:
awesome, thanks again!

PR #541, opened 2025-08-29
Bidirectional matrix scanning

archnode:


HaoboGu:
I think a separate `BidirectionalMatrix` which implements `Matrix` trait should be added. That's why `Matrix` trait exists: in this way the complex feature gate trick could be avoided and the users can still choose their matrix easily.

archnode:
Thank you for taking a look. You're right - I thought I could get away with feature gates, but I also had to change the signature a bit, which makes a separate `BidirectionalMatrix` a much better solution.

I updated my implementation and will try to complete and test it with my prototype in the next days.

archnode:
I changed the implementation a bit to simplify the pin setup and add a "scan matrix" that explicitly specifies the scanned pairs (this is also how [PRK](https://github.com/picoruby/prk_firmware) does it). I think its good to make the matrix scanning a bit more explicit, this also allows for a more flexible pin setups.

- Currently the new matrix seems to work fine with my prototype.
- the scan process currently is very simple and switches each pin through input / output each read. We could probably improve on this by pre-sorting matching output-pins and only conditionally switching.
- Async matrix functionality currently missing
- Flex pin implementation currently only for RP2040

Example configuration: https://github.com/archnode/weaver-keyboard-rmk/blob/main/src/main.rs

rmk/src/bidirectional_matrix.rs
```diff
@@ -0,0 +1,129 @@
...
+> {
+    /// Input pins of the pcb matrix
+    pins: [Pin; PIN_NUM],
+    /// Debouncer
+    debouncer: D,
+    /// Key state matrix
+    key_state: [[KeyState; COL]; ROW],
+    /// Start scanning
+    scan_start: Option<Instant>,
+    /// Current scan pos: (out_idx, in_idx)
```
HaoboGu: Should be (row, col)? or (col, row)? I'm not sure

rmk/src/driver/flex_pin/mod.rs
```diff
@@ -0,0 +1,11 @@
+use embedded_hal::digital::{ErrorType, InputPin, OutputPin};
+
+#[cfg(feature = "rp2040_bl")]
+pub mod flex_pin_rp;
```
HaoboGu: I prefer to use `rp` as the mod name.

archnode: Changed naming and added an implementation for nrf.

rmk/src/bidirectional_matrix.rs
archnode: When reporting the event its (row, col) - in this instance the "key_state" variable mirrors the "scan_map" and keyboard layout directly. This is different than in the regular matrix where it follows the col2row logic. I'd keep it like this to avoid indirection.

archnode:
Thank you for your review. I briefly commented on the issues.

One caveat with the whole thing is mcu specific code for FlexPin probably doesn't fit the overall structure all that well (and thus the feature gate is pretty arbitrary). Since embedded-hal [removed their FlexPin implementation](https://github.com/rust-embedded/embedded-hal/issues/340) because of stm32 compatibility we could move the mcu-specific FlexPin implementations to macros / keyboard specific code, as the implementation is trivial.

rmk/src/bidirectional_matrix.rs
HaoboGu: ah, I mean the `(out_idx, in_idx)` in comment might be confusing, in the matrix scanning, `scan_pos` is always `(x, y)`, which should be (col_idx, row_idx), am I right?

HaoboGu:
Yeah, without `IoPin` we have to use our own trait. For me I prefer to provide the MCU specific implementation behind corresponding feature gate(aka `rp`/`nrf`/`stm32`).

rmk/src/bidirectional_matrix.rs
archnode: Yes, you are! - The naming of the variables was off. I aligned this with x/y.

archnode:
I removed all async_matrix code, as I don't see a way to do this in a bidirectional matrix.

PR #544, opened 2025-09-05
ci(libssl): improve libssl installation performance with action/cache@v4

linrongbin16:
In "user_build.yml", wget libssl is too slow, use "action/cache@v4" to speed it up.

Tested here: https://github.com/linrongbin16/rmk-keyboards-nrfmicro/actions/runs/17485812490/job/49664467685

image


HaoboGu:
Thanks a lot!

PR #549, opened 2025-09-09
Per key morse profile for TapHold keys

tib888:
@HaoboGu, the initial draft of per key morse profile for TapHold keys with feature gate is ready (documentation is not yet updated).

Please review if it is worth to work on this more or not! *(I definitively do not need this...)

Two things to consider: 

## First:
```rust
pub enum KeyAction {
    /// No action. Serialized as 0x0000.
    No,
    /// Transparent action, next layer will be checked. Serialized as 0x0001.
    Transparent,
    /// A single action, such as triggering a key, or activating a layer. Action is triggered when pressed and cancelled when released.
    Single(Action),
    /// Don't wait the release of the key, auto-release after a time threshold.
    Tap(Action),
    /// Tap hold action
    #[cfg(not(feature = "per_key_profile"))]
    TapHold(Action, Action, ()), //if per_key_profile is not enabled, MorseProfile is replaced with a dummy `()` to eliminate the RAM consumption of MorseProfile
    #[cfg(feature = "per_key_profile")]
    TapHold(Action, Action, MorseProfile),

    /// Morse action, references a morse configuration by index.
    Morse(u8),
}
```
Since TapHold needs MorseProfile, it was moved to `rmk_types` together with MorseMode. 
I do not know the reason, why `rmk_types` were separated, so I do not know if this is a problem or not...

## Second:
I do not know how would it be possible to synchronize the usage of `per_key_profile` feature in all relevant crates?
Could you help with this?
(rmk_types, rmk_macro, rmk_config, rmk, rmk tests should all get (or not get) the `per_key_profile` feature together.)

### DIFF
--- a/.github/workflows/bloat.yml
+++ b/.github/workflows/bloat.yml
@@ -139,6 +139,7 @@ jobs:
             - pi_pico_w_ble_split
           example_type:
             - use_config
+            - use_rust
       outputs:
         size-results: ${{ steps.collect-sizes.outputs.size-results }}
       steps:
--- a/rmk/Cargo.toml
+++ b/rmk/Cargo.toml
@@ -123,6 +123,9 @@ storage = ["dep:embedded-storage", "dep:embedded-storage-async"]
 ## If your PCB diode's direction is col2row, enable this feature. If it's row2col, disable this feature by `default-features = false`.
 col2row = []
 
+## Enable bidirectional feature
+bidirectional = []
+
 ## Enable defmt feature
 defmt = [
     "dep:defmt",
--- a/rmk/src/bidirectional_matrix.rs
+++ b/rmk/src/bidirectional_matrix.rs
@@ -1,0 +1,116 @@
+use embassy_time::{Instant, Timer};
+
+use crate::{debounce::{DebounceState, DebouncerTrait}, matrix::{KeyState, MatrixTrait}};
+use crate::event::{Event, KeyboardEvent};
+use crate::input_device::InputDevice;
+use crate::driver::flex_pin::FlexPin;
+
+pub enum ScanLocation {
+    Pins(usize, usize),
+    Ignore
+}
+
+/// Matrix is the physical pcb layout of the keyboard matrix.
+pub struct BidirectionalMatrix<
+    Pin: FlexPin,
+    D: DebouncerTrait,
+    const PIN_NUM: usize,
+    const ROW: usize,
+    const COL: usize,
+> {
+    /// Input pins of the pcb matrix
+    pins: [Pin; PIN_NUM],
+    /// Debouncer
+    debouncer: D,
+    /// Key state matrix
+    key_state: [[KeyState; COL]; ROW],
+    /// Start scanning
+    scan_start: Option<Instant>,
+    /// Current scan pos: (col_idx, row_idx)
+    scan_pos: (usize, usize),
+    /// Scan map
+    scan_map: [[ScanLocation; COL]; ROW]
+}
+
+impl<
+    Pin: FlexPin,
+    D: DebouncerTrait,
+    const PIN_NUM: usize,
+    const ROW: usize,
+    const COL: usize,
+> BidirectionalMatrix<Pin, D, PIN_NUM, ROW, COL>
+{
+    /// Create a matrix from input and output pins.
+    pub fn new(pins: [Pin; PIN_NUM], debouncer: D, scan_map: [[ScanLocation; COL]; ROW]) -> Self {
+        BidirectionalMatrix {
+            pins,
+            debouncer,
+            key_state: [[KeyState::new(); COL]; ROW],
+            scan_start: None,
+            scan_pos: (0, 0),
+            scan_map
+        }
+    }
+}
+
+impl<
+    Pin: FlexPin,
+    D: DebouncerTrait,
+    const PIN_NUM: usize,
+    const ROW: usize,
+    const COL: usize,
+> InputDevice for BidirectionalMatrix<Pin, D, PIN_NUM, ROW, COL>
+{
+    async fn read_event(&mut self) -> crate::event::Event {
+        loop {
+            let (scan_x_start, scan_y_start) = self.scan_pos;
+            
+            // Scan following the scan map and send report
+            // Loop through rows.
+            for scan_y_idx in scan_y_start..self.scan_map.len() {
+                // Loop trough cols.
+                let scan_x_start_current = if scan_y_idx == scan_y_start { scan_x_start } else { 0 };
+                for scan_x_idx in scan_x_start_current..self.scan_map[scan_y_idx].len() {
+                    if let ScanLocation::Pins(in_idx, out_idx) = self.scan_map[scan_y_idx][scan_x_idx] {
+                        let [in_pin, out_pin] = self.pins.get_disjoint_mut([in_idx, out_idx]).unwrap();
+                        // Set output pin to high.
+                        out_pin.set_as_output();
+                        out_pin.set_high().ok();
+                        Timer::after_micros(1).await;
+                        
+                        // Check input pin and debounce
+                        let debounce_state = self.debouncer.detect_change_with_debounce(
+                            scan_y_idx,
+                            scan_x_idx,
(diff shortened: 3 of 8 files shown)


HaoboGu:
> Since TapHold needs MorseProfile, it was moved to rmk_types together with MorseMode.
I do not know the reason, why rmk_types were separated, so I do not know if this is a problem or not...

Those types were moved to `rmk-types` because I want them to be shared between the keyboard-side(firmware) and the host-side(GUI tools like Vial, or CLI tools).

> I do not know how would it be possible to synchronize the usage of per_key_profile feature in all relevant crates?
Could you help with this?

If we do this, the feature should be added to `rmk-types`, `rmk` and `rmk-config`. 

From #537 I can tell that memory usage increases for about `COL*ROW*8` bytes. So introducing per-key morse profile might lead to `LAYER*COL*ROW*8` bytes, which is acceptable for me.

tib888:
Does that mean that I should simply remove the feature gate?

HaoboGu:
i think so

tib888:
The serialization, documentation is not ready, but the tests pass now, so please check the bloat now!
I hope for `LAYER_NUM * COL * ROW * 4` increase or less, but that depends on the alignment...

HaoboGu:
It's quite surprising that `.bss` in `use_config/nrf52840_ble` increases 3.22KiB, while `.bss` in other examples increases about 672 Bytes.. It's strange because the size of keymap is 5 * 4 * 2 vs 4 * 3 * 2.

I guess it's https://github.com/rust-lang/rust/issues/62958 again..

tib888:
In that example the keymap is larger, but still that is not normal.

I have double checked the (typical) real increase of used memory below with `test mem usage change`.
and .bss changed from +672 to +960 when I increased from 2 to 4 layers of 3x4 key.

`(960-672)/(3*4*2) `= 12 bytes used per key which correct, since the sizeof(TapHold) =  sizeof(Action) + sizeof(Action) + sizeof(MorseProfile) + tag.

tib888:
Also I wanted to ask:
In ZMK the keymaps are collapsed from a 2D array into a 1D array, which contains only the used keys (with something similar to our matrix_map). So each key position has a single index instead of XY coordinates.
I like that approach. Don't we want to do that too? (Of course that would mean that the matrix_map will became obligatory even in use_rust mode).

HaoboGu:
Yes, other examples except `use_config/nrf52840_ble` seems correct. 

> In ZMK the keymaps are collapsed from a 2D array into a 1D array, which contains only the used keys (with something similar to our matrix_map). So each key position has a single index instead of XY coordinates.

How to restore the index to (row, col) pair? is the index like `row * NUM_COL + col`? If so it's like 2D array vs 1D array with extra index, I'm not sure which is better.

tib888:
https://zmk.dev/docs/config/layout

It is not only flattening a 2D Array to 1D, but they really include the matrix_map in the compiled code, so after matrix scan every event is converted into this 1D index space (which contains **ONLY** the used keys). 

If for some reason the index to XY conversion is needed, the included matrix_map can be used.

In our case that would mean to add a sequence_number to each positional key_info, so key_info[x,y].sequence_number could be used to convert the matrix scan events to 1D.

The backward conversion is likely needed only for Vial, but I'm not sure.

HaoboGu:
Yeah, I mean the 1D array is like a matrix map: you save only the used keys, but for each used keys you have to add some overhead(index, or sequence_number, something like that).

In Rust we can even use a `HashMap<Pos, KeyAction>` for it, I'm fine for that. But that means introducing another const generics, aka the total number of keys. `ROW/COL` cannot be eliminated in current implementation if I'm correct.

tib888:
The main advantage of ZMK solution is that in use_rust case you could have the same nicely arranged layout, like the use_config case with matrix_map: No unused key, no row jumps, because the pinky is shorter, nice arrangement even for nonstandard matrixes, etc.

If we will have positional combos once, they will be able to refer positions with a single number:

image

For that one mapping array the <COL, ROW> is still needed, yes, but not for the rest.
And this mapping array could be compile time constant - so no ram needed.

HaoboGu:
I have no opinion on the keymap storage, so both are fine to me. But I'm thinking that is there better way in Rust. 

The HashMap way is one, and I am considering to create another proc_macro that reuses current implementation and can be used in use_rust examples.

For example:

```rust
// keymap.rs

#[keymap]
mod key {
  rows = 5
  cols = 4
  layers = 3
  matrix_map = """
  (0,0) (0,1) (0,2) (0,3)
  (1,0) (1,1) (1,2) (1,3)
  (2,0) (2,1) (2,2)
  (3,0) (3,1) (3,2) (3,3)
     (4,0)    (4,1)
  """
  keymap = [
    """
    NumLock KpSlash KpAsterisk KpMinus
    Kp7     Kp8     Kp9        KpPlus
    Kp4     Kp5     Kp6
    Kp1     Kp2     Kp3        Enter
        Kp0         KpDot
    """,
    ...
  ]
}

``` 

this is just an idea, but reusing the parsing code to generate keymap at the compile-time and reduce the effort writing keymap in Rust code is great.

tib888:
Great idea, would work for both kinds of keymap arrays (just do not forget to include there aliases, layer names too!)

But before that, my next plan is a small refactor to allow usage of layer names and aliases in all places of the toml (including combo, fork config, etc.)

tib888:
I think this PR is ready for review/testing/merge!

The only problem I expect:
The storage / KeymapKey implementation will lose this profile info, since it is based on vial action conversion.
I can't fix that if that is a real problem, its up to you, @HaoboGu. Also there is no unit test for TapHold to storage serialization yet.

rmk-types/src/action.rs
```diff
@@ -59,9 +59,163 @@ impl EncoderAction {
...
+                    Some(MorseMode::Normal) => 0xC000_0000,
+                    Some(MorseMode::HoldOnOtherPress) => 0x8000_0000,
+                    Some(MorseMode::PermissiveHold) => 0x4000_0000,
+                    None => 0,
+                },
+        )
+    }
+
+    /// If the key is pressed longer than this, it is accepted as `hold` (in milliseconds)
+    /// /// if given, should not be zero
```
HaoboGu:
```suggestion
    /// if given, should not be zero
```

HaoboGu:
For the Vial issue, could position based config be used(if it's enabled) in this case?

tib888:
> For the Vial issue, could position based config be used(if it's enabled) in this case?

Position based config always available and if no per key config found, that is the next priority.
I already included this hint in the documentation:

```
::: tip
If you want to use Vial, the positional profile override is recommended instead of per key override, since Vial does not support the profile configuration of tap hold keys, so the profile information will likely be lost.
:::
```

My worry is that a serialization/deserialization round trip may run even if you do not connect to vial, so that has to be checked.

HaoboGu:
Oh yes, that ser/de cycle needs to be tested

HaoboGu:
I still want to know why the flash/ram usage increases that much for `use_config/nrf52840_ble`. Could you check if for `use_rust/nrf52840_ble` example, too? If it's https://github.com/rust-lang/rust/issues/62958 again, then there's workaround by using reference to reduce the mem/flash usage.

tib888:
I have expanded the source of use_config/nrf52840_ble:

``` rust
        let mut per_key_config = ::rmk::config::PerKeyConfig::new(
            ::core::option::Option::None,
        );
        let mut default_keymap = get_default_keymap();
        let mut encoder_keymap = get_default_encoder_map();
        let (keymap, mut storage) = ::rmk::initialize_encoder_keymap_and_storage(
                &mut default_keymap,
                &mut encoder_keymap,
                flash,
                &rmk_config.storage_config,
                &mut behavior_config,
                &mut per_key_config,
            )
            .await;
        let mut keyboard = ::rmk::keyboard::Keyboard::new(&keymap);
```

It seems, both keymap and behavior config is already passed by reference.
So the fix you have done is still there (https://github.com/HaoboGu/rmk/pull/521/commits/101c1f7c351d86246691ff408ae80975e2d3ae9b).

I do not know what else should I look for?

HaoboGu:
nvm, my thought is wrong. Great work, I'm merging it!

PR #552, opened 2025-09-16
Update examples, Add sf32lb52x_usb example

decaday:
# py32

This PR fixes and updates the py32f07x example, upgrading it to py32-hal 0.4.0 to support the new Embassy crates.

# sf32

Added a new sf32lb52x_usb example. Currently, [sifli-rs](https://github.com/OpenSiFli/sifli-rs) does not support Bluetooth.  When Bluetooth support becomes available, this example will be updated or a new one created.

# flip-link

Except ESP32, almost all other examples set flip-link in `build.rs`:

https://github.com/HaoboGu/rmk/blob/f4e13a7708742000755b997619adc22042928012/examples/use_rust/nrf52840_ble/build.rs#L53-L54

However, this is actually ineffective: (`cargo nm`)

```text
200022b4 B _stack_end
20004000 A _stack_start
````

The flip-link setting needs to be placed in `.cargo/config.toml`:

```text
# 20003ffc B _stack_end
# 20001d48 A _stack_start
```

I have updated all examples that previously set flip-link in `build.rs`.

### DIFF
--- a/examples/use_config/nrf52832_ble/.cargo/config.toml
+++ b/examples/use_config/nrf52832_ble/.cargo/config.toml
@@ -1,5 +1,7 @@
 [target.'cfg(all(target_arch = "arm", target_os = "none"))']
 runner = "probe-run --chip nRF52832_xxAA"
+
+# Use flip-link overflow check: https://github.com/knurling-rs/flip-link
 linker = "flip-link"
 
 [build]
--- a/examples/use_config/nrf52832_ble/build.rs
+++ b/examples/use_config/nrf52832_ble/build.rs
@@ -52,9 +52,6 @@ fn main() {
 
     // Set the extra linker script from defmt
     println!("cargo:rustc-link-arg=-Tdefmt.x");
-
-    // Use flip-link overflow check: https://github.com/knurling-rs/flip-link
-    println!("cargo:rustc-linker=flip-link");
 }
 
 fn generate_vial_config() {
--- a/examples/use_config/nrf52840_ble/.cargo/config.toml
+++ b/examples/use_config/nrf52840_ble/.cargo/config.toml
@@ -1,5 +1,7 @@
 [target.'cfg(all(target_arch = "arm", target_os = "none"))']
 runner = "probe-rs run --chip nRF52840_xxAA"
+
+# Use flip-link overflow check: https://github.com/knurling-rs/flip-link
 linker = "flip-link"
 
 [build]
--- a/examples/use_config/nrf52840_ble/build.rs
+++ b/examples/use_config/nrf52840_ble/build.rs
@@ -52,9 +52,6 @@ fn main() {
 
     // Set the extra linker script from defmt
     println!("cargo:rustc-link-arg=-Tdefmt.x");
-
-    // Use flip-link overflow check: https://github.com/knurling-rs/flip-link
-    println!("cargo:rustc-linker=flip-link");
 }
 
 fn generate_vial_config() {
--- a/examples/use_config/nrf52840_ble_split/.cargo/config.toml
+++ b/examples/use_config/nrf52840_ble_split/.cargo/config.toml
@@ -1,5 +1,7 @@
 [target.'cfg(all(target_arch = "arm", target_os = "none"))']
 runner = "probe-rs run --chip nRF52840_xxAA"
+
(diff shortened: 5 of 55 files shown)


decaday:
I noticed that the py32f07x example underwent many changes during rmk's updates,which caused multiple compile errors,and the CI currently skips py32f07x:
[https://github.com/HaoboGu/rmk/blob/f4e13a7708742000755b997619adc22042928012/scripts/check\_all.sh#L10-L12](https://github.com/HaoboGu/rmk/blob/f4e13a7708742000755b997619adc22042928012/scripts/check_all.sh#L10-L12)

At the same time, waiting for all HALs to synchronize during RMK releases is also quite cumbersome.

I suggest that after releasing RMK 0.7.9, examples using non-embassy official HALs (such as py32 and sf32 examples) should lock to a specific RMK version instead of following master (as lagging examples).
They can then be updated periodically by the example creators (such as myself) or contributors with a proper test environment.

If you think this approach is fine, I will update these two examples after RMK 0.7.9 is released.

HaoboGu:
Thanks for the work!

The next version will be v0.8.0, yeah non-official embassy HALs have different release cycle, I agree that their examples can be locked to a specific version. But it's better to have a README.md to tell users about it.

examples/use_rust/py32f07x/src/main.rs
```diff
@@ -68,26 +69,32 @@ async fn main(_spawner: Spawner) {
...
     let debouncer = DefaultDebouncer::<ROW, COL>::new();
-    // let mut matrix = Matrix::<_, _, _, ROW, COL>::new(input_pins, output_pins, debouncer);
-    let mut matrix = rmk::matrix::TestMatrix::<ROW, COL>::new();
+    let mut matrix = Matrix::<_, _, _, ROW, COL>::new(input_pins, output_pins, debouncer);
     let mut keyboard = Keyboard::new(&keymap);
 
+    // if let Some(_) = sifli_hal::rcc::get_clk_dll2_freq() {
+    //     panic!();
+    // }
+
```
HaoboGu, line 97:
```suggestion

```

examples/use_rust/sf32lb52x_usb/src/main.rs
```diff
@@ -0,0 +1,111 @@
...
+
+use rmk::channel::EVENT_CHANNEL;
+use rmk::config::{BehaviorConfig, KeyboardUsbConfig, PerKeyConfig, RmkConfig, VialConfig};
+use rmk::debounce::default_debouncer::DefaultDebouncer;
+use rmk::futures::future::join3;
+use rmk::input_device::Runnable;
+use rmk::keyboard::Keyboard;
+use rmk::matrix::Matrix;
+// use rmk::storage::async_flash_wrapper;
+// use rmk::{initialize_keymap_and_storage, run_devices, run_rmk};
```
HaoboGu, line 27:
```suggestion
```

HaoboGu (commented):
Having a readme which includes how to use the sf32lb52 example(for example, `ftab.bin` usage) would be great

examples/use_rust/sf32lb52x_usb/.cargo/config.toml
```diff
@@ -0,0 +1,16 @@
+[target.'cfg(all(target_arch = "arm", target_os = "none"))']
+runner = 'sftool -c SF32LB52 --port COM10 write_flash ftab.bin@0x12000000'
```
HaoboGu: Use a placeholder instead of `COM10`

examples/use_rust/py32f07x/src/main.rs
decaday: Thanks. I have forgot them..

examples/use_rust/sf32lb52x_usb/src/main.rs
decaday: I think maybe we can keep them and re-enable once sifli-hal completes the MPI driver,  
since Bluetooth will likely? need to use flash (MPI) in the future to store pairing information

HaoboGu: okay, that's fair

HaoboGu:
Thanks a loooooot!

PR #555, opened 2025-09-20
fix: incorrect use of `Display` instead of `Debug`

pcasotti:
This is preventing compilation when defmt feature is disabled


HaoboGu:
Thanks!

PR #556, opened 2025-09-20
Custom controllers execution mode

pcasotti:
We should allow the user to choose the execution mode of custom controllers defined via `#[controller]`. I propose to change the attribute to be either `#[controller(event)]` or `#[controller(poll)]` so we know what method to call when expanding the controller tasks.

### DIFF
--- a/docs/en/docs/features/controller.md
+++ b/docs/en/docs/features/controller.md
@@ -111,14 +111,19 @@ scrolllock.low_active = false
 
 ### Custom Controllers
 
-Custom controllers are declared using the `#[controller]` attribute within your keyboard module:
+Custom controllers are declared using the `#[controller(event)]` or `#[controller(poll)]` attribute within your keyboard module.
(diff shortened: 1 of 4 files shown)


docs/en/docs/features/controller.md
```diff
@@ -111,14 +111,14 @@ scrolllock.low_active = false
 
 ### Custom Controllers
 
-Custom controllers are declared using the `#[controller]` attribute within your keyboard module:
+Custom controllers are declared using the `#[controller]` attribute within your keyboard module. The execution mode must be specified with either `event` or `poll` parameter:
```
HaoboGu: Add some explanation of the two modes?

HaoboGu:
lgtm! Just need a few more doc for "event" and "poll" modes of the controller

pcasotti:
Also added verification for the controller feature when using the `[light]` option

HaoboGu (approved):
thanks a lot!

PR #572, opened 2025-10-10
feat(split): send indicator and active layer state to split peripherals

hdwong:
Enhance split keyboard support by transmitting complete LED indicator state and active layer information from central to peripherals, enabling peripherals to control their own LEDs through the controller system.


HaoboGu:
Oh I found that the `SPLIT_MESSAGE_PUBLISHER` is actually not needed, it achieves similar functionality as `Controller`. We can take the peripheral as a controller subscriber of central. Let me change it first.

HaoboGu:
I've removed `SPLIT_MESSAGE_PUBLISHER`, use `CONTROLLER_EVENT_CHANNEL` instead, so events are emitted only once. Please rebase

PR #579, opened 2025-10-13
feat(split): Add LED indicator and layer sync from central to peripheral

hdwong:
Add support for syncing keyboard LED indicators and active layer state from central to peripheral in split keyboard configurations.

Changes:
 - Add KeyboardIndicator(u8) and Layer(u8) to SplitMessage enum
 - Forward ControllerEvent::KeyboardIndicator and ControllerEvent::Layer from central to peripheral via split protocol
 - Peripheral receives these messages and republishes them to CONTROLLER_CHANNEL for local controller subscribers

This enables peripheral-side controllers to respond to LED indicator changes (CapsLock, NumLock, ScrollLock, etc.) and layer switches that occur on the central side, allowing synchronized LED feedback and layer-aware behavior across both halves of a split keyboard.


HaoboGu:
thanks!

PR #589, opened 2025-10-20
Fix py32 example

decaday:
This fixes the py32 examples.

In addition, following the discussion in #552, after RMK v0.8.0 release I will lock the py32 and sf32 examples to v0.8.0 and add a note in the README to indicate this.


HaoboGu (approved):
nice, thanks!

PR #591, opened 2025-10-21
docs: redirect 404 page to home page and some bugs fixed

liyang8246:
use "/" instead of join function that produces "\" on windows
fix guide button path
fix logo path


HaoboGu:
❤️

PR #593, opened 2025-10-23
Remove col2row feature gate

### LINKED ISSUE #380: Rethink default features
Currently, the ergonomics around default features are not great. Namely, if one wants to remove a specific default feature, one has to:

1. Set `default-features=false`
2. Find all the features that were default, and add them all back in other than the one they wanted to remove.

The second step is the one I have a problem with. Currently, there are only 3, `defmt`, `col2row`, and `storage`, but I can see that list growing.

I think RMK should try to improve this a bit. First off, `col2row` does not need to be a default feature. The `col2row` feature could be removed, and replaced with a non-default `row2col` feature.

For the other 2 default features, I see 2 solutions, both of which are non-ideal in their own ways, but are hopefully better than what RMK has now:

1. Make all features non-default, but add any 'default' features to the rmk templates that `rmkit` uses. The downside here is that documentation would have to be improved for any users not starting from a template, but as rmk already requires a fairly complex cargo project to be setup anyway, its only a small amount of additional stuff.
2. Remove the non-default features, and replace them with 'negative features', or features that turn off specific functionality rather than turn it on. `defmt` would become `no-defmt`, `storage`, `no-storage`. In a library crate, this is normally a bad idea, as features are unioned together by cargo, so negative features can lead to unexpected compilation failures. Now this isnt really a problem in RMKs current iteration, as the only crate to depend on RMK should be the users keyboard firmware crate, but it may lead to problems down the road if someone starts wrapping RMK or something.

Personally, I favor 1, but id like to have a discussion about this.

HaoboGu:
Should close #380, too.

### DIFF
--- a/rmk/src/ble/battery_service.rs
+++ b/rmk/src/ble/battery_service.rs
@@ -60,9 +60,10 @@ impl<P: PacketPool> BleBatteryServer<'_, '_, '_, P> {
             let battery_state = self.wait_until_battery_state_available().await;
             // Check if there's latest battery state update
             if let BatteryState::Normal(level) = BATTERY_UPDATE.try_take().unwrap_or(battery_state)
-                && let Err(e) = self.battery_level.notify(self.conn, &level).await {
-                    error!("Failed to notify battery level: {:?}", e);
-                }
+                && let Err(e) = self.battery_level.notify(self.conn, &level).await
+            {
+                error!("Failed to notify battery level: {:?}", e);
+            }
         }
     }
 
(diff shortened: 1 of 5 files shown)


willpuckett:
I have traditionally used row2col. When I change to the new col_pins/row_pins setup usIng `row2col = true`, RMK builds but does not produce any output on my board. FWIW, I tried building with toggling row2col, and switching the col/row pins. 

I was able to produce output by ditching `row2col = true` flip-flopping my col_pins/row_pins, and then reversing the coordinates in my matrix map.

It seems the matrix map removes the need for row2col altogether?

HaoboGu:
> It seems the matrix map removes the need for row2col altogether?

Yes, ideally `matrix_map` should not care about row2col/col2row. But it's weird that row2col doesn't work anymore.. Let me try to reproduce the issue

willpuckett:
It seems like having row2col is now redundant with the capability to organize keys in the matrix map... Could it be scrapped altogether?

HaoboGu:
row2col is needed for matrix scanning, but not needed for matrix_map and layer/layout config. so we cannot completely remove this option, but at least matrix map doesn't need it.

PR #603, opened 2025-10-29
fix changed links following rspress migration

willpuckett:


willpuckett:
the boards at https://rmk.rs/guide/user_guide/real_world_examples.html don't seem to be loading completely, either, but I can't determine how rmk.rs is currently being deployed there doesn't seem to be an action for it and the links in 'docs/docs/main/docs/getting_started/real_world_examples.md' seem to be correct...

HaoboGu:
thanks a lot!

PR #623, opened 2025-11-12
Draft: Add LowPowerMatrix

tompre:
This is a work in progress improvement of the default matrix reducing power drain while de-bouncing and while keys are pressed.

First results measured with nrf52840:
- default matrix scan: ~3 mA
- low power scan: 300 - 600 µA

The improvements are two fold:
- sleeping and disabling output pins to between scans during de-bouncing
- Using interrupts while buttons are pressed

There are a lot of trade-offs and customizations possible. While interrupts would allow for low latency detection of key presses and releases, there are some drawbacks. The pressed key closes the circuit which results in current flowing over the pull-up/-down resistors. Depending on their size this might lead to a significant energy loss. Additionally the pressed key will block the detection of presses/releases of other keys connected to the same input pin. This can be somewhat mitigated by always scanning row2col, assuming one row will be only used by one single finger during fast typing. However at some point we need to scan again to detect those key changes.

To achieve row2col scanning with a col2row matrix, this implementation uses inverse polarity (pulling input pin high, setting output pin low), while using normal polarity with a row2col matrix.

To-Does:
- [ ] Check out other interrupt (hybrid) patterns and their efficiency
- [ ] Measure efficiency of using interrupts vs polling when key is pressed
- [ ] Add logic to generate code for central key matrix in split mod

### DIFF
--- a/rmk-config/src/board.rs
+++ b/rmk-config/src/board.rs
@@ -78,7 +78,7 @@ impl KeyboardTomlConfig {
             },
             (Some(m), None) => {
                 match m.matrix_type {
-                    MatrixType::normal => {
+                    MatrixType::normal | MatrixType::low_power => {
                         if m.row_pins.is_none() || m.col_pins.is_none() {
                             return Err("`row_pins` and `col_pins` is required for normal matrix".to_string());
                         }
--- a/rmk-config/src/lib.rs
+++ b/rmk-config/src/lib.rs
@@ -324,6 +324,7 @@ pub enum MatrixType {
     #[default]
     normal,
     direct_pin,
+    low_power,
 }
 
 #[derive(Clone, Debug, Default, Deserialize)]
--- a/rmk-macro/src/gpio_config.rs
+++ b/rmk-macro/src/gpio_config.rs
@@ -1,12 +1,12 @@
 use quote::{format_ident, quote};
 use rmk_config::{ChipModel, ChipSeries};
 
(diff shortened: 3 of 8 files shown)


HaoboGu:
If possible, I want the normal `Matrix` is low-power ready. `async_matrix` is already used for power saving by putting the MCU to sleep when there's no key pressed, so we don't need a separate `LowPowerMatrix`, making current `Matrix` low power friendly would be great.

> Additionally the pressed key will block the detection of presses/releases of other keys connected to the same input pin

Why? In current `Matrix` the interrupt acts like a "trigger", and then the normal matrix scanning is performed, no key changes will be ignored imo. Do I miss something?

tompre:
> Why? In current Matrix the interrupt acts like a "trigger", and then the normal matrix scanning is performed, no key changes will be ignored imo. Do I miss something?

Yes, while waiting for key presses. This works fine with the current implementation and is identical to my proposal.

However, while a key is pressed, the default matrix scan is done permanently and drains ~3mA. I wanted to try to use interrupts or sleeping to reduce the current. I first thought using interrupts is bad, due to current flowing over the pull resistors. However, I suspect that, the gain is not worth the complexity and polling in 4-10 ms intervals might be good enough. In that case, a small adaptation of the default matrix would be enough.

I think it should be possible to achieve this with the default matrix with some small improvements.

tompre:
I will do some tests with the new default matrix and create a new pull request if I find potential energy savings.

PR #629, opened 2025-11-18
feat: change central sleep timeout to be in seconds

pcasotti:
Change the central sleep timeout parameter from minutes to seconds for better customization


HaoboGu:
thanks a lot!

PR #630, opened 2025-11-19
Fix missed keypresses in default matrix scan implementation.

tompre:
The matrix scan algorithm can miss releases in some cases when other tasks take to long.

This happens after a debounce was completed while other keys are still pressed in the following way:
- If a several keys are pressed and a key with a higher (row,col) number is released the event is reported
- The scan is continued from the (row,col) position and will not detect any pressed keys
- If the scan continuation takes longer than 1ms interrupts are enabled and the algorithm goes to sleep while a key is still pressed.

I fixed this by adding a flag `rescan_needed` to detect these situations and block sleep.

To Does:
- [x] I'm not sure if we still need the scan_start field. We can just go to sleep if no rescan is needed. Should I remove it?
- [x] Also adapt the implementation in [split/central.rs](rmk/src/split/central.rs). Should I refactor it so we do not have two copies?
- [x] Should I add feature flags for the rescan logic to only be included when `async_matrix` is enabled?

### DIFF
--- a/rmk/src/matrix.rs
+++ b/rmk/src/matrix.rs
@@ -2,7 +2,7 @@
 use core::pin::pin;
 use core::sync::atomic::Ordering;
 
-use embassy_time::{Instant, Timer};
+use embassy_time::Timer;
 use embedded_hal::digital::{InputPin, OutputPin};
 #[cfg(feature = "async_matrix")]
 use {embassy_futures::select::select_slice, embedded_hal_async::digital::Wait, heapless::Vec};
@@ -172,10 +172,10 @@ pub struct Matrix<
     debouncer: D,
     /// Key state matrix
     key_states: [[KeyState; ROW]; COL],
-    /// Start scanning
-    scan_start: Option<Instant>,
     /// Current scan pos: (out_idx, in_idx)
     scan_pos: (usize, usize),
+    /// Re-scan needed flag
+    #[cfg(feature = "async_matrix")]
     rescan_needed: bool
 }
 
@@ -346,8 +346,8 @@ where
             col_pins,
             debouncer,
             key_states: [[KeyState::new(); ROW]; COL],
-            scan_start: None,
             scan_pos: (0, 0),
+            #[cfg(feature = "async_matrix")]
             rescan_needed: false,
         }
     }
@@ -395,8 +395,6 @@ where
     async fn read_event(&mut self) -> crate::event::Event {
         loop {
             let (out_idx_start, in_idx_start) = self.scan_pos;
-            #[cfg(feature = "async_matrix")]
-            self.wait_for_key().await;
 
             // Scan matrix and send report
             for out_idx in out_idx_start..Self::OUTPUT_PIN_NUM {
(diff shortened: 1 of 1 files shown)


HaoboGu:
> I'm not sure if we still need the scan_start field. We can just go to sleep if no rescan is needed. Should I remove it?

If `scan_start` is not needed anymore, removing it is fine

> Also adapt the implementation in [split/central.rs](https://github.com/HaoboGu/rmk/pull/rmk/src/split/central.rs). Should I refactor it so we do not have two copies?

yes, reducing the duplication would be great 

> Should I add feature flags for the rescan logic to only be included when async_matrix is enabled?

maybe just put it under `async_matrix` feature? we don't need an extra feature gate for it.

rmk/src/matrix.rs
```diff
@@ -478,8 +483,32 @@ where
...
+pub struct OffsetMatrix<
+    const ROW: usize,
+    const COL: usize,
+    M: MatrixTrait<ROW, COL>,
+    const ROW_OFFSET: usize,
+    const COL_OFFSET: usize,
+>(pub M);
+
+impl<const ROW: usize, const COL: usize, M: MatrixTrait<ROW, COL>, const ROW_OFFSET: usize, const COL_OFFSET: usize>
+    InputDevice for OffsetMatrix<ROW, COL, M, ROW_OFFSET, COL_OFFSET>
```
tompre: I did only implement `InputDevice` as it seems to be enough. Should I implement MatrixTrait? I could not find any use of the functions in the examples or in rmk-macro. Is it still needed?

rmk/src/split/central.rs
```diff
@@ -57,531 +47,3 @@ pub async fn run_peripheral_manager<
...
-        if COL2ROW {
-            self.key_states[out_idx][in_idx].toggle_pressed();
-        } else {
-            self.key_states[in_idx][out_idx].toggle_pressed();
-        }
-    }
-}
-
-/// DirectPinMartex only has input pins.
-pub struct CentralDirectPinMatrix<
```
tompre, line 429: I removed `CentralDirectPinMatrix` as well, as `OffsetMatrix` can wrap anything that implements `MatrixTrait`.

Do you want a separate refactoring pull request since this touches also other code?

tompre:
I created a wrapper type `OffsetMatrix` to offset the keyboard events. It can be used with anything that implements `MatrixTrait`.

I think function wise this is ready to merge.

rmk/src/split/central.rs
HaoboGu: Yes, please

rmk/src/matrix.rs
HaoboGu: I think `wait_for_key` should be in the `MatrixTrait`

HaoboGu: It seems no need for `OffsetMatrix` to implement `MatrixTrait`?

HaoboGu: Yeah, the `OffsetMatrix` is just a wrapper, so just implementing the `InputDevice` trait is fine. Maybe it needs to be renamed to something like `OffsetMatrixWrapper`?

HaoboGu:
Thanks a lot! The `OffsetMatrix` is greeeeaaat, I really like it :P

rmk/src/matrix.rs
tompre: `OffsetMatrixWrapper` sounds great, will do.

HaoboGu:
thanks!

PR #631, opened 2025-11-19
Feat/sleep event

pcasotti:
Publish the `ControllerEvent::Sleep` event when entering or waking from sleep mode.

### DIFF
--- a/rmk/src/channel.rs
+++ b/rmk/src/channel.rs
@@ -91,6 +91,6 @@ pub fn send_controller_event(publisher: &mut ControllerPub, event: ControllerEve
 #[cfg(feature = "controller")]
 pub fn send_controller_event_new(event: ControllerEvent) {
     if let Ok(mut publisher) = CONTROLLER_CHANNEL.publisher() {
-        send_controller_event(publisher, event);
+        send_controller_event(&mut publisher, event);
     }
 }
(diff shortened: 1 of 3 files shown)


rmk/src/channel.rs
```diff
@@ -78,8 +78,19 @@ pub(crate) static FLASH_CHANNEL: Channel<RawMutex, FlashOperationMessage, FLASH_
...
 pub fn send_controller_event(publisher: &mut ControllerPub, event: ControllerEvent) {
     info!("Sending ControllerEvent: {:?}", event);
     publisher.publish_immediate(event);
 }
+
+/// Send the specified `event` to `CONTROLLER_CHANNEL`.
+/// Do not use this function if you plan to send events multiple times, use `send_controller_event`
+/// instead for better performance.
+#[cfg(feature = "controller")]
+pub fn send_controller_event_new(event: ControllerEvent) {
```
HaoboGu, line 92: send_controller_event_new -> send_controller_event_once?

HaoboGu:
Now I'm a little worry about the number of controller publisher/subscriber, I don't know how to calculate the maximum number of publishers/subscribers that the project use. Maybe we should manage them in a single place to avoid runtime panic of `unwrap!(CONTROLLER_CHANNEL.publisher());`.

pcasotti:
> Maybe we should manage them in a single place to avoid runtime panic of `unwrap!(CONTROLLER_CHANNEL.publisher());`.

You mean like a global publisher under a mutex? Seems good, but I don't know how the mutex overhead will affect performance, probably not by much.

HaoboGu:
> You mean like a global publisher under a mutex

I mean a manager which knows how many publishers is used and pre-assigns all publishers. IMO the best way is to calculate the number of publishers/subscribers at the compile-time, but I don't know whether it is possible

PR #638, opened 2025-11-25
Fix a bug of Caps Word

adong660:
This PR is a small fix of Caps Word.

According to [QMK](https://docs.qmk.fm/features/caps_word) and [ZMK](https://docs.qmk.fm/features/caps_word) documentations, Caps Word has two set of keys:
- keys that continue Caps Word, including A-Z, 0-9, minus (underscore), backspace, and delete, and
- keys that are shifted if Caps Word is activated now, including only A-Z and minus

This design makes typing capitalized acronyms (like “USA”) and constants (like `FOO_123`) easier.

However, the current implementation only recognizes when Caps Word is active (like, after tapping `KeyCode::CapsWordToggle`), and if so, capitalizes any key pressed. This results in an undesired behavior that the key sequence {CapsWord key, f, o, o, -, 1, 2, 3} is interpreted as “FOO_!@#” but not “FOO_123”.

This PR makes the following changes:
- Adds a boolean member variable for `Keyboard` to mark whether the current pressed key should be shifted
- Updates the variable when Caps Word is active and a key is pressed
- Adds a function for `KeyCode` which decides whether a key belongs to those that can be shifted by Caps Word

### DIFF
--- a/rmk/src/keyboard.rs
+++ b/rmk/src/keyboard.rs
@@ -76,6 +76,80 @@ impl<T> OneShotState<T> {
     }
 }
 
+/// State machine for Caps Word
+#[derive(Debug, Default)]
+enum CapsWordState {
+    /// Caps Word is activated (but may have timed out thus becoming inactive)
+    Activated {
+        /// Time since last key press
+        timer: Instant,
+        /// Whether the current key should be shifted
+        shift_current: bool,
+    },
+    /// Caps Word is deactivated
+    #[default]
+    Deactivated,
+}
+
+impl CapsWordState {
+    /// Caps Word timeout duration
+    const TIMEOUT: Duration = Duration::from_secs(5);
+
+    /// Activate Caps Word
+    fn activate(&mut self) {
+        *self = CapsWordState::Activated {
+            timer: Instant::now(),
+            shift_current: false,
+        };
+    }
+
+    /// Deactivate Caps Word
+    fn deactivate(&mut self) {
+        *self = CapsWordState::Deactivated;
+    }
+
+    /// Toggle Caps Word
+    fn toggle(&mut self) {
+        match self {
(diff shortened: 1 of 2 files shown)


rmk/src/keyboard.rs
```diff
@@ -1402,6 +1407,7 @@ impl<'a, const ROW: usize, const COL: usize, const NUM_LAYER: usize, const NUM_E
                     && timer.elapsed() < Duration::from_secs(5)
                 {
                     self.caps_word_timer = Some(Instant::now());
+                    self.caps_word_shift = key.is_caps_word_shifted_key();
                 } else {
                     self.caps_word_active = false;
                     self.caps_word_timer = None;
```
HaoboGu: Add `self.caps_word_shift = false;` here to avoid any possible wrong state?

HaoboGu (commented):
Oh I might have overlooked it. Thanks for the PR! Only a minor comment from me.

rmk/src/keyboard.rs
adong660: Sure that's better. But since `self.caps_word_timer` matters only when `self.caps_word_active == true`, omitting that should also be OK. Maybe we can do a small refactor to make things clearer, like:
```rust
enum CapsWordState {
    Active {
        timer: Instant,
        shift: bool,
    },
    Inactive,
}

struct Keyboard {
    ...
    caps_word: CapsWordState,
    ...
}
```

HaoboGu: Oh this looks much better! Please do this 😄

adong660: Hey, I just made some changes. Please check it :)

HaoboGu (approved):
Looks good to me, thanks a lot!

PR #639, opened 2025-11-26
nRF52840 define dcdc reg via config

aziddy:
I'm using a custom designed nRF52480 PCB with a external voltage regulator that provides 3v3 to the nRF in normal voltage mode.

My PCB doesn't contain the the necessary LC components on either DCDC regulator. Which the nRF52480 datasheet mentions are required. Only the LDO in REG1 is used

> _"**Note**: Do not enable DC/DC regulator without an external LC filter being connected as this will inhibit device operation, including debug access, until an LC filter is connected."_

Which results in my custom nRF52480 based keyboard to never boot (can't even access through a SWD debugger), unless I utilize the "use_rust" method and disable both DCDC regulators

My changes allow RMK consumers to configure the DCDC regulator settings for their own nRF based boards via the "use config" (keyboard.toml)

---

Rip into the PR how you like. I'm making some strong assertions like adding in a whole new section called "chip" to the keyboard.toml

So if you have any suggestions (especially on naming/abstractions) I'm open to it

### DIFF
--- a/rmk-config/src/lib.rs
+++ b/rmk-config/src/lib.rs
@@ -47,8 +47,8 @@ pub struct KeyboardTomlConfig {
     storage: Option<StorageConfig>,
     /// Ble config
     ble: Option<BleConfig>,
-    /// Chip config
-    chip: Option<ChipConfig>,
+    /// Chip-specific configs (e.g., [chip.nrf52840])
+    chip: Option<HashMap<String, ChipConfig>>,
     /// Dependency config
     dependency: Option<DependencyConfig>,
     /// Split config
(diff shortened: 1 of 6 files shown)


HaoboGu:
I'd like to see `[chip]` section in keyboard.toml, but because RMK supports lots of chips, for the chip-specific config, I think it is better to use something like `[chip.nrf52840]`?

HaoboGu (approved):
lgtm!

rmk-config/src/lib.rs
```diff
@@ -48,14 +47,16 @@ pub struct KeyboardTomlConfig {
     storage: Option<StorageConfig>,
     /// Ble config
     ble: Option<BleConfig>,
+    /// Chip-specific configs (e.g., [chip.nrf52840])
+    chip: Option<HashMap<String, ChipConfig>>,
```
HaoboGu, line 51: I'm thinking about making the supported chip-specific config an enum? Current implementation accepts **any** config in `keyboard.toml`, maybe we should make every chip-specific config a struct?

aziddy: Thats probably a good idea moving forward

PR #646, opened 2025-11-29
chore: expose fields and functions of keyboard

zhuren-plbm:
I change access modifier of some fields and functions of `rmk::keyboard` to be able to use for external crate.


HaoboGu:
Could you explain how do you use those functions?

zhuren-plbm:
My project re-implements `Runnable` trait, so I need to access these fields and functions in `rmk::keyboard::Keyboard` struct. I change them because they are in the following code:
https://github.com/HaoboGu/rmk/blob/0f6980ef8179321e5617cd829c540c6f50810adf/rmk/src/keyboard.rs#L152-L177

Issue #261, opened 2025-02-19
Control GPIO by config


HaoboGu:
https://zmk.dev/docs/keymaps/behaviors/power

HaoboGu:
Implemented, close

PR #651, opened 2025-12-02
feat: static output pin config

tompre:
This pull request adds a config feature for static pins as requested in #261.

Config looks like this, but I'm not sure about the naming:
``` toml
# Config output for normal keyboard or all splits
[[static_output]]
pin = "P0_13"
level_high = false

# Config output for individual splits
[[split.<central|peripheral>.static_output]]
pin = "P0_13"
level_high = false
```

It produces the following code:
``` rust
// for all platforms besides nrf:
::core::mem::forget(::<driver>::Output::new("GPIO021", <Level::Low | Level::High>, OutputConfig::default()))
// for nrf
::embassy_nrf::gpio::Output::new("P0_13", <Level::Low | Level::High>, OutputConfig::default())).persist()
```

### DIFF
--- a/rmk/CHANGELOG.md
+++ b/rmk/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add PMW3610 optical mouse sensor support for nRF and RP2040 with bit-bang SPI
+- Add support for configuring static output pins
 
 ### Changed
 


HaoboGu:
I think it's possible to make the output pin controllable by some events? The static output might not be exactly #261 want. But yeah, making it configurable in `keyboard.toml` is hard, some design is needed

tompre:
Hmm, we could create a controller that listens for the power on/off/toggle events as a first step. The controller setting would look something like this:
``` toml
[[output_pin]]
pin = "P0_13"
low_active = false
alias = "pin13"
default_active = true
```
The controller would than react to the following events, where id corresponds to the id of the output_pin declaration, with alias resolved to make configuration easier:
``` rust
pub enum Event {
    /// Keyboard event
    Key(KeyboardEvent),
    ...
    ExtPower(ExtPowerEvent)
}

pub struct ExtPowerEvent{
    id: u8,
    action: ExtPowerAction,
}

pub enum ExtPowerAction{
    On,
    Off,
    Toggle,
}
```

Than we would need some configuration to map other events/keys/controllers (rough idea):
``` toml
[[event_controller]]
type = "blinking"
input = "<some event name>"  # wakeup in case of the linked issue
timeout = "5ms"
toggle_time = "500ms"
output = "pin13"

 [[user_key]]
id = 15
action = "toggle"
output = "pin13"
```
The translation from trigger event to action event is however complicated to do in a generic way. We could create controllers for specific use cases together with a specific syntax in `keyboard.toml`. But in the end, there is always a limit in complexity that the config can provide.

I would like to have a way to disable the external power permanently to save energy, hence the proposal.
I could adapt it in a way, that it uses the controller and as event is sent at the start to set the pin low.
This controller could than be reused in a more general GPIO config and would be a first step.

HaoboGu:
Yeah, this looks quite complicate.. Maybe we should focus on the "function", for example, keep the output_pin only and add a keycode like `PinToggle0. In this way each output_pin is "binded" to a fixed keycode, eliminates the complexity of binding to different events and controllers

HaoboGu:
I feel that is out-of-the-scope of current PR, what about merging this PR as is and leaving the improvement to the future. 

Oh the documentation and changelog need to be added

tompre:
I did a rework and made the config more general:
```toml
[[output]]
pin = "PIN_13"
initial_state_active = false
low_active = false
```

This would allow us to add an additional field to enable a controller that would react to events.
I added documentation, only a note in the changelog is still missing.

HaoboGu:
Thanks a lot!

PR #659, opened 2025-12-05
Update the alias and docs

hvuitsme:
based on https://github.com/qmk/qmk_firmware/blob/master/quantum/keycodes.h and edited consumer alias to be shorter

### DIFF
--- a/rmk-types/src/keycode.rs
+++ b/rmk-types/src/keycode.rs
@@ -1057,10 +1057,10 @@ impl KeyCode {
             KeyCode::VolD => ConsumerKey::VolumeDecrement, // Audio Volume Dn
             KeyCode::MNxt => ConsumerKey::NextTrack, // Media Next Track
             KeyCode::MPrv => ConsumerKey::PrevTrack, // Media Previous Track
-            KeyCode::MStp => ConsumerKey::StopPlay, // Media Next Track
-            KeyCode::MPly => ConsumerKey::PlayPause, // Media Next Track
-            KeyCode::MSel => ConsumerKey::Record, // Media Next Track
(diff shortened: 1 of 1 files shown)


HaoboGu:
Sorry qmk is gpl licensed so we cannot use code "derived" from qmk. And the full name is preferred in RMK

hvuitsme:
> Sorry qmk is gpl licensed so we cannot use code "derived" from qmk. And the full name is preferred in RMK 

So is there another way to shorten alias?

HaoboGu:
> > Sorry qmk is gpl licensed so we cannot use code "derived" from qmk. And the full name is preferred in RMK 
> 
> So is there another way to shorten alias?
> 
> 

you can define your own alias in keyboard.toml: https://rmk.rs/docs/configuration/layout.html#aliases

is this what you need？

hvuitsme:
> > > Sorry qmk is gpl licensed so we cannot use code "derived" from qmk. And the full name is preferred in RMK 
> > 
> > So is there another way to shorten alias?
> > 
> > 
> 
> you can define your own alias in keyboard.toml: https://rmk.rs/docs/configuration/layout.html#aliases
> 
> is this what you need？

I know there is that but it's less cumbersome to just put an alias in the beginning

HaoboGu:
Or you can add aliases [here](https://github.com/HaoboGu/rmk/blob/main/rmk-config/src/keycode_alias.rs#L20) and keep the keycode unchanged?

edit: there're already some aliases available which can be used directly in keyboard.toml. You can add more if you want. I appreciate your work, would you please open another PR to add the doc only?

PR #660, opened 2025-12-07
feat: PMW3610 Input Device

kot149:
Implemented InputDevice and InputProcessor for PMW3610 optical mosue sensor. The driver is based on [Zephyr's implementation](https://github.com/zephyrproject-rtos/zephyr/blob/d31c6e95033fd6b3763389edba6a655245ae1328/drivers/input/input_pmw3610.c).
Also added bitbang SPI implemetation as PMW3610 requires half-duplex SPI.

## Example Usage

```toml
[[input_device.pmw3610]]
name = "trackball0"
spi.instance = "bitbang0"
spi.sck = "P0_05"
spi.mosi = "P0_04"
spi.miso = "P0_04"
spi.cs = "P0_09"
# or spi = { instance = "bitbang0", sck = "P0_05", mosi = "P0_04", miso = "P0_04", cs = "P0_09" }
motion = "P0_02"

force_awake = false
smart_mode = true
cpi = 800
invert_x = true
invert_y = true
swap_xy = true
```

## Todos
- [x] Verify working with nRF52840
  - [x] on unibody
  - [x] on split central
  - [x] on split peripheral
- [x] Documentation
- [x] Changelog
- [ ] Testing

In the future, refactoring may be needed to share implementation with JoyStick or to handle SPI/I2C configs or generic processing like input transformation for pointing devices.

### DIFF
--- a/rmk/CHANGELOG.md
+++ b/rmk/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add PMW3610 optical mouse sensor support for nRF and RP2040 with bit-bang SPI.
+
+### Changed
+
+- Make `embedded-hal-async` a required dependency.
+
 ## [0.8.1] - 2025-11-25
 
 ### Changed


rmk-config/src/lib.rs
```diff
@@ -703,6 +704,40 @@ pub struct JoystickConfig {
...
+#[derive(Clone, Debug, Default, Deserialize)]
+#[allow(unused)]
+#[serde(deny_unknown_fields)]
+pub struct Pmw3610Config {
+    /// Name of the sensor (used for variable naming)
+    pub name: String,
+    /// SPI clock pin
+    pub sclk: String,
+    /// SPI bidirectional data pin (SDIO)
+    pub sdio: String,
```
HaoboGu: these pins can be extracted as a general `spi` config, which can be reused

kot149: Updated `Pmw3610Config` to use common `SpiConfig` structure (with `#[serde(flatten)]` to keep the TOML simple) in be036db.

rmk-config/src/lib.rs
```diff
@@ -711,12 +711,9 @@ pub struct JoystickConfig {
...
     /// Name of the sensor (used for variable naming)
     pub name: String,
-    /// SPI clock pin
-    pub sclk: String,
-    /// SPI bidirectional data pin (SDIO)
-    pub sdio: String,
-    /// Chip select pin
-    pub cs: String,
+    /// SPI pins
+    #[serde(flatten)]
```
HaoboGu: Actually I was imaging something be like:


```toml
[input_device.pmw3610]
name = ""
spi = { sclk = "", cs = "", sdio = "" }
motion = ""
# other config ..
```

Which one would you prefer? Flat or not flat the config?

kot149: Personally, I prefer flat configs. Inline tables (where line breaks aren't allowed) tend to have long lines, which reduces readability.
The trade-off is that the config structure becomes harder to understand.

HaoboGu: The readability is not an issue because you can also have this:

```toml
[input_device.pmw3610]
name = ""
spi.sclk = ""
spi.cs = ""
spi.sdio = ""
motion = ""
# other config ..
```

Actually I'll choose the config without `spi.` too, because I'm lazy :P. I just worry about that the field name might be confusing for other users without `spi.` or `spi = {}`.

kot149: Oh, I didn't know that. In that case I'd choose the non-flat config.

HaoboGu: Okay, let's go with the non-flat way!

HaoboGu:
Do you want to add testing before I merge this PR?

kot149:
I honestly don't really understand whether the test is necessary or not, but if it's not strictly required, it's fine to merge it as is.

HaoboGu (approved):
Nice work! Thanks very much!

Issue #231, opened 2025-01-03
Provide easy-to-use configuration for input devices in `keyboard.toml`


HaoboGu:

tib888:
See also https://github.com/HaoboGu/rmk/issues/268#issuecomment-2701991941 as an idea.

adong660:
@HaoboGu 

Hi. I notice that the [docs](https://rmk.rs/docs/configuration/input_device/encoder.html#toml-configuration) documented on how to configure actions of rotary encoders by writing some `encoder_map = ...`. But if I'm not mistaken, the current implementation sets all of them to `k!(No)` in the expanded function `get_default_encoder_map`. Are we going to implement it or seeking a more general way to configure encoders first?

HaoboGu:
> Hi. I notice that the [docs](https://rmk.rs/docs/configuration/input_device/encoder.html#toml-configuration) documented on how to configure actions of rotary encoders by writing some encoder_map = .... But if I'm not mistaken, the current implementation sets all of them to k!(No) in the expanded function get_default_encoder_map. Are we going to implement it or seeking a more general way to configure encoders first?


https://github.com/HaoboGu/rmk/pull/669 has been merged, so the encoder configuration should work.

Edit: oh that part of doc should not appear in v0.8 doc, it might be a mistake. Anyway, the new version will be released soon

adong660:
> > Hi. I notice that the [docs](https://rmk.rs/docs/configuration/input_device/encoder.html#toml-configuration) documented on how to configure actions of rotary encoders by writing some encoder_map = .... But if I'm not mistaken, the current implementation sets all of them to k!(No) in the expanded function get_default_encoder_map. Are we going to implement it or seeking a more general way to configure encoders first?
> 
> 
> https://github.com/HaoboGu/rmk/pull/669 has been merged, so the encoder configuration should work.
> 
> Edit: oh that part of doc should not appear in v0.8 doc, it might be a mistake. Anyway, the new version will be released soon

Ah, sorry for not seeing that.

adong660:
> > Hi. I notice that the [docs](https://rmk.rs/docs/configuration/input_device/encoder.html#toml-configuration) documented on how to configure actions of rotary encoders by writing some encoder_map = .... But if I'm not mistaken, the current implementation sets all of them to k!(No) in the expanded function get_default_encoder_map. Are we going to implement it or seeking a more general way to configure encoders first?
> 
[...]

PR #675, opened 2025-12-14
Move `encoder_map` config from `[layout]` to `[[layers]]`

adong660:
This PR moves `layout.encoder_map` to `layers.encoders`. See [this comment](https://github.com/HaoboGu/rmk/issues/231#issuecomment-3650224200).

It also adds alias resolution for encoder map.

If you were happy with it, I would update docs and examples accordingly :)


HaoboGu:
Sure, I'd be happy to do it, but before we removing `layout.keymap`, the `encoder_map` under `[layout]` should be kept.

adong660:
Yeah, I'll append `layout.encoder_map` as how we do to `layout.keymap`.

PR #678, opened 2025-12-17
feat: add report_hz option to PMW3610

kot149:
Added an option to limit the report rate of the PMW3610 input device, as its high report rate caused significant delays when connected via BLE.
Also optimized the timing for motion read and sending reports.

In the future, it might be better to extract this as a common process for pointing devices.

### DIFF
--- a/rmk/src/input_device/pmw3610.rs
+++ b/rmk/src/input_device/pmw3610.rs
@@ -7,7 +7,9 @@ use core::cell::RefCell;
 
 use embassy_time::{Duration, Instant, Timer};
 use embedded_hal::digital::{InputPin, OutputPin};
+use embedded_hal_async::digital::Wait;
 use embedded_hal_async::spi::SpiBus;
+use futures::future::pending;
 use usbd_hid::descriptor::MouseReport;
 
 use crate::channel::KEYBOARD_REPORT_CHANNEL;
@@ -157,12 +159,7 @@ pub struct MotionData {
 }
 
 /// PMW3610 driver using embedded-hal SPI traits
-pub struct Pmw3610<SPI, CS, MOTION>
-where
-    SPI: SpiBus,
-    CS: OutputPin,
-    MOTION: InputPin,
-{
+pub struct Pmw3610<SPI: SpiBus, CS: OutputPin, MOTION: InputPin + Wait> {
     spi: SPI,
     cs: CS,
     motion_gpio: Option<MOTION>,
@@ -170,12 +167,7 @@ where
     smart_flag: bool,
 }
 
-impl<SPI, CS, MOTION> Pmw3610<SPI, CS, MOTION>
-where
-    SPI: SpiBus,
-    CS: OutputPin,
-    MOTION: InputPin,
-{
+impl<SPI: SpiBus, CS: OutputPin, MOTION: InputPin + Wait> Pmw3610<SPI, CS, MOTION> {
     /// Create a new PMW3610 driver instance
     pub fn new(spi: SPI, cs: CS, motion_gpio: Option<MOTION>, config: Pmw3610Config) -> Self {
         Self {
@@ -445,12 +437,7 @@ enum InitState {
 /// PMW3610 as an InputDevice for RMK
 ///
 /// This device returns `Event::Joystick` events with relative X/Y movement.
-pub struct Pmw3610Device<SPI, CS, MOTION>
-where
-    SPI: SpiBus,
-    CS: OutputPin,
-    MOTION: InputPin,
-{
+pub struct Pmw3610Device<SPI: SpiBus, CS: OutputPin, MOTION: InputPin + Wait> {
     sensor: Pmw3610<SPI, CS, MOTION>,
     init_state: InitState,
     poll_interval: Duration,
@@ -461,12 +448,7 @@ where
     accumulated_y: i32,
 }
 
-impl<SPI, CS, MOTION> Pmw3610Device<SPI, CS, MOTION>
-where
-    SPI: SpiBus,
-    CS: OutputPin,
-    MOTION: InputPin,
-{
+impl<SPI: SpiBus, CS: OutputPin, MOTION: InputPin + Wait> Pmw3610Device<SPI, CS, MOTION> {
     const MAX_INIT_RETRIES: u8 = 3;
     const DEFAULT_POLL_INTERVAL_US: u64 = 500;
     const DEFAULT_REPORT_HZ: u16 = 125;
@@ -617,12 +599,7 @@ where
     }
 }
 
-impl<SPI, CS, MOTION> InputDevice for Pmw3610Device<SPI, CS, MOTION>
-where
-    SPI: SpiBus,
-    CS: OutputPin,
-    MOTION: InputPin,
-{
+impl<SPI: SpiBus, CS: OutputPin, MOTION: InputPin + Wait> InputDevice for Pmw3610Device<SPI, CS, MOTION> {
     async fn read_event(&mut self) -> Event {
         use embassy_futures::select::{Either, select};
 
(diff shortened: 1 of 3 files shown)


HaoboGu:
I noticed that there's a `poll_interval` in `Pmw3610Device`, is there any particular reason that `poll_interval` doesn't work?

And to control the reporting frequency, I think [`PollingController`](https://github.com/HaoboGu/rmk/blob/main/rmk/src/controller/mod.rs#L98) is what expected to be used

kot149:
`poll_interval` is the interval of sensor readings, so increasing it may cause the registers accumulating motion data internally in the PMW3610 to saturate. Additionally, `poll_interval` delay occurs before motion detection checks via `motion_gpio`, so increasing it would delay handling of the motion and introduce latency.
I believe it's safer to perform sensor reads at a high frequency and then accumulate them in software to reduce the reporting frequency.

I'll look into using the `PollingController`.

kot149:
I looked into `PollingController`. It seems incompatible with the `InputDevice` / `InputProcessor` pipeline but maybe it can be implemented like this:

```
Pmw3610Device (InputDevice) 
  → EVENT_CHANNEL 
  → Pmw3610Processor (InputProcessor) sends ControllerEvent
  → CONTROLLER_CHANNEL 
  → Pmw3610ReportController (PollingController) sends mouse report periodically
```

but not sure if this is a good idea. Please let me know if there is a better way to implement it.

HaoboGu:
What `PollingController` does is keep reading events from the channel and `process_event`, then periodically `update` the controller. I think you can keep reading the input event from `EVENT_CHANNEL`, process events in `process_event`, then send the report in periodically `update`? 

I think this is same as what you mentioned:

> I believe it's safer to perform sensor reads at a high frequency and then accumulate them in software to reduce the reporting frequency.

Did I overlook something?

kot149:
Thank you for clarifying. I seem to have misunderstood, I now understand how to implement it.

I just want to confirm one thing: it appears that the [`INTERVAL` in `PollingController`](https://github.com/HaoboGu/rmk/blob/main/rmk/src/controller/mod.rs#L100) is a fixed value and cannot be passed a value read from `keyboard.toml`. Should I modify `PollingController` to allow dynamic values to be passed? This change might also require modifying other existing controllers.

HaoboGu:
> Should I modify PollingController to allow dynamic values to be passed? This change might also require modifying other existing controllers.

In the current implementation it binds to the specific `Controller` implementation, yeah, making it configurable would be great! Maybe in another PR?

kot149:
OK, I'll modify that in another PR and then update the changes in this PR.

kot149:
Rebased onto #680 and rewrote the `Pmw3610Processor` rate-limit logic with `PollingController`.
Verified working on unibody / split central / peripheral. It's still a bit laggy on peripherals, but it's improved significantly.

Current implementation seems incompatible with the `battery_adc_pin` setting (when both battery and pmw3610 are set, either of them stops working). I tried increasing `controller_channel_pubs` / `controller_channel_subs` to see if that would fix it, but it didn't work. @HaoboGu Is there anything you can tell me about this issue?

kot149:
As [discussed on Discord](https://discord.com/channels/1166665039793639424/1220319940939092008/1453639293573660713), I abandoned the idea of using `PollingController` and instead implemented rate limiting within the `Pmw3610Device`.

I noticed that when the motion pin is configured, using `wait_for_low()` could eliminate polling entirely, but maybe it should be another PR.

HaoboGu:
> I noticed that when the motion pin is configured, using wait_for_low() could eliminate polling entirely, but maybe it should be another PR.

Does motion pin == low indicate there's mouse report? In my understanding it's prefect for detecting whether there's mouse move, because the system will be put in low-power state at `.await` point until there's moving detected. But when reading the values during moving, the polling way is still needed for controlling the report frequency, am I right?

If so, you can just do like what's in async matrix: wait the motion pin until it's low, then start scanning, otherwise just sleep(aka `wait_for_low()`)

HaoboGu:
I just checked the pmw3610 datasheet, yea you can do it. Reading out the data will clear the motion register, which means the motion pins is pulled up again. So you can make it very simple:

```rust
loop {
  // Wait until there's motion, otherwise sleep (automatically)
  motion_pin.wait_for_low().await;
  
  // Read out the data
  ..

  // After reading, the motion pin is pulled up again, so just continue the loop and wait for another motion
}
```

kot149:
>  the polling way is still needed for controlling the report frequency, am I right?

Yes, the reporting rate limit is still necessary, as it's still too frequent for BLE communication even when using the motion pin.
Also, I think it's better to keep polling implementation when the motion pin is not specified.

Should we refactor for using `wait_for_low()` in this PR? Or leave this PR as-is for now and handle it in a separate PR?

HaoboGu:
> Should we refactor for using wait_for_low() in this PR? Or leave this PR as-is for now and handle it in a separate PR?

I prefer to finish it in this PR

kot149:
Updated the code to use `wait_for_low()`, please give it a review.

rmk/src/input_device/pmw3610.rs
```diff
@@ -157,11 +160,14 @@ pub struct MotionData {
 }
 
 /// PMW3610 driver using embedded-hal SPI traits
-pub struct Pmw3610<SPI, CS, MOTION>
-where
+pub struct Pmw3610<
```
HaoboGu: I prefer the following format:

```rust
pub struct Pmw3610<
    SPI: SpiBus,
    CS: OutputPin,
    #[cfg(feature = "async_matrix")] MOTION: InputPin + Wait,
    #[cfg(not(feature = "async_matrix"))] MOTION: InputPin,
> {
  ...
}
```

rmk/src/input_device/pmw3610.rs
```diff
@@ -525,48 +641,72 @@ where
...
+            self.last_report = Instant::now();
+        }
 
-            if !self.sensor.motion_pending() {
-                continue;
-            }
+        loop {
+            #[cfg(feature = "async_matrix")]
+            let poll_wait = async {
+                if self.sensor.motion_gpio.is_some() {
```
HaoboGu: Can it be simplified like

```rust
if let Some(gpio) = self.sensor.motion_gpio.as_mut() {
    let _ = gpio.wait_for_low().await;
} else {
    Timer::after(
        self.poll_interval
            .checked_sub(self.last_poll.elapsed())
            .unwrap_or(Duration::MIN),
    )
    .await;
}
```

HaoboGu: I'm thinking about the feature + motion_gpio combination. From my understanding, if the `motion_gpio` is present, it **should** be used like `wait_for_low()`. In other words, if async matrix feature is needed, then the `motion_gpio` is required?

HaoboGu (commented):
Overall it looks very great! Only a few comments from me.

rmk/src/input_device/pmw3610.rs
kot149: If `motion_gpio` is configured, it's reasonable to assume `async_matrix` is enabled, but the reverse is not necessarily true. Insufficient microcontroller pins are a common issue, so even when `async_matrix` is enabled, the option to omit motion pins should remain available.

I think it's possible to use `wait_for_low()` when `motion_gpio` is provided, regardless of whether the `async_matrix` feature is enabled. Is that preferable?

HaoboGu: > I think it's possible to use wait_for_low() when motion_gpio is provided, regardless of whether the async_matrix feature is enabled. Is that preferable?

Yes, so the usage of async_matrix can be removed

kot149:
Okay, I've addressed the points raised in the review. Thanks you for reviewing 🙏

PR #681, opened 2025-12-21
Fix an issue that charging via usb port will close ble

L1cardo:
In the current implementation, as long as the USB cable is plugged in and voltage (VBUS) is detected, the enabled callback will trigger and emit the USB_ENABLED signal, which causes the keyboard to try to switch to USB mode and turn off Bluetooth regardless of whether it is connected to a computer or a charger.

My solution is to move the timing of triggering the USB_ENABLED signal from enabled (only powered on) to configured (the computer host has configured the device).
This way:
1. When connected to the charger: only enabled will be triggered and the USB_ENABLED signal will not be sent. The keyboard will maintain the Bluetooth connection or continue to broadcast.
2. When connected to the computer: the host will configure the device, trigger configured, send the USB_ENABLED signal, and switch the keyboard to USB mode.

---
目前的实现中，只要USB 线插入并检测到电压（VBUS），enabled 回调就会触发并发出 USB_ENABLED 信号，这导致键盘无论连接的是电脑还是充电器，都会尝试切换到 USB 模式并关闭蓝牙。

我的方案是将触发 USB_ENABLED 信号的时机从 enabled（仅通电）移动到 configured（电脑主机已配置设备）。
这样：
1.连接充电器时：只会触发 enabled，不会发出 USB_ENABLED信号，键盘将保持蓝牙连接或继续广播。
2. 连接电脑时：主机会配置设备，触发 configured，发出 USB_ENABLED信号，键盘场换到 USB 模式。

### DIFF
--- a/rmk/src/usb/mod.rs
+++ b/rmk/src/usb/mod.rs
@@ -8,12 +8,12 @@ use embassy_usb::{Builder, Handler};
 use ssmarshal::serialize;
 use static_cell::StaticCell;
 
+use crate::RawMutex;
 use crate::channel::KEYBOARD_REPORT_CHANNEL;
 use crate::config::DeviceConfig;
 use crate::descriptor::CompositeReportType;
 use crate::hid::{HidError, HidWriterTrait, Report, RunnableHidWriter};
-use crate::state::ConnectionState;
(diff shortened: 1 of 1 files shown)


HaoboGu:
Fixed by refactoring in #829

PR #685, opened 2025-12-26
Add config for input devices

### LINKED ISSUE #231: Provide easy-to-use configuration for input devices in `keyboard.toml`

HaoboGu:
Closes #231


Schievel1:
Can processors themselves emit events that get funneled back in at the start of the chain? 

What I mean is: if I have an event coming from one device, that I recognize in a processor and if I recognize it I want to do something in a device. E.g. I press a User-Keycode and that should set the cpi of a trackball sensor. My matrix device emits the keycode event, a processor sees it and wants to set the cpi. How would the processor talk to the sensor/ the device of the sensor? 

/ maybe every device should also get a Channel and proc_message method to make them controllable. Every processor should get a reference to the set of all available device channels.

rmk-macro/src/input_device/mod.rs
```diff
@@ -4,10 +4,14 @@ use pmw3610::expand_pmw3610_device;
 use proc_macro2::{Ident, TokenStream};
 use quote::quote;
 use rmk_config::{BoardConfig, CommunicationConfig, InputDeviceConfig, KeyboardTomlConfig, UniBodyConfig};
+use std::collections::HashMap;
```
Dreamy-Jy, line 7: How is HashMap being used here?

Given that we're targeting bare metal should data structures that require allocation be used?

The heapless crate has a HashMap Alternative [heapless::index_map::FnvIndexMap](https://docs.rs/heapless/latest/heapless/index_map/type.FnvIndexMap.html)

Dreamy-Jy: After further review I realize that this code is likely run at compile time, and there for able to use std functions with out issue. Confirm for me please?

HaoboGu: yes, this is the proc-macro crate, which runs at compile time

HaoboGu:
> Can processors themselves emit events that get funneled back in at the start of the chain?
> 
> What I mean is: if I have an event coming from one device, that I recognize in a processor and if I recognize it I want to do something in a device. E.g. I press a User-Keycode and that should set the cpi of a trackball sensor. My matrix device emits the keycode event, a processor sees it and wants to set the cpi. How would the processor talk to the sensor/ the device of the sensor?
> 
> / maybe every device should also get a Channel and proc_message method to make them controllable. Every processor should get a reference to the set of all available device channels.

Yes, you can. There are two ways:

1. The processor can emit another `Event` in it's `ProcessResult`, the processor can be chained. But now it's impossible to send the `Event` back to the start of the processor chain.
2. The processor can also publish a `ControllerEvent`. The `ControllerEvent` can be subscribed by any struct that is interested to this event(aka a `Controller`). For example, you can subscribe the `ControllerEvent` in your sensor and change the cpi.

Generally the event flow is 

```
InputDevice --Event--> InputProcessor --ControllerEvent--> Controller
```

In you case, the sensor can also be a controller.

HaoboGu:
Closed

PR #689, opened 2025-12-28
feat: PMW3360 and PMW3389 input devices

Schievel1:
Inplemented InputDevice for PMW3360 and PMW3389 optical mouse sensors. The driver is based on the existing PWM3610 driver in RMK and [Zephyr's implementation](https://github.com/zephyrproject-rtos/zephyr/blob/d31c6e95033fd6b3763389edba6a655245ae1328/drivers/input/input_pmw3610.c). 

Please note that this needs https://github.com/HaoboGu/rmk-template/pull/18 in order for new projects from rmkit to work. (At least for RP2040 and STM32). Otherwise users need to do `cargo add embassy-embedded-hal`. Users of RP2040 and STM32 also need to add embassy-embedded-hal to their keyboards `Cargo.toml` to use the PMW33xx sensor. 

**BREAKING**: PointingDevice and PointingProcessor replace Pmw3610Device and Pmw3610Processor. For the Pmw3610 the calls of ::new() for these stay the same, only the name changes. If using Rust to configure the keyboard change the calls, if using Toml nothing needs to be done.

## Example Usage
 
 ```toml
[[input_device.pmw33xx]]
name = "trackball0"
sensor_type = "PMW3360" # or PMW3389

spi.instance = "SPI0"
spi.sck = "PIN_18"
spi.mosi = "PIN_19"
spi.miso = "PIN_16"
spi.cs = "PIN_17"
# By default this driver uses blocking SPI. For a small performance gain,
# you can define the DMA channels. Then the driver is using SPI with DMA.
# Figure out the DMA channel using the Peripherals from the embassy hal crate
# (embassy-rp, embassy-stm32 etc.) similar to the PINs.
# On nrf52 the driver alsways uses DMA, the channels do not need to be specified.
# spi.tx_dma = "DMA_CH2" # omit for nrf52
# spi.rx_dma = "DMA_CH3" # omit for nrf52

motion = "PIN_20" # Optional. If omitted, the sensor is polled.

cpi = 1600
rot_trans_angle = -15
liftoff_dist = 8
invert_x = true
# invert_y = true
# swap_xy = true

```

## Todos

* [x]  Verify working with nRF52840
  * [x]  on unibody
  * [x]  on split central
  * [x]  on split peripheral

* [x]  Verify working with RP2040
  * [x]  on unibody
  * [x]  on split central
  * [x]  on split peripheral

* [x]  Verify working with STM32
  * [x]  on unibody
 
* [x]  Documentation
* [x]  Changelog
* [x]  Testing

* [ ] Test PMW3389 sensor

* [x] merge PR https://github.com/HaoboGu/rmk-template/pull/18

In the future, refactoring is needed to share implementation with JoyStick/ PWM3610 sensor or to handle SPI/I2C configs or generic processing like input transformation for pointing devices.

### DIFF
--- a/rmk/src/input_device/pmw33xx.rs
+++ b/rmk/src/input_device/pmw33xx.rs
@@ -803,15 +803,13 @@ mod tests {
         let config = Pmw33xxConfig::default();
 
         // Create PMW33xx device
-        let mut device = Pmw33xx::<_, _, _, Pmw3360Spec>::new(spi, cs, None::<PinMock>, config);
+        let mut device = Pmw33xx::<_, _, _, Pmw3360Spec>::new(0, spi, cs, None::<PinMock>, config);
 
         // Validate default configuration values
+        assert_eq!(device.id, 0);
         assert_eq!(device.config.res_cpi, 1600);
         assert_eq!(device.config.rot_trans_angle, 0);
         assert_eq!(device.config.liftoff_dist, 0x02);
-        assert_eq!(device.config.invert_x, false);
-        assert_eq!(device.config.invert_y, false);
-        assert_eq!(device.config.swap_xy, false);
         device.spi.done();
         device.cs.done();
     }
@@ -828,7 +826,7 @@ mod tests {
 
         let config = Pmw33xxConfig::default();
 
-        let mut device = Pmw33xx::<_, _, _, Pmw3360Spec>::new(spi, cs, None::<PinMock>, config);
+        let mut device = Pmw33xx::<_, _, _, Pmw3360Spec>::new(0, spi, cs, None::<PinMock>, config);
 
         let _ = block_on(device.set_resolution(2000));
         device.spi.done();
@@ -845,7 +843,7 @@ mod tests {
         let spi = SpiMock::new(&spi_expectations);
         let cs = PinMock::new(&[PinTrans::set(PinState::Low), PinTrans::set(PinState::High)]);
 
-        let mut device = Pmw33xx::<_, _, _, Pmw3360Spec>::new(spi, cs, None::<PinMock>, Pmw33xxConfig::default());
+        let mut device = Pmw33xx::<_, _, _, Pmw3360Spec>::new(0, spi, cs, None::<PinMock>, Pmw33xxConfig::default());
 
         block_on(device.set_resolution(100)).unwrap();
 
@@ -863,7 +861,7 @@ mod tests {
         let spi = SpiMock::new(&spi_expectations);
         let cs = PinMock::new(&[PinTrans::set(PinState::Low), PinTrans::set(PinState::High)]);
 
-        let mut device = Pmw33xx::<_, _, _, Pmw3360Spec>::new(spi, cs, None::<PinMock>, Pmw33xxConfig::default());
+        let mut device = Pmw33xx::<_, _, _, Pmw3360Spec>::new(0, spi, cs, None::<PinMock>, Pmw33xxConfig::default());
 
         block_on(device.set_resolution(12000)).unwrap();
 
@@ -878,8 +876,7 @@ mod tests {
         let spi = SpiMock::new(&expectations);
         let cs = PinMock::new(&[]);
 
-        let config = Pmw33xxConfig::default();
-        let mut device = Pmw33xx::<_, _, _, Pmw3360Spec>::new(spi, cs, None::<PinMock>, config);
+        let mut device = Pmw33xx::<_, _, _, Pmw3360Spec>::new(0, spi, cs, None::<PinMock>, Pmw33xxConfig::default());
 
         // 99 is illegal (below RES_MIN = 100 for PMW3360)
         let result = block_on(device.set_resolution(99));
@@ -895,8 +892,7 @@ mod tests {
         let spi = SpiMock::new(&expectations);
         let cs = PinMock::new(&[]);
 
-        let config = Pmw33xxConfig::default();
-        let mut device = Pmw33xx::<_, _, _, Pmw3360Spec>::new(spi, cs, None::<PinMock>, config);
+        let mut device = Pmw33xx::<_, _, _, Pmw3360Spec>::new(0, spi, cs, None::<PinMock>, Pmw33xxConfig::default());
 
         // 12001 is illegal (below RES_MAX = 12000 for PMW3360)
         let result = block_on(device.set_resolution(12001));
@@ -921,7 +917,7 @@ mod tests {
             PinTrans::set(PinState::High),
         ]);
 
-        let mut device = Pmw33xx::<_, _, _, Pmw3389Spec>::new(spi, cs, None::<PinMock>, Pmw33xxConfig::default());
+        let mut device = Pmw33xx::<_, _, _, Pmw3389Spec>::new(0, spi, cs, None::<PinMock>, Pmw33xxConfig::default());
 
         block_on(device.set_resolution(13000)).unwrap();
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by Schievel1** [comment]
This builds on NRF52xy, STM32 and RP2040 at least. I am lacking hardware right now to test out if the sensor really works on the NRF52 chips and STM32.

**Comment by Schievel1** [comment]
@whitelynx Maybe you can test the PMW3389 support when your mouse is ready. I am lacking the hardware and that sensor is too expensive for me to buy it just to test that out.

**Comment by Schievel1** [comment]
I think the PointingDevices should be able to set the resolution and rotational transform angle at runtime.

For this I would implement Controller for the PointingDevice,  and the new functions of PointingDevice would require an ID for the sensor. 
 
Then I would like to add to the keyboardevents(?) the event Keyboardevents::PointingSetResolution((u8, i16)).
The u8 being the ID, the i16 the desired resolution. 

Is this the best way to do it? I'm not sure if this should be in keyboardevents or if pointing devices should get their own events enum.

**Comment by HaoboGu** [comment]
> I think the PointingDevices should be able to set the resolution and rotational transform angle at runtime.
> 
> For this I would implement Controller for the PointingDevice, and the new functions of PointingDevice would require an ID for the sensor.
> 
> Then I would like to add to the keyboardevents(?) the event Keyboardevents::PointingSetResolution((u8, i16)). The u8 being the ID, the i16 the desired resolution.
> 
> Is this the best way to do it? I'm not sure if this should be in keyboardevents or if pointing devices should get their own events enum.

Yeah, the simplest way is to add a `PointingSetResolution(u8, i16)` variant to `ControllerEvent`.

I think it still needs some design. The main issues of the current event system are

1. Now once a variant of `ControllerEvent` is added, it exists even no subscriber uses it at all, ideally in the enum only the variant which will be consumed exists
2. If users don't want to modify RMK source code, now there's no way to make runtime change of a device/processor/controller
3. Now there's no id for controllers, and the published events have no id too

**Comment by Schievel1** [comment]
I solved the ID thing, at least for Pointing devices, and with your other PR we solve the thing with custom processors. 
For the first point I have no idea how you could solve this. Other than Feature mess. But is it really a big problem not all variants of the enum are used?

**Comment by HaoboGu** [comment]
It isn't a big problem, it's just not very elegant -- having unnecessary variants wastes some RAM

I'm thinking about introducing some feature gates, but if we want users to add their own event, the current event enum should be revised maybe?

**Comment by Schievel1** [comment]
I think this is good now. I merged the change from main into the new PointingDriver-System. I tested it again with RP2040, NRF52 and STM32.
Maybe PointingDriver-Trait is a bit too restrictiv now and demand too much details. But we can refactor when we have a sensor that implements too many dummy functions from the trait and actually does nothing of those.

**Comment by HaoboGu** [comment]
Nice work! I'll have time reviewing it tomorrow

**Comment by kot149** [comment]
Good job. Just to be sure, as an integration test, I confirmed that the PMW3610 works without problems via keyboard.toml :)

File: rmk/src/input_device/pointing.rs [hunk 1]
```diff
@@ -0,0 +1,805 @@
...
+
+    async fn init(&mut self) -> Result<(), PointingDriverError>;
+    async fn read_motion(&mut self) -> Result<MotionData, PointingDriverError>;
+    async fn set_resolution(&mut self, cpi: u16) -> Result<(), PointingDriverError>;
+    async fn set_rot_trans_angle(&mut self, angle: i8) -> Result<(), PointingDriverError>;
+    async fn set_liftoff_dist(&mut self, dist: u8) -> Result<(), PointingDriverError>;
+    async fn force_awake(&mut self, enable: bool) -> Result<(), PointingDriverError>;
+    async fn set_invert_x(&mut self, onoff: bool) -> Result<(), PointingDriverError>;
+    async fn set_invert_y(&mut self, onoff: bool) -> Result<(), PointingDriverError>;
+    async fn swap_xy(&mut self, onoff: bool) -> Result<(), PointingDriverError>;
```
kot149: It might be better to unify whether or not to use the `set_` prefix.

rmk/src/input_device/pmw3610.rs
```diff
@@ -415,47 +392,126 @@ impl<SPI: SpiBus, CS: OutputPin, MOTION: InputPin + Wait> Pmw3610<SPI, CS, MOTIO
...
 
-/// Initialization state for the device
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
-enum InitState {
-    Pending,
-    Initializing(u8),
-    Ready,
-    Failed,
-}
+    async fn set_invert_x(&mut self, onoff: bool) -> Result<(), PointingDriverError> {
```
kot149: Similar to `set_resolution()`, we should enable the SPI clock, switch to SPI page 1, and restore the settings when finished.
The same applies to `set_invert_y()` and `swap_xy()`.

Schievel1: I don't have the sensor to test out if what I did really works. Can you propose a change and test it with your sensor as well?

Also, have you changed that in your implementation before? I thought I copied yours 1:1 for pmw3610. If what we have here works I would rather merge this PR and do the changes/ improvements for the pmw3610 in a different PR. 
This PR is about adding pmw3360/89 and unifying pointing. And it's already bulky enough.

kot149: These methods (except `set_resolution()`) were not present in the original PMW3610 driver and were introduced in this PR. They aren't actually called from anywhere so I think you can leave them unimplemented for now and I'll address it in another PR.

Schievel1: You're right, I copied the code from the initialization. Good catch. 
Let's do it this way then: I'll comment them out and drop a log message like I do with the other unsupported trait functions in the drivers. Later on you can add them in a different PR, alright?

kot149: Yes, I think that's fine.

Schievel1:
It just came to me that the pointing device driver trait should rather implement the default functions for the setters that say "that's not implemented" and the actual drivers should only implement if they actually do something in those functions

rmk/src/input_device/pmw33xx.rs
```diff
@@ -0,0 +1,1564 @@
...
+            .write(&[register.value() & 0x7f])
+            .await
+            .map_err(|_| Pmw33xxError::Spi)?;
+
+        Timer::after(Duration::from_micros(T_SRAD_MOTBR_US)).await;
+
+        self.spi.read(data).await.map_err(|_| Pmw33xxError::Spi)?;
+
+        Self::short_delay();
+        let _ = self.cs.set_high();
```
HaoboGu, line 464: Why `short_delay` here? I'm kind of confusing about the mixed usage of `short_delay` and `Timer::after`

HaoboGu:
What do you think about tools like tools like [device-driver](https://github.com/diondokter/device-driver)? Will it simplify the driver implementation?

rmk/src/input_device/pointing.rs
```diff
@@ -0,0 +1,826 @@
...
+    /// Invalid rotational transform angle
+    InvalidRotTransAngle,
+}
+
+pub trait PointingDriver {
+    type MOTION: InputPin + Wait;
+
+    async fn init(&mut self) -> Result<(), PointingDriverError>;
+    async fn read_motion(&mut self) -> Result<MotionData, PointingDriverError>;
+    async fn set_resolution(&mut self, _cpi: u16) -> Result<(), PointingDriverError> {
```
HaoboGu: For the optional setting, I think a separate trait would help:

```rust
pub trait PointingDriver { ... }

pub trait InvertablePointing: PointingDriver {
async fn set_invert_x();
async fn set_invert_y();

// ... other similar traits
```

A driver with invertable axis should impl both PointingDriver and InvertablePointing.

rmk/src/input_device/pmw33xx.rs
Schievel1: Maybe we can ask @kot149 because I copied this mostly from the 3610 driver

rmk/src/input_device/pointing.rs
Schievel1: I was thinking about making this optional, as not all drivers have the ability. And I asked myself if the driver is really the best place to implement the reversing of the axis in a software way if a  sensor doesn't have the ability. 

Then again: if you still want to have the ability to invert an axis  you would have to add it in PointingDevice or PointingProcessor. 

Maybe we should do both?

rmk/src/input_device/pmw33xx.rs
kot149: `short_delay()` has less overhead and could be more accurate than `Timer::after()`, especially for very small delays, but not sure if such minor errors would actually be a problem. For readability, it might be acceptable to unify the code using `Timer::after()`.

rmk/src/input_device/pointing.rs
HaoboGu: For software invert/swap, I think a separate processor might be better

rmk/src/input_device/pmw33xx.rs
HaoboGu: )z`short_delay` is actually a busy loop, I prefer `Timer::after()` which frees the CPU and is more power efficiency(the actually difference might be subtle tho)

rmk/src/input_device/pointing.rs
Schievel1: Hm, I will add it into PointingProcessor, because I think this is where it should be. But since a PointingProcessor processes every pointing input, that would mean you invert it for everything, not just one sensor. But we can leave it for now at that and wait until the processor chain got it's update.

Schievel1: ugh, I added it to PointingPrecossor as a static setting in its config. 
Who needs to change that at runtime anyway? So no need for controller sub capability on this. 
If someone does something this obscure, they can just implement their own processor that subscribes to those event.

rmk/src/input_device/pmw3610.rs
Schievel1: I removed those messages again. So I think that is done from your side.

rmk/src/input_device/pmw33xx.rs
Schievel1: I will change everything to Timer::after and give it a try if the sensor complains. Maybe you do the same with the PMW3610 @kot149

Schievel1:
I added the new event system events. Please check if I'm doing this right. Because it's all a bit confusing to me. 
PointingProcessor has a config now to set the invert x/y and sway x/y settings. 
PointingDevice now does only what it dd before: Read the sensor end emit joystick events.
I moved the whole event stuff into the drivers, the Driver trait is now small again with only a few functions that PointingDevice needs. 
Drivers only listen for events that they implement a setting for. 

I will fix the tests once this is a state where it isn't changing that fast anymore.

Schievel1:
I don't really know where sensor.event_loop() should be called. I feel like this should be in the main(), together with the other controllers. But this won't work because of own&move, the sensor is in PointingDevice. Maybe with Refcell<Arc<Mutex<Whatnot<Cell<>>>> but that's ugly as hell.
Calling it in the read_event()'s select does not work, because the other futures in select finish first and the event_loop() only very seldom can do it's thing. (That is with polling though) Also it doesn't feel like this should be called in the Inputdevice::read_event().

HaoboGu:
> I don't really know where sensor.event_loop() should be called. I feel like this should be in the main(), together with the other controllers. But this won't work because of own&move, the sensor is in PointingDevice. Maybe with Refcell<Arc<Mutex<Whatnot<Cell<>>>> but that's ugly as hell.
> Calling it in the read_event()'s select does not work, because the other futures in select finish first and the event_loop() only very seldom can do it's thing. (That is with polling though) Also it doesn't feel like this should be called in the Inputdevice::read_event().

Since you have `PointingDevice<S: PointingDriver>`, where `PointingDriver` is also a `Controller`, the main loop should be manually implemented for `PointingDevice`, which combines both `read_event()`(InputDevice) and `event_loop()`(Controller).

In main function, something like `PointintDevice::run()` is called to process both.

Schievel1:
> > I don't really know where sensor.event_loop() should be called. I feel like this should be in the main(), together with the other controllers. But this won't work because of own&move, the sensor is in PointingDevice. Maybe with Refcell<Arc<Mutex<Whatnot<Cell<>>>> but that's ugly as hell.
> > Calling it in the read_event()'s select does not work, because the other futures in select finish first and the event_loop() only very seldom can do it's thing. (That is with polling though) Also it doesn't feel like this should be called in the Inputdevice::read_event().
> 
> Since you have `PointingDevice<S: PointingDriver>`, where `PointingDriver` is also a `Controller`, the main loop should be manually implemented for `PointingDevice`, which combines both `read_event()`(InputDevice) and `event_loop()`(Controller).
> 
> In main function, something like `PointintDevice::run()` is called to process both.

In the main loop I need to call read_event(&mut self), but at the same time I need to call event_loop(&mut self). Doing this with sensor at the same time is impossible. In the end I got it going with a separate controller_receiver structs and a small channel between this struct and the pointing driver. But it was to ugly to use. 

Therefore I removed the whole controller thing for pointing devices for now. Since the event system is a moving target anyway (#717) and this PR is already by broad enough, I think it's best to get this merged first and add the controller event system for pointing devices at a later stage.

You can review the PR again now. I will fix the tests in the meantime.

HaoboGu:
Yes, because the `event/polling_loop` in `Controller` never returns, so you cannot use it with `read_event`. In #717, they are automatically processed by the [proc-macro code](https://github.com/HaoboGu/rmk/pull/717/changes#diff-fb5e5d6f84742333d10526a96e69d34cde112ca4f7604d9cbd8e0019fe726343R213-R229)(which is very complex). I can add the controller part back after #717 is merged.

rmk-macro/src/bind_interrupt.rs
```diff
@@ -138,15 +138,59 @@ pub(crate) fn bind_interrupt_default(keyboard_config: &KeyboardTomlConfig, item_
...
+            };
+
+            // Generate SPI interrupts for each sensor
+            let mut pmw33xx_spi_interrupts = Vec::new();
+
+            for sensor in &pmw33xx_config {
+                let instance_ident = format_ident!("{}", &sensor.spi.instance);
+
+                pmw33xx_spi_interrupts.push(quote! {
+                    #instance_ident => spim::InterruptHandler<peripherals::#instance_ident>;
```
HaoboGu: Use the full path like `::embassy_nrf::spim::InterruptHandler<::embassy_nrf::peripherals::#instance_ident>;` here, so you don't need `spim_import`

rmk-macro/src/input_device/pmw33xx.rs
```diff
@@ -0,0 +1,279 @@
...
+    pmw33xx_config: Vec<Pmw33xxConfig>,
+    chip: &ChipModel,
+) -> (Vec<Initializer>, Vec<Initializer>) {
+    if pmw33xx_config.is_empty() {
+        return (Vec::new(), Vec::new());
+    }
+
+    // PMW33xx is only supported on nRF52, STM32 and RP2040
+    match chip.series {
+        // NOTE:; Nonblocking SPI with DMA is still marked unstable in esp_hal.
```
HaoboGu:
```suggestion
        // NOTE: Nonblocking SPI with DMA is still marked unstable in esp_hal.
```

HaoboGu:
Looks good, I'll fix the unittest later.

Many thanks for the great work!

PR #690, opened 2025-12-29
Update and lock RMK version for py32/sf32 example

decaday:
As discussed in #552, I have locked the RMK dependency for `py32f07x` and `sf32lb52x` to decouple them from `main`. This handles the different HAL release cycles better and reduces maintenance overhead.

Since the build is now fixed and stable, I also re-enabled `py32f07x` in the CI script (`scripts/check_all.sh`).

I've added a note in the READMEs to clarify this update policy—let me know if the wording looks good to you.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -52,7 +52,7 @@ jobs:
       - name: Install cargo-make and flip-link
         run: cargo binstall cargo-make flip-link
       - name: Install target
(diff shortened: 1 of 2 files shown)


decaday:
I added sf32/py32 examples to the CI. I'm not sure if adding a new job in `.github/build.yml` is the best way to handle this.

PR #694, opened 2025-12-30
Rewrite `KeyCode`

HaoboGu:


Schievel1:
Now that I see that, do we have the ability to define custom keycodes, and then react to them in our custom processors form #685 ?

HaoboGu:
> Now that I see that, do we have the ability to define custom keycodes, and then react to them in our custom processors form https://github.com/HaoboGu/rmk/pull/685 ?

I hope so! I think the custom keycode can be added at least in this PR

HaoboGu:
> I hope so! I think the custom keycode can be added at least in this PR

`Action::User(u8)` can be used for this case

PR #705, opened 2026-01-05
Fix invalid button state for mouse

### LINKED ISSUE #704: Problems with dragging and dropping
keyboard.toml
```toml
[[split.central.input_device.pmw3610]]
name = "trackball0"
spi.instance = "bitbang0"
spi.sck = "P1_13"
spi.mosi = "P0_10"
spi.miso = "P0_10"
spi.cs = "P0_20"
report_hz = 90
motion = "P1_11" # Optional. If omitted, the sensor is polled.
force_awake = false
smart_mode = true
cpi = 600
invert_x = true
# invert_y = true
# swap_xy = true
```

I put the mouse buttons `Mouse 1`, `Mouse 2`, `Mouse 3` on one of the layers
When I try to drag something, the object grabs, but after I start moving the cursor, the object is released
I checked in the vial Matrix tester, `Mouse 1` after actually releasing the object, still glows as pressed.
It doesn't depend on layers. It repeats on the base layer, the second layer, and so on.

HaoboGu:
Closes #704

### DIFF
--- a/rmk/src/input_device/joystick.rs
+++ b/rmk/src/input_device/joystick.rs
@@ -61,8 +61,9 @@ impl<'a, const ROW: usize, const COL: usize, const NUM_LAYER: usize, const NUM_E
 
         debug!("JoystickProcessor::generate_report: report = {:?}", report);
         // map to mouse
+        let buttons = self.keymap.borrow().mouse_buttons;
         let mouse_report = MouseReport {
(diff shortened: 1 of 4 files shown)


HaoboGu:
This is not a final solution tho, how to share state and keymap info across the project needs some design and refactoring

PR #720, opened 2026-02-02
feat: add one-shot sticky modifiers

braindefender:
Added One-Shot Sticky Modifiers functionality
Moved OSM/OSL logic into `oneshot.rs` file

Added two boolean config variables:
- `activate_on_keypress`: Should modifiers be active from keypress (sticky modifiers)
- `send_on_second_press`: Should the second keypress send modifiers and unstick them

Updated `behavior.md` docs to correspond with the new config structure:
```toml
[behavior.one_shot_modifiers]
  timeout = "1s"
  activate_on_keypress = false
  send_on_second_press = false

[behavior.one_shot_layers]
  timeout = "1s"
```

### DIFF
--- a/docs/docs/main/docs/configuration/appendix.md
+++ b/docs/docs/main/docs/configuration/appendix.md
@@ -109,9 +109,21 @@ MouseWheelLeft   MouseDown  MouseWheelRight  MouseWheelDown
 # Behavior configuration, if you don't want to customize anything, just ignore this section
 [behavior]
 # Tri Layer configuration
-tri_layer = { upper = 1, lower = 2, adjust = 3 }
-# One Shot configuration
-one_shot = { timeout = "1s" }
+tri_layer = {
+  upper = 1,
+  lower = 2,
+  adjust = 3,
+}
+
+# OneShot configuration
+one_shot = {
+  timeout = "1s"
+}
+
+# One Shot Modifiers configuration
+one_shot_modifiers = {
+  activate_on_keypress = false,
+}
 
 [behavior.morse]
 # default profile for morse, tap dance and tap-hold keys:
--- a/docs/docs/main/docs/configuration/behavior.md
+++ b/docs/docs/main/docs/configuration/behavior.md
@@ -4,8 +4,17 @@ The `[behavior]` section contains configuration for how different keyboard actio
 
 ```toml
 [behavior]
-tri_layer = { upper = 1, lower = 2, adjust = 3 }
-one_shot = { timeout = "1s" }
+tri_layer = {
+  upper = 1,
+  lower = 2,
+  adjust = 3,
+}
+one_shot = {
+  timeout = "1s",
+}
+one_shot_modifiers = {
+  activate_on_keypress = false,
+}
 ```
 
 ## Tri Layer
@@ -25,13 +34,34 @@ In this example, when both layers 1 (`upper`) and 2 (`lower`) are active, layer 
 
 Note that `"#layer_name"` could also be used in place of layer numbers.
 
-## One Shot
+## One-Shot
 
-The `one_shot` sub-table configures one-shot modifiers or one-shot layers (OSM/OSL). Use `timeout` to specify how long the modifier/layer remains active. The value is a string suffixed with `s` or `ms` (default: `1s`).
+The `one_shot` sub-table contains common one-shot configuration (for both OSM and OSL)
 
+Currently, there are only `timeout` field that specifies how long the one-shot modifier/layer remains active.
+When no key is pressed within this time, the one-shot modifier/layer will be canceled.
+`timeout` value is a string suffixed with `s` or `ms` (default: `1s`).
+
+## One-Shot Modifiers
+
+The `one_shot_modifiers` sub-table configures one-shot modifiers (OSM).
+
+By default, one-shot modifiers do not activate on keypress and will be sent only when other key is pressed.
+You can change this behavior by setting `activate_on_keypress` to `true`.
+This behavior is also known as One-Shot Sticky Modifiers (OSSM).
+
+If you press One-Shot Modifier again, it will be sent as a normal modifier key press and, therefore, released.
+
+Default values:
 ```toml
-[behavior.one_shot]
-timeout = "5s"
+[behavior.one_shot_modifiers]
+activate_on_keypress = false
+```
+
+OSSM example:
+```toml
+[behavior.one_shot_modifiers]
+activate_on_keypress = true
 ```
 
 ## Combo
--- a/rmk-config/src/lib.rs
+++ b/rmk-config/src/lib.rs
@@ -547,6 +547,7 @@ pub struct KeyInfo {
 pub(crate) struct BehaviorConfig {
     pub tri_layer: Option<TriLayerConfig>,
     pub one_shot: Option<OneShotConfig>,
+    pub one_shot_modifiers: Option<OneShotModifiersConfig>,
     pub combo: Option<CombosConfig>,
     #[serde(alias = "macro")]
     pub macros: Option<MacrosConfig>,
@@ -583,11 +584,18 @@ pub(crate) struct TriLayerConfig {
     pub adjust: u8,
 }
 
-/// Configurations for one shot
+/// Configurations for oneshot modifiers/layers
 #[derive(Clone, Debug, Deserialize)]
 #[serde(deny_unknown_fields)]
 pub(crate) struct OneShotConfig {
     pub timeout: Option<DurationMillis>,
+}
+
+/// Configurations for oneshot modifiers
+#[derive(Clone, Debug, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct OneShotModifiersConfig {
+    pub activate_on_keypress: Option<bool>,
 }
 
 /// Configurations for combos
--- a/rmk-macro/src/codegen/behavior.rs
+++ b/rmk-macro/src/codegen/behavior.rs
@@ -5,7 +5,7 @@ use std::collections::HashMap;
 use quote::quote;
 use rmk_config::resolved::Behavior;
 use rmk_config::resolved::behavior::{
-    Combos, Forks, MacroOperation, Macros, Morse, MorseActionPair, MorseKey, MorseProfile,
+    Combos, Forks, MacroOperation, Macros, Morse, MorseActionPair, MorseKey, MorseProfile, OneShot,
 };
 
 use super::action_parser::{expand_profile, expand_profile_name, get_key_with_alias, parse_key};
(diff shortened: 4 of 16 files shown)


rmk-config/src/lib.rs
```diff
@@ -656,10 +657,19 @@ pub struct TriLayerConfig {
...
 
-/// Configurations for one shot
+/// Configurations for one shot modifiers
 #[derive(Clone, Debug, Deserialize)]
 #[serde(deny_unknown_fields)]
-pub struct OneShotConfig {
+pub struct OneShotModifiersConfig {
+    pub timeout: Option<DurationMillis>,
+    pub activate_on_keypress: Option<bool>,
+    pub send_on_sedond_press: Option<bool>,
```
HaoboGu: send_on_second_press?

rmk-macro/src/behavior.rs
```diff
@@ -22,19 +22,41 @@ fn expand_tri_layer(tri_layer: &Option<TriLayerConfig>) -> proc_macro2::TokenStr
...
                 Some(t) => t.0,
                 None => return default,
             };
 
             let timeout = quote! {::embassy_time::Duration::from_millis(#millis)};
 
             quote! {
-                ::rmk::config::OneShotConfig {
+                ::rmk::config::OneShotModifiersConfig {
+                    timeout: #timeout,
```
HaoboGu, line 33: `activate_on_keypress` and `send_on_second_press` are omitted here, is it correct?

rmk-config/src/lib.rs
braindefender: yeah, fixed the typo

rmk-macro/src/behavior.rs
braindefender: I didn't know how to properly expand them. With the help of LLM, it should be fixed now (:

braindefender:
Updated the docs, squashed all WIP commits.
Now, I think, it's ready for review.
@HaoboGu

docs/docs/main/docs/configuration/behavior.md
```diff
@@ -4,8 +4,19 @@ The `[behavior]` section contains configuration for how different keyboard actio
...
+  lower = 2,
+  adjust = 3,
+}
+one_shot = {
+  timeout = "1s",
+}
+one_shot_modifiers = {
+  activate_on_keypress = false,
+  send_on_second_press = false,
+  release_modifier_keys = [],
```
HaoboGu: `send_on_second_press` indicates that the current modifier is released by pressing the same key, so it's kinda overlapped with `release_modifier_keys`?

I suggest to remove `release_modifier_keys` and make `send_on_second_press` defaults to true.

braindefender: I think that ability to set explicit keys to cancel/release OSSM is better, because with that you can set one key to release all modifiers. Without that, you'll need to press every OSM to cancel and you can make mistake.

`send_on_second_press` is primarily for the case where you want to trigger the key tap, for example, bring Windows Start Menu or something like that in other OS.

while `release_modifier_keys` are for explicit release of all OSM modifiers.

HaoboGu: I don't really understand, IIRC in current implementation, all OSM modifiers are cancelled after a non-letter keys is pressed, I thought it was enough for this case?

braindefender: OSM is canceled/released when:
- other non-OSM key is pressed (that will trigger a shortcut ModifierCombination+Key)
- timeout has passed (ModifierCombination will release alone)

If the timeout is set to a very big number (like minutes), then the second option doesn't suit us.
But we want to release modifiers without triggering unnecessary shortcut.
Therefore, we use third option: release_modifier_keys which release mods without triggering any shortcuts.

Why timeout could be set to very big number: in some apps, there are alternative scroll mode (fast scroll, or side scroll). If you want to trigger that behavior, but you don't want to hold the mod for the whole period, you use OSSM with a big timer and release that modifier when it is not needed anymore.

HaoboGu: > Why timeout could be set to very big number: in some apps, there are alternative scroll mode (fast scroll, or side scroll). If you want to trigger that behavior, but you don't want to hold the mod for the whole period, you use OSSM with a big timer and release that modifier when it is not needed anymore.

Sorry, maybe I didn't make it clear. I think the "cancelling" behavior can be achieved by tapping the same OSSM key again. My expected behavior is:

1. Tap OSSM key for the first time: activate OSM immediately
2. Tap the same OSSM key for the second time within the timeout time: cancel current OSSM
3. Tap other keys within the timeout time: modifier combination + key

With this behavior, you can open Windows Start menu by pressing OSM twice, and can also use OSSM in the scroll mode.

That simplifies the configuration too and saves lots of ram introduced by release_modifier_keys.

braindefender: Okay, fixed it.

braindefender: And synced with the upstream

braindefender:
@HaoboGu Some tests failed, but I can't figure out why. Is `main` branch ok?

rmk/src/keyboard/oneshot.rs
```diff
@@ -1 +1,194 @@
...
+                        OneShotState::Single(cur_modifiers | new_modifiers)
+                    }
+                }
+                OneShotState::Held(cur_modifiers) => OneShotState::Held(cur_modifiers | new_modifiers),
+            };
+
+            self.update_osl(event);
+
+            // Send report for updated osm_state modifiers
+            if was_active && send_on_second_press {
```
HaoboGu: If I understand it correctly, if `send_on_second_press` is true, when the key is pressed for the second time, `self.send_keyboard_report_with_resolved_modifiers(true).await` is called twice(L53 and L71)?

HaoboGu:
I've fixed the test

docs/docs/main/docs/configuration/appendix.md
```diff
@@ -109,9 +109,22 @@ MouseWheelLeft   MouseDown  MouseWheelRight  MouseWheelDown
...
+}
+
+# OneShot configuration
+one_shot = {
+  timeout = "1s"
+}
+
+# One Shot Modifiers configuration
+one_shot_modifiers = {
+  activate_on_keypress = false
```
HaoboGu:
```suggestion
  activate_on_keypress = false,
```

rmk/tests/keyboard_one_shot_test.rs
```diff
@@ -102,75 +184,422 @@ mod one_shot_test {
...
                     [0, [0, 0, 0, 0, 0, 0]], // All released
                 ]
             };
         }
 
+        /// OSM Test Case 11
+        ///
+        /// Config:
+        /// - timeout: 100ms
+        /// - activate_on_keypress: true
```
HaoboGu: In the test settings, it's `false`

HaoboGu:
When `send_on_second_press` is enabled, pressing OSM(Shift), releasing it, then pressing and holding OSM(Shift) again while typing A sends an unshifted A, this feels like incorrect.

If the second tap of the same OSM cancels(releases) itself, when `activate_on_keypress` is enabled, the double-tap already works without `send_on_second_press`, right?

braindefender:
> When `send_on_second_press` is enabled, pressing OSM(Shift), releasing it, then pressing and holding OSM(Shift) again while typing A sends an unshifted A, this feels like incorrect.

Why? `send_on_second_press` has fired modifier release on **second press** (not on press+release) as stated. So it won't already be active while you press A. 

> If the second tap of the same OSM cancels(releases) itself, when `activate_on_keypress` is enabled, the double-tap already works without `send_on_second_press`, right?

No, this is not quite right. When `send_on_second_press` is `false`, any consecutive taps on same OSM should do nothing. `activate_on_keypress` just holds it from the initial keypress.

rmk/src/keyboard/oneshot.rs
braindefender: Yep, that's why I asked you if identical reports are okay on Discord.

CleanShot 2026-02-22 at 23 38 23@2x

So.. seems like it won't be a problem?

HaoboGu:
> No, this is not quite right. When send_on_second_press is false, any consecutive taps on same OSM should do nothing. activate_on_keypress just holds it from the initial keypress.

Fair enough.

The only thing I'm thinking about now is:

Is `send_on_second_press` really needed except the "double-tap" case. For the double-tap case, I prefer to make 

"Tap the same OSSM key for the second time within the timeout time: cancel current OSSM"

default behavior when `activate_on_keypress = true`. In this case, only one option(`activate_on_keypress`) is needed, and the second send in [this comment](https://github.com/HaoboGu/rmk/pull/720#discussion_r2810903921) can be eliminated.

But if there are other cases which need `send_on_second_press`, the current design is fine.

braindefender:
> default behavior when `activate_on_keypress = true`. In this case, only one option(`activate_on_keypress`) is needed, and the second send in [this comment](https://github.com/HaoboGu/rmk/pull/720#discussion_r2810903921) can be eliminated.
> 
> But if there are other cases which need `send_on_second_press`, the current design is fine.

On the one hand, we need `send_on_second_press` for compatibility with the default QMK/ZMK behavior. Cause the user can just misclick when trying to combine mods and `send_on_second_press` will be unwanted in that case.

On the other hand, making this behavior default will make this behavior "smarter". But I don't think that the code simplicity wins against better configurability. So, it's up to you to decide which way it should be `(:`

HaoboGu:
> On the one hand, we need send_on_second_press for compatibility with the default QMK/ZMK behavior. Cause the user can just misclick when trying to combine mods and send_on_second_press will be unwanted in that case.
> 
> On the other hand, making this behavior default will make this behavior "smarter". But I don't think that the code simplicity wins against better configurability. So, it's up to you to decide which way it should be (:

I’d like to keep the configuration simple. For now, the `activate_on_keypress` option should be sufficient for most users. If we get more requests later, we can always add `send_on_second_press`. Let's remove it for the time being. Thank you a lot for the hard work, I really appreciate it

braindefender:
@HaoboGu 

Synced with the upstream and removed `send_on_second_press` making that behavior true by default. 

Please, check that everything is okay.

rmk-macro/src/codegen/behavior.rs
```diff
@@ -3,9 +3,10 @@
 use std::collections::HashMap;
 
 use quote::quote;
+use rmk_config::OneShotModifiersConfig;
```
HaoboGu:
```suggestion

```

braindefender:
> It looks nice now, only a minor import fix here

Fixed. Ready to merge?

HaoboGu:
> > It looks nice now, only a minor import fix here
> 
> Fixed. Ready to merge?

Yes! Thanks a lot for your effort!

PR #724, opened 2026-02-09
feat: add Tabber action

braindefender:
Adds Tabber action that provides Alt+Tab-like tab/window switching behavior.

When you press a Tabber key:

1. **First press**: Sends the specified modifier(s) + Tab, then holds the modifier(s)
2. **Subsequent presses**: Sends only Tab (while modifier is held)
3. **Layer switch**: Releases the held modifier(s)

Possible improvements:
- Make the "Tab" key configurable, leading to "HoldTapper" behavior (ex. HoldTapper(ModifierCombination, Key)).

Current limitations:
- Tabber cannot be used in the base layer. This is a safety measure to prevent held modifiers from being stuck. 
> I don't see the how this could be used in the base layer


braindefender:
closing in favor of #939

Issue #703, opened 2026-01-04
Scroll and other mode for pointing devices


Jaraxzus:
Now there are a few modes for working with the trackball
I only use `scroll` mode, but there is also a `high-precision` mode, etc.
I searched through the source code and pool requests, and I liked the idea of an abstract `PointingProcessor` from [PMW3360 and PMW3389 input devices](https://github.com/HaoboGu/rmk/pull/689)
It seemed to me a good idea to make such modes options of the contact Pointing device, and not a specific implementation, since the algorithms are most likely universal for all devices, at least software implementations
I made one for myself 
```rust
struct ScrollConfig {
    delta_per_tick: u8,
}

impl Default for ScrollConfig {
    fn default() -> Self {
        Self { delta_per_tick: 25 }
    }
}

struct MotionAccumulator {
    acc_x: i32,
    acc_y: i32,
}

impl MotionAccumulator {
    const fn new() -> Self {
        Self { acc_x: 0, acc_y: 0 }
    }

    fn process(&mut self, dx: i16, dy: i16, divisor: u8) -> (i16, i16) {
        self.acc_x += dx as i32;
        self.acc_y += dy as i32;

        let out_x = (self.acc_x / divisor as i32) as i16;
        let out_y = (self.acc_y / divisor as i32) as i16;

        self.acc_x -= out_x as i32 * divisor as i32;
        self.acc_y -= out_y as i32 * divisor as i32;

        (out_x, out_y)
    }

    fn reset(&mut self) {
        self.acc_x = 0;
        self.acc_y = 0;
    }
}

enum PointingMode {
    Cursor,
    Scroll {
        config: ScrollConfig,
        accumulator: MotionAccumulator,
    },
}

impl PointingMode {
    fn generate_report(&mut self, x: i16, y: i16) -> MouseReport {
        match self {
            Self::Cursor => MouseReport {
                buttons: 0,
                x: x.clamp(i8::MIN as i16, i8::MAX as i16) as i8,
                y: y.clamp(i8::MIN as i16, i8::MAX as i16) as i8,
                wheel: 0,
                pan: 0,
            },
            Self::Scroll { config, accumulator } => {
                let (pan, wheel) = accumulator.process(x, y, config.delta_per_tick);
                MouseReport {
                    buttons: 0,
                    x: 0,
                    y: 0,
                    wheel: wheel.clamp(i8::MIN as i16, i8::MAX as i16) as i8,
[...]

PR #728, opened 2026-02-11
Add per-layer-pointing-mode

Raymond8196:
## Summary                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                  
  Add per-layer pointing device modes to RMK, allowing trackballs to behave differently on different layers (e.g., scroll on layer 1, precision mode on layer 2).                                                                                                                 
                                                                                                                                                                                                                                                                                  
  ## Key Features                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                  
  **Per-Layer Pointing Modes:**                                                                                                                                                                                                                                                   
  - **Cursor**: Normal trackball mode (default)
  - **Scroll**: Convert motion to scroll wheel events
  - **Sniper**: Reduced sensitivity for precision control

  **Usage:**
  ```rust
  processor.set_layer_mode(1, PointingMode::Scroll(ScrollConfig::default()));
  processor.set_layer_mode(2, PointingMode::Sniper(SniperConfig::default()));

  Configure MO(1) and MO(2) keys in keymap. Fully Vial compatible.

  Technical Details:
  - Sub-pixel motion accumulation prevents precision loss
  - Configurable sensitivity divisors for each mode
  - Memory overhead: ~17 bytes (4-layer keyboard)
  - Backward compatible: default behavior unchanged

  Test Plan

  - 23 unit tests for mode switching, accumulation, and edge cases
  - All existing tests pass, zero clippy warnings
  - Tested with PMW33xx/PMW3610 examples
  - Verified Vial compatibility

  Related Issues

  Addresses #703

### DIFF
--- a/docs/docs/main/docs/configuration/input_device/pmw33xx.md
+++ b/docs/docs/main/docs/configuration/input_device/pmw33xx.md
@@ -143,17 +143,74 @@ This should be added to the `central.rs`-File even if the sensor is on split per
 :::
 
 ```rust
-    use rmk::input_device::pointing::{ PointingProcessor, PointingProcessorConfig };
+use rmk::input_device::pointing::{
+    PointingProcessor, PointingProcessorConfig, PointingMode, ScrollConfig, SniperConfig
+};
 
-    let pmw3360_proc_config = PointingProcessorConfig {
-        // invert_x: true, // invert axis if neccesary
-        // invert_y: true,
-        // swap_y: true,
-        ..Default::default()
-    };
+let pmw3360_proc_config = PointingProcessorConfig {
+    // invert_x: true, // invert axis if necessary
+    // invert_y: true,
+    // swap_xy: true,
+    ..Default::default()
+};
 
-    let mut pmw3360_processor = PointingProcessor::new(&keymap, pmw3360_proc_config);
+let mut pmw3360_processor = PointingProcessor::new(&keymap, pmw3360_proc_config);
 
-    run_all!(pmw3360_processor, /* other processors and devices */)
+run_all!(pmw3360_processor, /* other processors and devices */)
 ```
+
+## Per-Layer Pointing Modes
+
+The `PointingProcessor` supports configuring different pointing behaviors for each layer. This is useful for:
+
+- **Gaming**: Normal cursor on layer 0, sniper mode on layer 1
+- **Productivity**: Cursor on layer 0, scroll mode on layer 1 for document navigation
+- **CAD/Design**: Different precision levels for different tasks
+
+### Available Modes
+
+- **Cursor mode** (default): Normal mouse movement
+- **Scroll mode**: Movement becomes scroll wheel/pan
+- **Sniper mode**: Precision mode with reduced sensitivity
+
+### Example Configuration
+
+```rust
+use rmk::input_device::pointing::{
+    PointingProcessor, PointingProcessorConfig, PointingMode, ScrollConfig, SniperConfig
+};
+
+let mut pointing_processor = PointingProcessor::new(&keymap, PointingProcessorConfig::default());
+
+// Configure different modes for each layer
+pointing_processor
+    .set_layer_mode(0, PointingMode::Cursor)                    // Layer 0: Normal cursor
+    .set_layer_mode(1, PointingMode::Scroll(ScrollConfig {
+        divisor_x: 8,  // Pan sensitivity (higher = slower)
+        divisor_y: 8,  // Wheel sensitivity (higher = slower)
+    }))
+    .set_layer_mode(2, PointingMode::Sniper(SniperConfig {
+        divisor: 4,    // Precision divisor (higher = slower)
+    }));
+```
+
+### Mode Details
+
+**Cursor Mode**
+- Direct 1:1 mapping of sensor movement to cursor movement
+- Default mode for all layers
+
+**Scroll Mode**
+- X-axis → horizontal pan, Y-axis → vertical scroll
+- `divisor_x` and `divisor_y` control sensitivity (recommended: 4-16, default: 8)
+- Perfect for document browsing and web navigation
+
+**Sniper Mode**
+- Reduces movement speed for precision
+- Single divisor for both axes (recommended: 2-8, default: 4)
+- Ideal for gaming, CAD, or detailed work
+
+::: tip
+Use momentary layer keys (`MO(n)`) in your keymap to temporarily activate different pointing modes. The motion accumulator automatically resets when switching layers to ensure smooth transitions.
+:::
 
--- a/docs/docs/main/docs/configuration/input_device/pmw33xx.md
+++ b/docs/docs/main/docs/configuration/input_device/pmw33xx.md
@@ -148,9 +148,10 @@ use rmk::input_device::pointing::{
 };
 
 let pmw3360_proc_config = PointingProcessorConfig {
-    // invert_x: true, // invert axis if necessary
-    // invert_y: true,
-    // swap_xy: true,
+    device_id: 0,        // Match the id set on the PointingDevice (default 0)
+    // invert_x: true,   // Invert X axis globally (all modes)
+    // invert_y: true,   // Invert Y axis globally (all modes)
+    // swap_xy: true,    // Swap X and Y axes globally
     ..Default::default()
 };
 
@@ -186,11 +187,15 @@ let mut pointing_processor = PointingProcessor::new(&keymap, PointingProcessorCo
 pointing_processor
     .set_layer_mode(0, PointingMode::Cursor)                    // Layer 0: Normal cursor
     .set_layer_mode(1, PointingMode::Scroll(ScrollConfig {
-        divisor_x: 8,  // Pan sensitivity (higher = slower)
-        divisor_y: 8,  // Wheel sensitivity (higher = slower)
+        divisor_x: 8,    // Pan sensitivity (higher = slower)
+        divisor_y: 8,    // Wheel sensitivity (higher = slower)
+        invert_x: false, // Set true to reverse horizontal pan direction
+        invert_y: false, // Set true to reverse scroll wheel direction
     }))
     .set_layer_mode(2, PointingMode::Sniper(SniperConfig {
-        divisor: 4,    // Precision divisor (higher = slower)
+        divisor: 4,      // Precision divisor (higher = slower)
+        invert_x: false, // Set true to reverse X movement in sniper mode
+        invert_y: false, // Set true to reverse Y movement in sniper mode
     }));
 ```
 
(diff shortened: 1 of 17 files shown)


rmk/src/input_device/pointing.rs
```diff
@@ -238,6 +238,88 @@ impl<S: PointingDriver> PointingDevice<S> {
...
+impl MotionAccumulator {
+    /// Reset accumulator (call when mode changes)
+    pub fn reset(&mut self) {
+        self.remainder_x = 0;
+        self.remainder_y = 0;
+    }
+
+    /// Accumulate motion and return the divided output, keeping remainder
+    pub fn accumulate(&mut self, dx: i16, dy: i16, divisor_x: u8, divisor_y: u8) -> (i16, i16) {
+        let div_x = divisor_x.max(1) as i16;
```
Schievel1: Maybe we can use 0 as an indicator to switch off this axis in the given mode. This is how I use trackball with qmk: In scroll mode there is no panning. 
On some keyboards (i.e. mine) it can be ergonomically difficult to move the trackball vertically without moving it horizontally accidentally.

rmk/src/input_device/pointing.rs
```diff
@@ -238,6 +238,88 @@ impl<S: PointingDriver> PointingDevice<S> {
...
+        let div_y = divisor_y.max(1) as i16;
+
+        let total_x = self.remainder_x.saturating_add(dx);
+        let total_y = self.remainder_y.saturating_add(dy);
+
+        let out_x = total_x / div_x;
+        let out_y = total_y / div_y;
+
+        self.remainder_x = total_x - out_x * div_x;
```
Schievel1: First we divide by div_x then we multiply by div_x to normalize against total_x. By changing the order of operations we can probably avoid this.

rmk/src/input_device/pointing.rs
```diff
@@ -289,16 +434,64 @@ impl<'a, const ROW: usize, const COL: usize, const NUM_LAYER: usize, const NUM_E
...
+                    wheel: 0,
+                    pan: 0,
+                }
+            }
         };
+
         KEYBOARD_REPORT_CHANNEL.send(Report::MouseReport(mouse_report)).await;
     }
+
+    async fn on_layer_change_event(&mut self, event: LayerChangeEvent) {
```
Schievel1, line 550: Instead of keeping track of layers here, I would prefer having three new events similar to the PointingSetCpiEvent but for PointingProcessors instead of PointingDevice. Those events would be PointingSetCursor, -Scrolling, -Sniping and the PointingProcessor listens to them to set the mode. 
This way we stay more flexible and users can implement controllers and activate the modes from there however they like. They are not bound to layers, they could activate the modes with a special key combination, user keys or whatever comes to their mind.

When implementing this please keep in mind that there could be more than one PointingProcessor on a keyboard, so you need a way to identify them and to send events to them separately. (Similar to what is done in the PointingDevices)

rmk/src/input_device/pointing.rs
```diff
@@ -238,6 +238,88 @@ impl<S: PointingDriver> PointingDevice<S> {
...
+    /// Sniper mode - XY maps to cursor but at reduced sensitivity
+    Sniper(SniperConfig),
+}
+
+/// Configuration for scroll mode
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct ScrollConfig {
+    /// Divisor for X axis (pan). Higher = slower scrolling. 0 treated as 1.
+    pub divisor_x: u8,
```
Schievel1, line 262: Maybe make those i8 so users can invert the scrolling, panning independently
(Do the calculations in accumulator still work then?)

docs/docs/main/docs/configuration/input_device/pmw33xx.md
```diff
@@ -143,17 +143,74 @@ This should be added to the `central.rs`-File even if the sensor is on split per
...
+```rust
+use rmk::input_device::pointing::{
+    PointingProcessor, PointingProcessorConfig, PointingMode, ScrollConfig, SniperConfig
+};
+
+let mut pointing_processor = PointingProcessor::new(&keymap, PointingProcessorConfig::default());
+
+// Configure different modes for each layer
+pointing_processor
+    .set_layer_mode(0, PointingMode::Cursor)                    // Layer 0: Normal cursor
```
Schievel1, line 188: As I understand it everything should be possible to configure via keyboard.toml as well. Mostly because once you go configuring in Rust you can't use keyboard.toml anymore. So if users want to use those layer modes with their pointing processor they would be forced to convert their whole configuration to rust. 

But see my comment in pointing.rs suggesting to only listen for events in PointingProcessor and let the users program controllers to emit them. This way we can avoid this whole configuration section.
Controllers must be written in Rust, but they can be combined with keyboard.toml configuration.

Schievel1 (commented):
Thanks for doing this. I was about to do a similar thing, but I use scrolling mode on my trackball rather seldom. So the pressure of not having it wasn't motivating me enough. :)

I find sticking this to layers a bit too restrictive, also PointingProcessor is doing a bit too much I think. A more granular approach with different structs, where one, a controller, handles what should be done in a given situation (e.g. a layer change, a special key code etc.) and PointingProcessor just taking commands and executing them (the mode change) is more user friendly and flexible imo. 

Also I do not know what this Gazell stuff is about. Is it there accidentally and belongs to a different PR actually?

Raymond8196:
> Thanks for doing this. I was about to do a similar thing, but I use scrolling mode on my trackball rather seldom. So the pressure of not having it wasn't motivating me enough. :)
> 
> I find sticking this to layers a bit too restrictive, also PointingProcessor is doing a bit too much I think. A more granular approach with different structs, where one, a controller, handles what should be done in a given situation (e.g. a layer change, a special key code etc.) and PointingProcessor just taking commands and executing them (the mode change) is more user friendly and flexible imo.
> 
> Also I do not know what this Gazell stuff is about. Is it there accidentally and belongs to a different PR actually?

Thanks for pointing that out! The Gazell code was accidentally included — it belongs to another experimental branch and isn't relevant to this PR. I have removed it and make sure only the intended changes are included.

rmk/src/input_device/pointing.rs
Raymond8196: great suggestion! implemented it in commit 323f46cd: now setting ScrollConfig.divisor_x/y to 0 completely disables output for that axis (no movement and no remainder accumulation).

rmk/src/input_device/pointing.rs
Raymond8196: Thanks for the suggestion! I kept explicit invert_x/invert_y bools for the following reasons:
    1.Readability — invert_y: true is clearer in intent than divisor_y: -8
    2.divisor = 0 already has special meaning (disables the axis in ScrollConfig). Adding negative values as another special case increases cognitive load.
The accumulator math stays simple this way — it only deals with unsigned divisors, and inversion is applied as a straightforward sign flip afterward.

kot149: I think it's better to have multipliers in addition to divisors. That allows more flexible configurations.
And set multiplier = 0 to disable the axis, that's more intuitive than divisor = 0.

Schievel1:
I still think this should be bound to events switching the modes on instead of layers. Users can always implement their own processor that changes modes with layers, if that is really needed but the built-in stuff should be as generic as it gets.

rmk/src/input_device/pointing.rs
Raymond8196: > I think it's better to have multipliers in addition to divisors. That allows more flexible configurations. And set multiplier = 0 to disable the axis, that's more intuitive than divisor = 0.

Good point — using multiplier = 0 to disable an axis is indeed more intuitive than divisor = 0.
However, I prefer to stick with divisors only for the following reasons:

   1. Avoid overflow on embedded systems: If we used a multiplier, an input value multiplied by the multiplier could overflow 8-bit or 16-bit integers, forcing us to promote to 32-bit on every motion event. Using only divisors keeps the per-event arithmetic to a single integer division.
   2.  Consistency with existing firmware: Using divisor = 0 as a disable flag is already a common pattern in other firmware (e.g., QMK), so it's familiar to developers and reduces surprise.

kot149: > overflow on embedded systems

I'm not sure if it actually produces such large values that it overflows, but it's worth noting that ZMK input processor supports both multipliers and divisors and handles them in uint16.
https://github.com/zmkfirmware/zmk/blob/main/app/src/pointing/input_processor_scaler.c#L25-L40


> Using divisor = 0 as a disable flag is already a common pattern in other firmware (e.g., QMK), so it's familiar to developers and reduces surprise.

I'm against it. Even if I had experience using divisor = 0 in QMK or other firmware, I wouldn't know if RMK is the same until I read the exception explained in the docs, and would brace myself for a potential division-by-zero error. That could be stress for users.

vinniefranco:
Any updates on this effort?

rmk/src/input_device/pointing.rs
scottlamb: I'm working on support for iqs5xx trackpads. They're a bit different than the existing pointing devices: we probably shouldn't send them commands to change the CPI during runtime, because sending them commands can stall the I2C bus until they're ready to open a communication window, which would also stall other devices (some SoflePLUS2 configs have a trackpad and OLED on the same I2C bus). Instead, I think we should configure the hardware for its maximum resolution at init time, leave the hardware config alone, and do scaling for cursor mode as well as scrolling/sniper modes.

...and following from that, it's helpful to have more precision than offered by the divisor-only configuration. The jumps from input/1 to input/2 and from input/2 to input/3rd are nasty.

Locally, I have a change that does multiplication and divmod, and I haven't noticed problems running it once every ~10 ms. But let's say we need to optimize it. At least on the RP2040, (after doing a bit of online research) my understanding is 8-bit vs 16-bit vs 32-bit operands makes no cycle difference. Multiplication is single-cycle. Division is either 8-cycle (if you have linked in the `rp2040-hal` crate which offers optimized versions of calls like `__divsi3` that use a fixed-cycle divide peripheral) or 100+ cycles (if you have not linked in that crate and thus are just doing what the Cortex-M0 offers). There also may be function call indirection involved in all of these ops. In any case, the division is the most expensive part, which I'd expect to hold for other microcontrollers as well.

How about defining scaling in terms of i32 multiplication and right shift instead? That would improve both precision and performance.

docs/docs/main/docs/configuration/input_device/pmw33xx.md
```diff
@@ -143,17 +143,84 @@ This should be added to the `central.rs`-File even if the sensor is on split per
...
-    let mut pmw3360_processor = PointingProcessor::new(&keymap, pmw3360_proc_config);
+let mut pmw3360_processor = PointingProcessor::new(&keymap, pmw3360_proc_config);
 
-    run_all!(pmw3360_processor, /* other processors and devices */)
+run_all!(pmw3360_processor, /* other processors and devices */)
 ```
 
+## Per-Layer Pointing Modes
+
+The `PointingProcessor` supports configuring different pointing behaviors for each layer. This is useful for:
```
scottlamb, line 165: How would you feel about say taking a path to a Rust function/trait impl that takes a context (allowing it to get the layer and other things) and returns the config to apply, kind of similar to the OLED config's `renderer`? This would allow the per-layer stuff here (with a few lines of Rust code) and other options, such as:

* sniper mode based on a modifier key (the SoflePLUS2's official firmware has this feature)
* sniper mode while dragging
* applying a scroll config during a trackpad 2-finger scroll gesture (I'm working on trackpad support)

Schievel1:
I will try to do another PR replacing this, since the underlying upstream moved to much in the meantime.

Schievel1:
"Verified Vial compatibility"

What does that mean? 
There's no vial stuff in this PR, is it?

Schievel1:
I think this can be closed now since we merged #874

HaoboGu:
Yes, closing it.

PR #730, opened 2026-02-13
add pointing control event

Schievel1:
I added only the SetCpi event for the time being. 

If that is the desired approach I can add a few more. 

@kot149 maybe you can help me with the set_force_awake-mode? I am not sure how to do it and can't test.

### DIFF
--- a/rmk/src/input_device/pmw33xx.rs
+++ b/rmk/src/input_device/pmw33xx.rs
@@ -362,19 +362,6 @@ where
         }
     }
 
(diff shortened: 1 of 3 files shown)


kot149:
> @kot149 maybe you can help me with the set_force_awake-mode?

yes, but I think it's fine to wait until the specs are clearer, as there's not much demand to change force_awake mode at runtime, it's sufficient if it can be configured at compile time for most cases

Issue #733, opened 2026-02-18
Feature: Add bilateral keys override


alxzh:
When the unilateral_tap is enabled it is sometimes useful to allow same hand keys to be accessible with the modifiers. For example if I have Home Row Mod with the Command/Win/Meta/Super and the Tab on the same hand I want the unilateral_tap to be ignored for the Tab button, but be in effect for the rest of the buttons. 

It probably makes sense to put this into the morse behavior section as it only affects the morse settings. I'm thinking it can be something like following:
```
[behavior.morse]
bilateral_keys = [[3, 2], [3, 4], [3, 5], [3, 7]]
```
here the numbers are row and column positions of the keys.

alxzh:
Actually after looking through the code and understanding it better it is probably better to just make this part of the layout.matrix_map configuration. Morse profiles are very tightly packed and something like bilateral overrides would cause a regression there. Then I looked at how QMK implemented handedness for the chordal holds https://docs.qmk.fm/tap_hold#chordal-hold-handedness and thought that we can do something similar just introduce a marker in the matrix_map to mark keys as bilateral.

HaoboGu:
Yes, we already have this marker: https://rmk.rs/main/docs/configuration/layout.html

PR #735, opened 2026-02-19
feat: Adding bilateral marker into matrix_map

alxzh:
This diff covers the proposed change from issue #733. The idea is that with unilateral_tap some keys are desirable to bypass the unilateral_tap logic. For example: when in Home Row Mode on MacOS holding  Cmd key I want to press Tab key which is on the same hand. With current unilateral_tap implementation user needs to wait for the timeout to access the Tab key with modifier. 

This change mimics what QMK [chordal holds](https://docs.qmk.fm/tap_hold#chordal-hold-handedness) does, nominating some keys as exempt from Left or Right designation. 

I use it it my keymap like this:
```
matrix_map = """
(0,0,L)  (0,1,L)  (0,2,L)  (0,3,L)  (0,4,L)    (0,5,R)  (0,6,R)  (0,7,R)  (0,8,R)  (0,9,R)
(1,0,L)  (1,1,L)  (1,2,L)  (1,3,L)  (1,4,L)    (1,5,R)  (1,6,R)  (1,7,R)  (1,8,R)  (1,9,R)
(2,0,L)  (2,1,L)  (2,2,L)  (2,3,L)  (2,4,L)    (2,5,R)  (2,6,R)  (2,7,R)  (2,8,R)  (2,9,R)
                  (3,2,*)  (3,3,L)  (3,4,*)    (3,5,*)  (3,6,R)  (3,7,*)
"""
```


HaoboGu:
lgtm, thanks very much!

PR #736, opened 2026-02-20
feat: add Kinesis Advantage 360 Pro split BLE example

Filip-Pajalic:
Add a working configuration for the Kinesis Advantage 360 Pro, a wireless split keyboard with nRF52840 MCUs and Adafruit nRF52 bootloader.

Key details:
- 5x20 matrix (10 cols per half), 4 layers, BLE split
- Storage explicitly set to 0xEC000 to avoid overlap with the ~426KB central firmware binary (default 0x60000 causes crash loop)
- async_matrix disabled due to GPIOTE channel conflicts on this board
- P0.13 output pin enabled to power the LED MOSFET on both halves
- Includes UF2 build pipeline via cargo-make
- README covers build, flash, BLE pairing, storage warning, and recovery procedure for switching back to ZMK/Clique

### DIFF
--- a/examples/use_config/nrf52840_ble_split_kinesis360/.cargo/config.toml
+++ b/examples/use_config/nrf52840_ble_split_kinesis360/.cargo/config.toml
@@ -1,0 +1,12 @@
+[target.'cfg(all(target_arch = "arm", target_os = "none"))']
+runner = "probe-rs run --chip nRF52840_xxAA"
+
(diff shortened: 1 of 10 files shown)


Filip-Pajalic:
Sorry opened this PR by mistake, it is 100% not ready for a PR. The idea behind it though is to add suppot for the kinesis 360 pro.

PR #737, opened 2026-02-21
Fix `rmk.mouse_wheel_interval` and `rmk.mouse_key_interval` doing nothing

nik-rev:
Documentation of these settings states:

    - mouse_key_interval: Mouse key interval in milliseconds, default value is 20. This parameter controls the mouse movement speed; lower values result in faster movement.
    - mouse_wheel_interval: Mouse wheel interval in milliseconds, default value is 80. This parameter controls the scrolling speed; lower values result in faster scrolling. 

These settings are currently unused, however - and do nothing at all.

This PR changes that by using the user's settings instead of hard-coded defaults


HaoboGu (approved):
Great catch, thanks a lot!

PR #742, opened 2026-02-25
Fix wrong early decision of morse key

### LINKED ISSUE #740: hold-after-tap for morse not working
None of the `hold_after_tap` in the following configurations works for me:
```toml
morse = { hold_timeout = "200ms", permissive_hold = true, unilateral_tap = false, morses = [
	{ tap = "Enter", hold = "MO(1)", hold_after_tap = "Enter" },
	{ tap = "Enter", hold = "MO(1)", hold_after_tap = "Enter", double_tap = "Enter" },
	{ tap = "Enter", hold = "MO(1)", hold_after_tap = "Enter", double_tap = "No" },
	{ morse_actions = [
		{ pattern = ".", action = "Enter" },
		{ pattern = "-", action = "MO(1)" },
		{ pattern = ".-", action = "Enter" },
	] },
	{ morse_actions = [
		{ pattern = ".", action = "Enter" },
		{ pattern = "-", action = "MO(1)" },
		{ pattern = "..", action = "Enter" },
		{ pattern = ".-", action = "Enter" },
	] }
]}
```

I'm not sure if this is an expected behavior or a bug. Please close this issue if it is expected behavior.

## Steps to reproduce

1. Define morses as shown above in the `keyboard.toml`
2. Add them as `TD(n)` to the keymap
3. Build and flash the firmware (with `clear_layout = true` to reset the vial settings)
4. Tap the `TD(n)` key
5. Hold down the `TD(n)` key very shortly after the tap

## Expected behavior

- On the first tap: `Enter` is sent
- After holding down: `Enter` is repeated

## Actual behavior

- On ther first tap: `Enter` is sent
- After holding down: layer 1 is activated (`MO(1)`) and `Enter` is not sent

## Other information

I'm using the `95a3cdc` commit of the rmk.

HaoboGu:
Fixes #740

### DIFF
--- a/rmk/src/keyboard/morse.rs
+++ b/rmk/src/keyboard/morse.rs
@@ -116,6 +116,26 @@ impl<'a, const ROW: usize, const COL: usize, const NUM_LAYER: usize, const NUM_E
                             pattern.followed_by_tap()
                         };
 
+                        // If the computed pattern is beyond all configured patterns (not in
+                        // actions and no configured pattern extends it), reset to base tap/hold.
+                        // This handles re-tapping after an early-fired tap when the continuation
+                        // pattern (e.g., double_tap) isn't configured.
+                        let pattern = match &k.action {
+                            KeyAction::Morse(idx) => {
+                                let km = self.keymap.borrow();
+                                if let Some(morse) = km.behavior.morse.morses.get(*idx as usize) {
+                                    if !morse.has_pattern_or_continuation(pattern) {
+                                        if hold { HOLD } else { TAP }
+                                    } else {
+                                        pattern
+                                    }
+                                } else {
+                                    pattern
+                                }
+                            }
+                            _ => pattern,
+                        };
+
                         let final_action =
                             Self::try_predict_final_action(self.keymap.borrow().behavior, &k.action, pattern);
                         if let Some(action) = final_action {
--- a/rmk/src/morse.rs
+++ b/rmk/src/morse.rs
@@ -274,6 +274,12 @@ impl Morse {
             .is_some_and(|a| *a == *current_action)
     }
 
+    /// Returns true if the pattern exists in the actions or if any configured
+    /// pattern is a continuation of (starts with) the given pattern.
+    pub fn has_pattern_or_continuation(&self, pattern: MorsePattern) -> bool {
+        self.actions.iter().any(|(p, _)| p.starts_with(pattern))
+    }
+
     pub fn get(&self, pattern: MorsePattern) -> Option<Action> {
         self.actions.get(&pattern).copied()
     }
--- a/rmk/tests/keyboard_morse_tap_dance_test.rs
+++ b/rmk/tests/keyboard_morse_tap_dance_test.rs
@@ -69,10 +69,10 @@ pub fn create_tap_dance_test_keyboard() -> Keyboard<'static, 1, 4, 2> {
 /// This triggers the early fire optimization: tap fires immediately on release,
 /// hold_after_tap still works on re-press.
 /// Uses HoldOnOtherPress mode to reproduce the double-press bug scenario.
-fn create_early_fire_keyboard() -> Keyboard<'static, 1, 5, 2> {
+fn create_early_fire_keyboard() -> Keyboard<'static, 1, 6, 2> {
     let keymap = [
-        [[k!(A), k!(B), k!(C), k!(D), KeyAction::Morse(0)]],
-        [[k!(Kp1), k!(Kp2), k!(Kp3), k!(Kp4), k!(Kp5)]],
+        [[k!(A), k!(B), k!(C), k!(D), KeyAction::Morse(0), KeyAction::Morse(1)]],
+        [[k!(Kp1), k!(Kp2), k!(Kp3), k!(Kp4), k!(Kp5), k!(Kp6)]],
     ];
 
     let behavior_config = BehaviorConfig {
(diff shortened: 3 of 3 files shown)


kot149:
Thanks, `hold_after_tap` works fine now.

Just one thing: when the `tap` and `tap_after_hold` have the same key(action), and the `double_tap` action is not set, the first tap can send the key immediately after the tap, without waiting for timeout (and after that, `hold_after_tap` repeats the key).
Is it possible to make it work like this?

HaoboGu:
> Just one thing: when the tap and tap_after_hold have the same key(action), and the double_tap action is not set, the first tap can send the key immediately after the tap, without waiting for timeout (and after that, hold_after_tap repeats the key).
Is it possible to make it work like this?

You mean, when holding a key after tapping a key, the keyboard's output pattern is like A(First tap) + AAAAAAAAAAAA(Hold after tapping)?

kot149:
> You mean, when holding a key after tapping a key, the keyboard's output pattern is like A(First tap) + AAAAAAAAAAAA(Hold after tapping)?

Yes

HaoboGu:
Yes, we can add a state like `early tap triggered` and keep tracking mouse state, but things become complex when the following key is not the same key or timeouts. That needs some special processing, I'm not sure it's worth to do it.

I'll try

kot149:
Thanks. I think it's worth doing it, delay on single-tap is a big disadvantage of tap-dances/morses and there's demand (at least from me) to eliminate it when it's possible.

Another approach might be to add `eager_mode` option globally to morse feature: an option that forces every taps to send the key immediately without waiting for the timeout. In other words, instead of determining to send tap key immediately automatically by checking whether the next hold key is the same & next tap key is none, the user manually specifies it.
This option is useful when using tap-dance for switching something on/off on double-tap/single-tap, for example, single-tap to toggle IME off, and the double-tap to toggle IME on. In this case, it is no problem if the key is sent at the first tap even if it's actually a double-tap, because the double-tap action overrites the effect of the single-tap action.

HaoboGu:
> Thanks. I think it's worth doing it, delay on single-tap is a big disadvantage of tap-dances/morses and there's demand (at least from me) to eliminate it when it's possible.

Yes, delay on single-tap needs to be removed, what I'm not sure is that whether to make hold_after_tap produces A + AAAAA, not AAAAAA in your case.

For me, just producing AAAAAAA is what I expected.

HaoboGu:
Oh I just realized that current fix introduces delay on single-tap, nvm

HaoboGu:
@kot149 Could you try the latest commit?

kot149:
It looks good. Thank you!

HaoboGu:
I'll merge it tomorrow if it works well

alxzh:
the latest change has a bug, sometimes, very often actually it reports double press after one tap. Here is my config for reference https://github.com/alxzh/rmk-config-saurus/blob/main/keyboard.toml

alxzh:
Actually here is a better link, because I reverted last change https://github.com/alxzh/rmk-config-saurus/commit/45ef74a816aed44b32fa1320427253ec26cb0508

HaoboGu:
> the latest change has a bug, sometimes, very often actually it reports double press after one tap. Here is my config for reference https://github.com/alxzh/rmk-config-saurus/blob/main/keyboard.toml

It generates the second tap when you tap only once?

alxzh:
Yes. For example if I type “all” where “a” is the tap dance key it produces “aall”. This happens pretty consistently, but not every time.

HaoboGu:
> Yes. For example if I type “all” where “a” is the tap dance key it produces “aall”. This happens pretty consistently, but not every time.

Could you try the latest commit? @alxzh

alxzh:
OK, just tried now, the double report is gone, but now it looks like there is some timing issue, at least with the permissive hold. E.g. in my keymap when I press 'r' followed by 'd' with overlap, but 'r' released first then 'd': 
- on main branch I get 'rd', as expected.
- with this change I get 'dr', for some reason.

Hope this helps.

alxzh:
It seems to work better now, I'll test more later today and report here.

alxzh:
Another issue is now (wouldn't be surprised if I need to fix my config) if I have single tap, hold and hold after setup trying to type "account" (where 'c' is a tap-dance) second 'c' is not produced on the double tap. To get second 'c' I need to wait between taps. The same happens with multiple taps on tap dance keys, e.g. 3, 4 or 5 quick taps on the tap dance keys produce just one tap.

I would assume Morse sequence should be interrupted after first not matching sequence.

alxzh:
[morse.patch](https://github.com/user-attachments/files/25640726/morse.patch)

This patch seems to fix this problem. It basically checks if given morse sequence is registered like double tap, and if not it short cuts the previous one applying whatever the decision would be and starts a new one.

alxzh:
There seems to be one more bug even with that change. Some weird keys reordering typing 'count' I get 'cuont'. The 'c' is TD key as before. I'll write test that reproduces this.

alxzh:
Interestingly enough if I do ‘c’ held then ‘o’ and ‘u’ roll then release of ‘c’ - this always produces “cou”. But if ‘c’ gets released before roll ends it reorders the ‘o’ and ‘u’. It is kind of hard to to catch what exact sequence causes the issue, because sometimes it produces expected sequence. But before this change the output was consistent regardless of timing (at least I didn’t notice this happening). 

I’ll try to come up with a test that reproduces this.

alxzh:
After running several monkeytype tests I don't see issue with any other rolls, just this one, may be it is just me typing it wrong. I think that with my proposed patch it all works and should be good to merge. Let me know if you want me to test any other changes with this.

HaoboGu:
> [morse.patch](https://github.com/user-attachments/files/25640726/morse.patch)
> 
> This patch seems to fix this problem. It basically checks if given morse sequence is registered like double tap, and if not it short cuts the previous one applying whatever the decision would be and starts a new one.

I've applied this patch, could you test it again? Thanks a lot!

alxzh:
Yes, everything seems to be working as expected now. Thank you for quick responses!

Issue #450, opened 2025-07-08
OLED Screen suport.


lucas11222:
add suport for adding an OLED screen.
Like for example showing how many times you touch the key.

pcasotti:
It's technically possible to do on the user side using controllers.

If we want to add it in the library we need to think how will the display options be available to the user (how will they customize the screen) both via rust and toml.

Also it should be somewhat generic over the type of display, so we can add support for spi an others later on.

lucas11222:
also we can inspire with the way that QMK uses for images in the oled screens
https://docs.qmk.fm/features/oled_driver
https://www.youtube.com/watch?v=OJSOEStpPIo

alex-tdrn:
I also think letting the user choose the driver will be important, as for e.g. the [nice!view](https://nicekeyboards.com/nice-view/), there exists a [rust driver](https://crates.io/crates/sharp-memory-display) that is licensed under the GPL.

RMK would not be able to pull in this dependency itself (e.g. via a feature flag) without RMK becoming GPL itself. But end-user's firmware might as well be GPL, since they're binaries not libraries.

Also since we're at it, I guess a fancy interface based on [embedded-graphics](https://github.com/embedded-graphics/embedded-graphics) could be offered by RMK, with some built-in widget options, with the user being responsible for plugging in the right driver.

Schievel1:
In user code I got the ssd1306 working.

For toml config, how about  something like an expanded controller makro? That you can give the display type and it expands to pulling in all the dependencies for that display and a polling controller which is used to manipulate the displays frame buffer.

HaoboGu:
We have display and OLED support now :D

PR #782, opened 2026-03-29
feat: add OLED display support with DisplayRenderer trait

gbPagano:
This PR adds initial OLED display support to RMK via a new `display` feature flag. The goal is to get early feedback on the **architecture and direction** before investing in the full feature set.

### What's included

- **`DisplayRenderer` trait** — the core abstraction that allows users to fully customize what is drawn on the display:
  ```rust
  pub trait DisplayRenderer {
      fn render<D: DrawTarget<Color = BinaryColor>>(
          &mut self, ctx: &RenderContext, display: &mut D,
      );
  }
  ```
- **`RenderContext`** — a snapshot of keyboard state (layer, WPM, caps/num lock, battery) passed to renderers on every redraw, decoupled from the display driver.
- **`DefaultRenderer`** — a built-in renderer that automatically adapts between landscape and portrait layouts based on the logical display dimensions.
- **`OledDisplayProcessor<DI, SIZE, R>`** — a processor (using the existing `#[processor]` macro) generic over the renderer. `R` defaults to `DefaultRenderer`, so `::new(display)` works out of the box and `::with_renderer(display, my_renderer)` gives full control.
- **`examples/use_rust/rp2040_oled`** updated with a working I2C OLED setup.

### Architecture

```
                    ┌─────────────────────────────────────┐
  use_config ─────▶│  keyboard.toml [display]            │
  (future)          │  rmk-macro generates a configured   │
                    │  renderer at compile-time           │
                    └──────────────┬──────────────────────┘
                                   │ impl DisplayRenderer
                    ┌──────────────▼──────────────────────┐
  use_rust ───────▶│  OledDisplayProcessor<DI, SIZE, R>  │
                    │  R: DisplayRenderer                 │
                    │  R = DefaultRenderer (if omitted)   │
                    └─────────────────────────────────────┘
```

The design mirrors the two user profiles RMK already supports:
- **`use_rust`** users implement `DisplayRenderer` for full control.
- **`use_config`** users (future) would declare widgets in `keyboard.toml` and the macro would generate the renderer.

*Not yet included*: support for displays other than SSD1306 (SH1106, SSD1309, etc.), an SPI example, `keyboard.toml` configuration for widget selection and orientation, and public helper functions for custom renderers to reuse.

### Questions for review

1. Does this overall architecture (trait-based renderer + processor generic over it) align with where you'd like RMK's display support to go?
2. For the `keyboard.toml` integration: would a widget-slot model or a simpler theme-based approach be preferred?

Open to any thoughts on the direction, would love to hear feedback before moving forward.


HaoboGu:
Thanks for proposing it!

The overall architecture looks really good to me. Decoupling renderer and the processor is great.

What I'm thinking about is, can we make both processor and renderer be abled to be implemented in a third-party crate? I'm imaging:


```toml
# Cargo.toml
rmk = { version = "*", feature = ["display", ] }

rmk-ssd1306-processor = { .. } # Display driver support
rmk-custom-renderer = { .. } # Display renderer support
``` 

Then in `keyboard.toml`, we can declare something like

```toml
# keyboard.toml
[display]
processor = "Ssd1306Processor"
renderer = "CustomRenderer"

```

where `Ssd1306Processor` and `CustomRenderer` are exported in `rmk-ssd1306-processor` and `rmk-custom-renderer`.

I prefer not to configure "widgets" in keyboard.toml because it's hard to define all widgets in the RMK's display interface -- I think it might limit the design of the display UI. It should be somehow defined in the renderer crate, not in RMK

gbPagano:
Thanks for the feedback! I really like the direction of decoupling the display driver from the RMK core. Here's what I'm thinking for the architecture:

### 1. `rmk` core  - traits and default renderer only

The core would export only the generic abstractions, with no driver dependencies:

- `DisplayRenderer<C: PixelColor>` - generic over the pixel color type so it works for both monochrome OLEDs and color LCDs
- `RenderContext` - the keyboard state snapshot (unchanged)
- `DefaultRenderer` - built-in renderer implementing `DisplayRenderer<BinaryColor>`

### 2. `rmk-displays` - a single crate with feature-gated drivers

Rather than one crate per display chip, I think a single `rmk-displays` crate with features makes more sense, since `embedded-graphics` already unifies the rendering API — the processors share the same event-handling logic and only differ in the driver init/flush layer:

```toml
# User's Cargo.toml
rmk-displays = { version = "*", features = ["ssd1306"] }
```

Each feature would expose a processor (`Ssd1306DisplayProcessor<R>`, `Sh1106DisplayProcessor<R>`, etc.), all generic over the renderer. This could live in the same repo as a workspace member.

### 3. Custom renderers via `keyboard.toml`

For `use_config` users who want a custom renderer, I'm thinking the TOML would accept an optional full path:

```toml
[display]
driver = "ssd1306"
interface = "i2c"
size = "128x64"
i2c_scl = "PIN_5"
i2c_sda = "PIN_4"
# Optional - omit for DefaultRenderer
renderer = "my_crate::MyRenderer"
```

The macro would generate:
- If `renderer` is omitted: instantiate with `DefaultRenderer`
- If `renderer` is set: parse it as a Rust path and call `Default::default()` to construct it (the custom renderer must implement `Default`)

The constraint that custom renderers need `Default` is reasonable since renderers are typically stateless. For anything more complex, `use_rust` gives full control.

### 4. `use_rust`  - unchanged

Users who want full control keep using the processor directly, optionally with a custom renderer, exactly as it works today. They could also use `#[register_processor]` to wire things up manually.

---

I agree with your point about not defining widgets in `keyboard.toml`, the renderer crate is the right place for UI design decisions, not the config file.

What do you think about this approach?

HaoboGu:
In RMK core, the built-in renderer(`DefaultRenderer`) and processor(such as `Ssd1306DisplayProcessor`) can be provided, that's fine. I just want to ensure that the Renderer/Processor *can* be implemented as a third crate.

I agree with your for other points. The only comment from me is the `keyboard.toml` configuration, I prefer to group the configuration for common interfaces. For example, for a display driver using i2c:

```toml
driver = "ssd1306"
interface = "i2c"
i2c = { scl = "", sda = "",  }
size = "128x64"
```

gbPagano:
Ok! I'll move forward with the implementation and come back with updates soon

gbPagano:
Hey @HaoboGu , here's an update on the progress since our last discussion.

### `keyboard.toml` configuration

The `[display]` section uses the grouped interface config you suggested, reusing the existing `CommunicationProtocol` enum (`I2c`/`Spi`):

```toml
[display]
driver = "ssd1306"
size = "128x32"
rotation = 270
renderer = "custom_renderer::BigLayerRenderer"  # optional

[display.protocol.i2c]
instance = "I2C1"
scl = "PIN_3"
sda = "PIN_2"
address = 60
```

It also works in split configs via `[split.central.display]` and `[split.peripheral.display]`.

### Custom renderer support

As we discussed, custom renderers are referenced by path in `keyboard.toml`. I added an example library at `examples/use_rust/custom_renderer/` that the `use_config/rp2040_oled` example uses as a dependency.

### Driver re-exports

Driver crates (`ssd1306`, `oled_async`, `display-interface-i2c`) are re-exported from `rmk::display::*` so `use_config` users don't need to add them manually.

### Other changes since last push

- **Render rate-limiting** — configurable minimum interval (default 30ms) to prevent high-frequency events from overwhelming displays
- **More event subscriptions** — `KeyboardEvent`, `SleepStateEvent`, `BleStatusChangeEvent`, `PeripheralConnectedEvent`, `CentralConnectedEvent`, `PeripheralBatteryEvent`.
- **oled_async driver support** — SH1106, SH1107, SH1108, SSD1309 via feature flags

### SPI interface

The config and codegen are prepared for SPI (via `CommunicationProtocol::Spi`), but the actual SPI initialization is not implemented yet — I don't have SPI display hardware to test against, so I preferred to leave it as a clear panic rather than ship untested code.

## Next steps

If the current architecture looks good to you, my planned next steps are:

1. Improve the `custom_renderer` example to better showcase the API
2. Improve the default renderers (BLE status, split peripheral info, better layouts)
3. Write user-facing documentation for the RMK docs site

HaoboGu:
Thanks for your effort!

I just went through the code roughly, overall it's really great. Some initial comments from me:

1. In `render` the `DisplayProcessor` checks the min interval, is it possible to use the polling based processor here? If so, all `self.render().await;` in `on_xx_event` can also be removed.
2. There are `width` and `height` in `RenderContext`, I'm not sure this is the correct way, maybe associated const is better in this case? I understand that introducing many generics is verbose, but it's just kind of "strange" that display's `width` and `height` are in `RenderContext`. It should be a parameter/field of the display imo
3. `oled_async` is a git dependency. It's fine right now but if we want to release the next version, the crates.io version should be there.

And I also found a bug that might block the `DisplayProcessor`, I've pushed a fix: #787

rmk/src/display/renderer.rs
```diff
@@ -0,0 +1,300 @@
...
+    pub caps_lock: bool,
+    /// Whether Num Lock is active.
+    pub num_lock: bool,
+    /// Current battery state.
+    pub battery: BatteryStateEvent,
+    /// Whether the keyboard is sleeping.
+    pub sleeping: bool,
+    /// Current BLE connection status (profile + state).
+    #[cfg(feature = "_ble")]
+    pub ble_status: BleStatus,
```
HaoboGu: These values are feature gated, custom renderers should be aware it, that requires some extra feature gating mechanism for out-of-tree renderers. Additional documents is needed here.

rmk-macro/src/codegen/orchestrator.rs
```diff
@@ -228,8 +229,27 @@ fn expand_main(
...
     let matrix_and_keyboard = expand_matrix_and_keyboard_init(hardware);
-    let (registered_processor_initializers, registered_processors) =
+    let (registered_processor_initializers, mut registered_processors) =
         expand_registered_processor_init(hardware, &item_mod);
+
+    // Display configuration — for unibody use top-level, for split use central's config
+    let display_config = match &hardware.board {
+        BoardConfig::UniBody(_) => hardware.display.as_ref(),
+        BoardConfig::Split(split_config) => split_config.central.display.as_ref(),
+    };
```
HaoboGu, line 239: For split, the central's display config is under `[central.display]` section, and peripherals display config is under `[[split.peripheral]]`, right? 

The top-level `[display]` config is for unibody keyboards only

rmk-config/src/default_config/event_default.toml
```diff
@@ -22,7 +22,7 @@ subs = 2
 [event.keyboard]
 channel_size = 16
 pubs = 2
-subs = 2
+subs = 3
```
HaoboGu, line 25: This is increased for all keyboards with/without display. 🤔 Maybe we need a way to increase/update the default channel sizes according to the enabled features.

rmk/src/display/renderer.rs
gbPagano: I see two options here:

1. Remove the #[cfg] gates and make all fields always present, using Option for conditional data. This is simpler for third-party renderers, they just check if let Some(_) = ... without needing to match RMK's feature flags.

2. Keep the #[cfg] gates for memory efficiency and document that renderers using these fields need the corresponding features enabled.

rmk-macro/src/codegen/orchestrator.rs
gbPagano: Yes, that's correct. [display] is for unibody keyboards only, [split.central.display] for the central, and display config under [[split.peripheral]] for peripherals. 

I'll add a split example with display later, still testing everything on my end.

rmk-config/src/default_config/event_default.toml
gbPagano: I saw you opened #788 to address this. Can I consider this resolved for now?

gbPagano:
Thanks for the review!

**1. Polling-based rendering**

I'm actually working on this right now, the `DisplayProcessor` will support a configurable `render_interval` that controls the poll frequency, with `Duration::MAX` (effectively disabled) as the default. I'll push a commit soon.

That said, I don't think we should remove `self.render()` from the event handlers entirely, for two reasons:
- Some state changes need to be reflected on the display practically instantly (e.g. a key press triggering a visual response).
- Users may want to disable polling altogether to save battery and allow the keyboard to enter sleep mode. In that case, event-driven renders are the only way the display updates.

---
2. You're right, these don't belong in `RenderContext`. The renderer already receives `&mut D` which implements `DrawTarget`, so it can call `display.bounding_box().size` directly. I'll remove them.

3. The reason for the git dependency is that the SH1106 driver isn't included in the latest crates.io release yet. I'll reach out to the crate author about publishing a new version.

rmk/src/display/renderer.rs
HaoboGu: I prefer option 2

rmk-macro/src/codegen/orchestrator.rs
HaoboGu: neat!

rmk-config/src/default_config/event_default.toml
HaoboGu: Yes, let me test #788 a little bit more before merging it.

HaoboGu:
yeah, making `DisplayProcessor` support both polling and event driven mode is fine. Let's keep `self.render()` then

rmk/src/display/renderer.rs
gbPagano: I added doc comments to RenderContext explaining this

HaoboGu:
#788 was merged

gbPagano:
Finished implementing the built-in display renderers, now living under `rmk::display`:

- **`OledRenderer`**: full-featured landscape/portrait adaptive layout with modifier icons, BLE indicator, battery widget, and lock dots.
- **`LogoRenderer`**: static RMK logo splash screen. Now the **default renderer** for `DisplayProcessor`. Logo data is embedded as a `const`.

Both are exported directly from `rmk::display` and can be referenced by bare name in `keyboard.toml`:

```toml
renderer = "OledRenderer"   # instead of rmk::display::OledRenderer
renderer = "LogoRenderer"
```

The macro resolves bare names to `::rmk::display::` automatically.

| | | |
|:---:|:---:|:---:|
|  |  |  |
|  |  |  |

### Display events forwarded to split peripherals

Under the `display` feature, the central now forwards three additional events to each peripheral:

*WpmUpdateEvent*, *ModifierEvent* and *SleepStateEvent*

All three variants and their handling on both sides are gated behind `#[cfg(feature = "display")]`, so there is no impact on builds without the display feature.

Peripherals republish these as local events, allowing a peripheral-side display to render the same keyboard state (WPM, active modifiers, sleep) as the central.

gbPagano:
I think the implementation is done. What do you think, @HaoboGu? If everything looks good, only the documentation is left.

rmk/src/display/mod.rs
```diff
@@ -0,0 +1,401 @@
...
+}
+
+impl<D, R> PollingProcessor for DisplayProcessor<D, R>
+where
+    D: DisplayDriver,
+    R: DisplayRenderer<D::Color>,
+{
+    fn interval(&self) -> Duration {
+        if self.ctx.sleeping {
+            Duration::MAX
```
HaoboGu: Is there a way to wake up the keyboard display from sleeping?

rmk/src/split/driver.rs
```diff
@@ -61,139 +61,164 @@ impl<const ROW: usize, const COL: usize, const ROW_OFFSET: usize, const COL_OFFS
...
             let wait_time = if elapsed >= 3000 { 1 } else { 3000 - elapsed };
-            match select3(
-                self.transceiver.read(),
-                select3(
-                    keyboard_indicator_sub.next_event(),
-                    layer_sub.next_event(),
+
+            // Wait for the next event to forward to the peripheral.
+            // Each async block converts to SplitMessage; disabled features use pending().
+            let next_event = select6(
```
HaoboGu: There is a macro called `select_biased_with_feature` which might be useful here.

rmk/src/split/driver.rs
```diff
@@ -61,139 +61,164 @@ impl<const ROW: usize, const COL: usize, const ROW_OFFSET: usize, const COL_OFFS
...
-                            let message_to_peri =
-                                SplitMessage::KeyboardIndicator(indicator_event.indicator.into_bits());
-                            debug!("Sending message to peripheral {}: {:?}", self.id, message_to_peri);
-                            if let Err(e) = self.transceiver.write(&message_to_peri).await {
-                                match e {
-                                    SplitDriverError::Disconnected => return,
-                                    _ => error!("SplitDriver write error: {:?}", e),
-                                }
+                        // Opportunistically drain one pending event to forward
+                        let msg = if let Some(e) = indicator_sub.try_next_message_pure() {
```
HaoboGu: Totally removing the try-send message blocks after `self.process_peripheral_message(split_message).await;` is fine(I should do that, but I forgot)

rmk/src/processor.rs
```diff
@@ -94,28 +94,27 @@ pub trait PollingProcessor: Processor {
     async fn update(&mut self);
 
     /// Polling loop that processes events and calls `update()` at the specified interval.
+    ///
+    /// When `interval()` is so large that `last + interval` overflows, the timer
```
HaoboGu: I don't think so, the interval should not be that large. PollingController is not designed for this case. Please revert changes here.

HaoboGu (commented):
Thanks! I've left several comments about the polling controller, other parts are all good!

rmk/src/display/mod.rs
gbPagano: Yes, the display automatically wakes up

Since self.interval() is evaluated dynamically on each loop tick via select!(), when the keyboard receives a key event and ctx.sleeping flips to false, the next iteration returns the normal render_interval and polling resumes

The event itself is also rendered immediately via self.render().await in the event handler

rmk/src/processor.rs
gbPagano: I actually ran into panics specifically when polling was disabled (interval() = Duration::MAX to effectively turn off polling)

Any cleaner alternative to avoid Duration::MAX in polling/sleep?

rmk/src/display/mod.rs
HaoboGu: I don't get how it works, when the keyboard is sleeping, `render()` will be skipped and only peripheral publishes `SleepStateEvent`. How does the processor quit the sleeping mode?

rmk/src/processor.rs
HaoboGu: What was the panic message?

If I understand it correctly, you want the processor to switch between polling mode and event based mode by setting `interval() = Duration::MAX`? The processor can only operate in only one mode by design

rmk/src/display/mod.rs
gbPagano: The processor never truly stops while sleeping, it's still running and processing every event normally. The only difference is that when interval() returns Duration::MAX, checked_add overflows and the loop skips the Timer entirely, blocking on sub.next_event() alone.

So when the keyboard wakes up, the sleep manager publishes a new SleepStateEvent → the display processor receives it like any other event → on_sleep_state_event sets ctx.sleeping = false and calls render() → on the next loop iteration, interval() returns render_interval again and the timer-based polling resumes.

For split peripherals, the central forwards SleepStateEvent, so the peripheral-side display wakes up through the exact same path

rmk/src/processor.rs
gbPagano: The panic came from embassy-time overflowing internally when Timer::after(Duration::MAX) was called, essentially Instant::now() + Duration::MAX overflows a u64

The reason for supporting Duration::MAX (event-only mode) in PollingProcessor is intentional: if the user doesn’t need animated displays, running a periodic timer just to redraw the same static content causes unnecessary wake-ups that drain the battery. Event-driven only makes more sense in that case.

At the same time, even when polling is enabled for animations, event-driven rendering is still valuable, modifier indicators, layer changes, and BLE status updates need to be reflected instantly, not waiting for the next poll tick.

So the design is: polling + events for animated displays, pure events for static displays. The Duration::MAX path in polling_loop is what enables that without requiring two separate processor implementations.

rmk/src/display/mod.rs
HaoboGu: If I understand it correctly, the processor operates in event based mode when sleeping and polling mode when waked up? Okay that makes sense.

If my understanding is correct I prefer to change run function instead of updating interval dynamically to switch between polling mode and event based mode. Setting the interval to "disable" the polling mode is not very good.

gbPagano: Makes sense, done! Replaced PollingProcessor with a manual Runnable that explicitly branches between event-driven and polling modes

rmk/src/processor.rs
gbPagano: Reverted! The checked_add workaround is no longer needed since DisplayProcessor now implements Runnable manually and handles the event-only mode explicitly

rmk/src/display/mod.rs
HaoboGu: Great, thanks! Could you please remove the `manual_polling` option in `rmk-macro` as well?

gbPagano: done!

rmk-macro/src/event_macros/utils.rs
```diff
@@ -82,6 +82,35 @@ impl AttributeParser {
         })
     }
 
+    /// Get a boolean value for `name = true/false`.
+    pub fn get_bool(&self, name: &str) -> Result<Option<bool>, TokenStream> {
```
HaoboGu: This is not used anymore

rmk-macro/src/event_macros/runnable/generator.rs
```diff
@@ -57,11 +57,11 @@ fn build_polling_loop_body(
     };
 
     quote! {
-        let elapsed = last.elapsed();
         let interval = ::embassy_time::Duration::from_millis(#interval_ms);
-        let timer = ::embassy_time::Timer::after(
-            interval.checked_sub(elapsed).unwrap_or(::embassy_time::Duration::MIN)
-        );
```
HaoboGu, line 64: This should be reverted too

rmk/src/display/mod.rs
```diff
@@ -0,0 +1,420 @@
...
+
+impl<D, R> Runnable for DisplayProcessor<D, R>
+where
+    D: DisplayDriver,
+    R: DisplayRenderer<D::Color>,
+{
+    async fn run(&mut self) -> ! {
+        use crate::event::EventSubscriber;
+        let mut sub = <Self as Processor>::subscriber();
+        let mut last = Instant::now();
```
HaoboGu: Call the `.render` once to initialize the display before loop

rmk/src/display/mod.rs
```diff
@@ -0,0 +1,420 @@
...
+
+    /// Periodic poll — drives animations even when no events arrive.
+    async fn poll(&mut self) {
+        self.render().await;
+    }
+
+    /// Redraw the display if enough time has passed since the last render.
+    async fn render(&mut self) {
+        let now = Instant::now();
+        if now.duration_since(self.last_render) < self.min_render_interval {
```
HaoboGu, line 323: I'm thinking about it, when the two different events come at the same time, this hard limit drops the later one -- then if there's no newer events, the second event is dropped permanently and the content displayed is wrong. 

Consider totally removing the rate limiter maybe? For event based processor, all events should be processed. Dropping events seems not correct.

rmk/src/display/renderers/default_oled.rs
```diff
@@ -0,0 +1,376 @@
...
+            .draw(display)
+            .ok();
+    }
+}
+
+/// Returns true when the keyboard considers itself connected.
+fn is_connected(_ctx: &RenderContext) -> bool {
+    // Split peripheral: connected when central is paired
+    #[cfg(feature = "split")]
+    return _ctx.central_connected;
```
HaoboGu: On central it always returns `false`

rmk-macro/src/codegen/display.rs
```diff
@@ -0,0 +1,233 @@
...
+        }
+        ChipSeries::Nrf52 => {
+            quote! {
+                #instance => ::embassy_nrf::twim::InterruptHandler<::embassy_nrf::peripherals::#instance>;
+            }
+        }
+        ChipSeries::Stm32 => {
+            let ev_irq = format_ident!("{}_EV", i2c.instance);
+            quote! {
+                #ev_irq => ::embassy_stm32::i2c::EventInterruptHandler<::embassy_stm32::peripherals::#instance>;
```
HaoboGu, line 62: {}_ER should be added too, see: https://github.com/embassy-rs/embassy/blob/main/examples/stm32h7/src/bin/i2c.rs

rmk-macro/src/event_macros/runnable/generator.rs
HaoboGu: And I found a better way for polling controller:

https://docs.embassy.dev/embassy-time/git/default/struct.Ticker.html、

Note: when transitioning out of sleep, call `ticker.reset()` so the first post-sleep tick fires interval from now, not immediately (otherwise waking from sleep produces a render storm catching up missed ticks).

examples/use_rust/custom_renderer/src/bongo_cat.rs
```diff
@@ -0,0 +1,167 @@
...
+}
+
+/// Bongo Cat OLED renderer.
+///
+/// Draws the classic Bongo Cat animation on a 128×32 OLED.
+/// - **Normal**: each key press alternates left/right paw.
+/// - **Fury** (high WPM): both paws smashing, one toggle per render.
+///
+/// Animation speed is controlled by `render_interval` in `keyboard.toml`.
+/// Frame data from QMK firmware (GPL-2.0, keyboards/torn).
```
HaoboGu: Sorry we cannot use code/data from QMK project because the license is not compatible

HaoboGu (commented):
Some more minor comments from me. Overall it is ok and we can move forward

examples/use_rust/rp2040_oled/rp2040.svd
HaoboGu, line 1: I know you just copy it from other examples, could you please remove the two svd files in the added examples?

rmk/src/display/mod.rs
```diff
@@ -0,0 +1,420 @@
...
+        use crate::event::EventSubscriber;
+        let mut sub = <Self as Processor>::subscriber();
+        let mut last = Instant::now();
+
+        loop {
+            if self.ctx.sleeping || self.render_interval.is_none() {
+                let event = sub.next_event().await;
+                self.process(event).await;
+            } else {
+                let interval = self.render_interval.unwrap();
```
HaoboGu: `unwrap` here should be removed.

rmk-macro/src/codegen/display.rs
```diff
@@ -0,0 +1,233 @@
...
+fn parse_oled_async_variant(driver: &DisplayDriver, size: &str) -> TokenStream {
+    let (module, prefix) = match driver {
+        DisplayDriver::Sh1106 => ("sh1106", "Sh1106"),
+        DisplayDriver::Sh1107 => ("sh1107", "Sh1107"),
+        DisplayDriver::Sh1108 => ("sh1108", "Sh1108"),
+        DisplayDriver::Ssd1309 => ("ssd1309", "Ssd1309"),
+        _ => unreachable!(),
+    };
+
+    let variant_name = format!("{}_{}", prefix, size.replace('x', "_"));
```
HaoboGu: I think the size should be checked, users may pass illegal sizes like `12x34`, which leads to "unresolved import" error, which might be confusing.

rmk-macro/src/codegen/display.rs
```diff
@@ -0,0 +1,233 @@
...
+fn get_i2c_address(config: &DisplayConfig) -> u8 {
+    match &config.protocol {
+        CommunicationProtocol::I2c(i2c) => i2c.address,
+        CommunicationProtocol::Spi(_) => panic!("SPI display interface is not yet supported"),
+    }
+}
+
+fn expand_display_driver_init(config: &DisplayConfig) -> TokenStream {
+    let address = get_i2c_address(config);
+    let rotation = match config.rotation {
```
HaoboGu, line 130: The rotation is u16, so if the user sets a value other than `90/180/270`, the setting fails without warning.

examples/use_rust/rp2040_oled/Cargo.toml
```diff
@@ -0,0 +1,66 @@
...
+description = "Keyboard firmware written in Rust"
+homepage = "https://github.com/haobogu/rmk"
+repository = "https://github.com/haobogu/rmk"
+readme = "../../README.md"
+edition = "2024"
+license = "MIT OR Apache-2.0"
+
+[dependencies]
+rmk = { path = "../../../rmk", features = ["host", "log", "rp2040", "storage", "usb_log", "vial", "vial_lock", "sh1106"], version = "0.8.2", default-features = false }
+oled_async = { git = "https://github.com/cschuhen/oled_drivers", rev = "0aa4b3d", features = ["i2c"] }
```
HaoboGu: The released version should be used

rmk/src/display/mod.rs
gbPagano: Removed, but this branch only runs when interval is not none, so is guaranteed to exist.

examples/use_rust/custom_renderer/src/bongo_cat.rs
gbPagano: Sorry about that, I`ll remove

rmk/src/display/mod.rs
gbPagano: I do not think removing the rate limiter entirely is the best fix, because some event sources can be very noisy and would end up flushing the display too often. Instead, I  changed it to coalesce redraws: when a redraw is rate-limited, we keep a pending redraw  and flush the latest state as soon as the interval expires. That preserves the final state without hammering the display.

HaoboGu: Yeah, this solution looks better!

gbPagano:
Hey @HaoboGu, applied all your latest feedback! Also, this PR closes #450.

HaoboGu:
Great work! No more comments from me about the code. Let's move to documentation now.

gbPagano:
Added the display docs! Split into two pages following the existing pattern:
  - `features/display.md` : overview, supported drivers, renderers, Rust API, custom renderers/drivers
  - `configuration/display.md` : `keyboard.toml` reference for `[display]` fields and split keyboard setup

HaoboGu (approved):
Thanks a lot for your effort!

PR #785, opened 2026-03-29
fix: Improve BLE passkey handling for split peripherals

alxzh:
- validate passkey config only when passkey entry is enabled
- add split-peripheral-aware BLE stack setup and update affected examples
- clean up passkey session lifecycle across GATT events and keyboard tasks

### DIFF
--- a/examples/use_rust/esp32c3_ble/src/main.rs
+++ b/examples/use_rust/esp32c3_ble/src/main.rs
@@ -49,8 +49,6 @@ async fn main(_s: Spawner) {
     let central_addr = [0x18, 0xe2, 0x21, 0x80, 0xc0, 0xc7];
     let mut host_resources = HostResources::new();
     let stack = build_ble_stack(controller, central_addr, &mut rng, &mut host_resources).await;
-    #[cfg(feature = "passkey_entry")]
-    stack.set_io_capabilities(rmk::IoCapabilities::KeyboardOnly);
 
     // Initialize the flash
     let flash = FlashStorage::new(peripherals.FLASH);
--- a/examples/use_rust/esp32c3_ble/src/main.rs
+++ b/examples/use_rust/esp32c3_ble/src/main.rs
@@ -49,6 +49,8 @@ async fn main(_s: Spawner) {
     let central_addr = [0x18, 0xe2, 0x21, 0x80, 0xc0, 0xc7];
     let mut host_resources = HostResources::new();
     let stack = build_ble_stack(controller, central_addr, &mut rng, &mut host_resources).await;
+    #[cfg(feature = "passkey_entry")]
+    stack.set_io_capabilities(rmk::IoCapabilities::KeyboardOnly);
 
     // Initialize the flash
     let flash = FlashStorage::new(peripherals.FLASH);
--- a/rmk-config/src/resolved/build_constants.rs
+++ b/rmk-config/src/resolved/build_constants.rs
@@ -80,7 +80,11 @@ impl crate::KeyboardTomlConfig {
         );
 
         // Only validate passkey settings when the build will emit passkey constants.
-        let passkey = resolve_passkey(self.ble.as_ref(), std::env::var("CARGO_FEATURE_PASSKEY_ENTRY").is_ok())?;
+        let passkey = if std::env::var("CARGO_FEATURE_PASSKEY_ENTRY").is_ok() {
+            self.ble.as_ref().map(resolve_passkey_enabled).transpose()?
+        } else {
+            None
+        };
 
         Ok(BuildConstants {
             combo_max_num: rmk.combo_max_num,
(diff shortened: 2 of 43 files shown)


alxzh:
The examples seems to be broken on trunk.

HaoboGu:
> The examples seems to be broken on trunk.

Please rebase to the latest main branch

rmk/src/ble/mod.rs
```diff
@@ -108,14 +197,30 @@ pub async fn build_ble_stack<
     host_address: [u8; 6],
     random_generator: &mut RNG,
     resources: &'a mut HostResources<P, CONNECTIONS_MAX, L2CAP_CHANNELS_MAX>,
+) -> Stack<'a, C, P> {
+    build_ble_stack_with_peripheral_flag(controller, host_address, random_generator, resources, false).await
+}
+
+/// Build a BLE stack with split-peripheral-specific security capabilities.
+pub async fn build_ble_stack_with_peripheral_flag<
```
HaoboGu: I prefer not to add an additional function for this. Emitting a constant that indicates the role of the current binary(Central or Peripheral) might be better.

rmk/src/ble/mod.rs
```diff
@@ -97,6 +99,93 @@ pub(crate) const CONNECTIONS_MAX: usize = crate::SPLIT_PERIPHERALS_NUM + 1;
 /// Max number of L2CAP channels
 pub(crate) const L2CAP_CHANNELS_MAX: usize = CONNECTIONS_MAX * 4; // Signal + att + smp + hid
 
+async fn wait_for_keyboard_activity() {
+    let mut keyboard_events = KeyboardEvent::subscriber();
+    let _ = keyboard_events.next_message_pure().await;
+}
+
+struct PasskeyInputState {
```
HaoboGu, line 103: feature gate the whole `PasskeyInputState`?

rmk/src/ble/mod.rs
```diff
@@ -97,6 +99,93 @@ pub(crate) const CONNECTIONS_MAX: usize = crate::SPLIT_PERIPHERALS_NUM + 1;
 /// Max number of L2CAP channels
 pub(crate) const L2CAP_CHANNELS_MAX: usize = CONNECTIONS_MAX * 4; // Signal + att + smp + hid
 
+async fn wait_for_keyboard_activity() {
+    let mut keyboard_events = KeyboardEvent::subscriber();
```
HaoboGu: This event will only be emitted after processing the events. But in sleep state, no keyboard event will be processed, so this change is incorrect. `KEYBOARD_REPORT_CHANNEL` is the way.

rmk-config/src/resolved/build_constants.rs
```diff
@@ -120,6 +116,14 @@ fn resolve_passkey_enabled(ble: &crate::BleConfig) -> Result<Passkey, String> {
     Ok(Passkey { enabled, timeout_secs })
 }
 
+fn resolve_passkey(ble: Option<&crate::BleConfig>, enabled: bool) -> Result<Option<Passkey>, String> {
```
HaoboGu: Extracting this has no gain, makes the code not cleaner as before

rmk/src/ble/mod.rs
alxzh: Honestly trying to avoid it makes it even messier. I can try to find a way to not introduce an extra function. But I don't have a good way to do it, especially in a way that it would fit well with current examples and keyboard.toml. Any suggestions would be welcome though.

rmk/src/ble/mod.rs
alxzh: removed this

rmk/src/ble/mod.rs
alxzh: done

rmk-config/src/resolved/build_constants.rs
alxzh: removed

alxzh:
This is weird, that test is passing for me.

rmk/src/ble/mod.rs
HaoboGu: I checked the code again, `set_io_capabilities` is a public function of trouble-host `Stack`, that means that you can set it back to `None` for peripherals, which eliminates the public API change and needs only 1 extra line

HaoboGu: https://github.com/HaoboGu/rmk/blob/main/rmk/src/split/ble/peripheral.rs#L144

is the place.

btw could you please help me to fix the function name? Now it serves for all BLE chips, not only nrf.

HaoboGu:
> This is weird, that test is passing for me.

It's an existing bug when concurrently running unittests. Rerunning the ci makes it green again.

rmk/src/ble/mod.rs
```diff
@@ -508,15 +606,30 @@ async fn gatt_events_task(server: &Server<'_>, conn: &GattConnection<'_, '_, Def
...
+    let mut passkey_state = PasskeyInputState::new();
+
     loop {
-        match conn.next().await {
+        #[cfg(feature = "passkey_entry")]
+        let Some(event) = next_gatt_event(conn, &mut passkey_state).await else {
+            continue;
+        };
+        #[cfg(not(feature = "passkey_entry"))]
+        let Some(event) = next_gatt_event(conn).await else {
```
HaoboGu: When `passkey_entry` is not enabled, just use `let event = conn.next().await;`

rmk/src/ble/mod.rs
```diff
@@ -936,6 +1019,9 @@ async fn run_ble_keyboard<
     #[cfg(feature = "host")] rmk_config: &'d mut RmkConfig<'static>,
     #[cfg(feature = "storage")] storage: &mut Storage<F, ROW, COL, NUM_LAYER, NUM_ENCODER>,
 ) {
+    #[cfg(feature = "passkey_entry")]
+    let _passkey_cleanup = crate::ble::passkey::PasskeyCleanupGuard::new();
```
HaoboGu: There's already a guard in `PasskeyInputState`, why there's an additional guard?

rmk-config/src/resolved/build_constants.rs
```diff
@@ -79,12 +79,7 @@ impl crate::KeyboardTomlConfig {
             action,
         );
 
-        // Only validate passkey settings when the build will emit passkey constants.
-        let passkey = if std::env::var("CARGO_FEATURE_PASSKEY_ENTRY").is_ok() {
```
HaoboGu: Removing the `CARGO_FEATURE_PASSKEY_ENTRY` guard means users with invalid `passkey_entry_timeout` in `keyboard.toml` will now get build errors even without the feature enabled.

alxzh: Yep, reverted this.

examples/use_rust/esp32c3_ble/src/main.rs
```diff
@@ -49,6 +49,8 @@ async fn main(_s: Spawner) {
     let central_addr = [0x18, 0xe2, 0x21, 0x80, 0xc0, 0xc7];
     let mut host_resources = HostResources::new();
     let stack = build_ble_stack(controller, central_addr, &mut rng, &mut host_resources).await;
+    #[cfg(feature = "passkey_entry")]
```
HaoboGu: This should be removed in examples. The example binary doesn't have feature "passkey_entry". Adding this before the `stack.build()`

alxzh:
All the feedback was addressed. I've been running this version for several weeks without any issues.

HaoboGu (approved):
nice, thanks a lot!

PR #796, opened 2026-04-10
feat: add configurable one-shot modifier behaviors

ldsands:
Again, I'm trying to replicate all the features I was using previously on KMK. I've only got one more after this. I think I tested this fairly extensively on my own keyboard, and almost everything worked perfectly. I add this so that it is much easier to send a bare one-shot key to the OS than it was before. This allows you to double tap a one-shot key to immediately send that one-shot key to the os with no other keys. It also allows for sending the one-shot key once after tapping it when the timeout is reached. This is useful for using the gui (start) key which I couldn't use on its own to get to the start menu (on Windows and similar on Linux) while keeping the one-shot functionality. Thanks again.

Add three new configuration options for one-shot modifiers (OSM) and fix OSM interaction with mouse keys.

### New OSM configuration options

All options default to `false`, preserving existing behavior for current users.

**Re-press behavior** — what happens when you press the same OSM key again while one-shot is active:

- `tap_on_double_press` — send a bare modifier tap to the host and consume the one-shot state. Useful for intentionally firing the modifier by itself (e.g., tapping LGui twice to open the Start menu).
- `retap_cancel` — cancel the one-shot silently without sending anything. Takes priority over `tap_on_double_press` if both are enabled.

**Timeout behavior:**

- `tap_on_timeout` — when the one-shot timeout expires with no follow-up key, send a bare modifier tap instead of silently cancelling. Useful for triggering OS actions tied to a lone modifier press.

```toml
[behavior.one_shot_modifiers]
tap_on_timeout = true
tap_on_double_press = true
```

### Behavior change: unconditional unstick-on-repress removed

The `main` branch (post-v0.8.2) added unconditional logic in `process_action_osm()` where re-pressing the same OSM key always removed the matching modifier bits and sent a report pair. This PR replaces that unconditional behavior with the explicit `tap_on_double_press` and `retap_cancel` options above.

Since the unstick behavior was added after v0.8.2 and has not been in a published release, this should not affect any released users. Users who adopted the behavior from `main` can restore it by adding `tap_on_double_press = true` to their config.

### Bug fixes

**OSM + mouse click:** Mouse HID reports don't carry keyboard modifier bits. When an OSM modifier was active and a mouse button was pressed, the host received the click without modifier context, so combinations like Ctrl+Click didn't work. Fixed by sending a keyboard report with resolved OSM modifiers before the mouse report in `process_action_mouse`.

**Select race in OSM timeout:** Guard against the `select` race where the timer wins even though a key event arrived at the subscriber simultaneously. Now checks `try_next_message_pure()` before treating a timeout as genuine.

**update_osm timing:** Clear `Single` state on any key event (press or release), not just release. Since `resolve_modifiers` runs before `update_osm`, the modifier is already included in the HID report by the time `update_osm` executes, so clearing on press prevents the state from lingering between key press and release.

### Tests

Four new tests covering the configurable behaviors:
- `test_osm_tap_on_timeout` — bare modifier tap on timeout expiry
- `test_osm_tap_on_double_press` — bare modifier tap on re-press
- `test_osm_retap_cancel` — silent cancellation on re-press
- `test_osm_double_press_different_modifier_still_stacks` — stacking different OSM modifiers still works with `tap_on_double_press` enabled

### Context

I'm using a Sofle V2 split keyboard with sticky modifier keys (OSM) and needed these behaviors to match what I had in KMK firmware (`KC.SK` with `retap_cancel=True` and timeout tap). The mouse fix was discovered during daily use — Ctrl+Click for opening links in new tabs wasn't working with OSM-Ctrl.

### Changes

- `rmk/src/keyboard/oneshot.rs` — core OSM state machine changes
- `rmk/src/keyboard.rs` — mouse + OSM modifier fix
- `rmk/src/config/behavior.rs` — new config struct fields
- `rmk-config/src/lib.rs` — TOML deserialization
- `rmk-config/src/resolved/behavior.rs` — resolved config
- `rmk-macro/src/codegen/behavior.rs` — proc macro codegen
- `rmk/tests/keyboard_one_shot_test.rs` — 4 new tests
- `docs/docs/main/docs/configuration/behavior.md` — user-facing docs
- `docs/docs/main/docs/configuration/appendix.md` — full config reference

### Minor Issue with Windows

The one "issue" is something that I'm pretty sure is an OS issue (with Windows gui+arrow keys) anyway. It works normally but when tapping (once) the one-shot gui key then layer key (MO) then tap an arrow key with another tap. It sends the correct combination but it also sends the gui key (seemingly independently). I realized that my KMK firmware also acted weird when doing this as well. This does not occur with any other key as far as I can tell (e.g. I tried using the exact same sequence but with the `a` key and it worked perfectly).

So again the only reason I came across it was due to trying to be comprehensive in my testing so it doesn't effect me but I figured I would be transparent. I tired to fix it for a while but couldn't figure out anything that worked. If anyone wants to take a stab at it feel free.

### DIFF
--- a/rmk/src/keyboard.rs
+++ b/rmk/src/keyboard.rs
@@ -1480,6 +1480,13 @@ impl<'a> Keyboard<'a> {
 
     /// Process mouse key action with acceleration support.
     async fn process_action_mouse(&mut self, key: HidKeyCode, event: KeyboardEvent) {
+        // Mouse reports don't carry keyboard modifiers. When an OSM modifier
+        // is active and a mouse button is pressed, send a keyboard report with
+        // the modifier first so the host applies it to the click (Ctrl+Click).
+        if event.pressed && self.osm_state.value().is_some() {
+            self.send_keyboard_report_with_resolved_modifiers(true).await;
+        }
+
         let action = {
             let config = self.keymap.mouse_key_config();
             self.mouse.process(key, event.pressed, &config)
--- a/docs/docs/main/docs/configuration/appendix.md
+++ b/docs/docs/main/docs/configuration/appendix.md
@@ -123,6 +123,9 @@ one_shot = {
 # One Shot Modifiers configuration
 one_shot_modifiers = {
   activate_on_keypress = false,
+  tap_on_timeout = false,
+  tap_on_double_press = false,
+  retap_cancel = false,
 }
 
 [behavior.morse]
--- a/docs/docs/main/docs/configuration/behavior.md
+++ b/docs/docs/main/docs/configuration/behavior.md
@@ -50,18 +50,41 @@ By default, one-shot modifiers do not activate on keypress and will be sent only
 You can change this behavior by setting `activate_on_keypress` to `true`.
 This behavior is also known as One-Shot Sticky Modifiers (OSSM).
 
-If you press One-Shot Modifier again, it will be sent as a normal modifier key press and, therefore, released.
+### Re-press behavior
+
+When you press the same OSM key again while one-shot is active, the behavior depends on these options:
+
+- `tap_on_double_press` (default: `false`) — send a bare modifier tap to the host and consume the one-shot state. Useful for intentionally firing the modifier by itself (e.g., tapping LGui twice to open the Start menu).
+- `retap_cancel` (default: `false`) — cancel the one-shot silently without sending anything.
+
+If neither is enabled, re-pressing the same OSM key has no effect (the one-shot remains active). If both are enabled, `retap_cancel` takes priority.
+
+Pressing a *different* OSM key while one-shot is active will stack the modifiers (e.g., OSM-Shift then OSM-Ctrl applies both to the next key).
+
+### Timeout behavior
+
+- `tap_on_timeout` (default: `false`) — when the one-shot timeout expires with no follow-up key, send a bare modifier tap to the host instead of silently cancelling. Useful for triggering OS actions tied to a lone modifier press (e.g., tapping LGui to open the Start menu).
 
 Default values:
 ```toml
 [behavior.one_shot_modifiers]
 activate_on_keypress = false
+tap_on_timeout = false
+tap_on_double_press = false
+retap_cancel = false
 ```
 
 OSSM example:
 ```toml
 [behavior.one_shot_modifiers]
 activate_on_keypress = true
+```
+
+Example with re-press and timeout behaviors:
+```toml
+[behavior.one_shot_modifiers]
+tap_on_timeout = true
+tap_on_double_press = true
 ```
 
 ## Combo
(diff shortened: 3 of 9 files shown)


ldsands:
Oops I accidentally pushed my other feature I've been working on into this pull request i'll remove it from here to make it simpler then make that a separate pr

ldsands:
Okay now this is ready for review. Sorry about that branch mix up.

HaoboGu:
When `activate_on_keypress` is enabled, you can tapping the modifier by double-tap the key, why an extra `tap_on_double_press` is needed for this case?

rmk/src/keyboard.rs
```diff
@@ -1480,6 +1480,13 @@ impl<'a> Keyboard<'a> {
 
     /// Process mouse key action with acceleration support.
     async fn process_action_mouse(&mut self, key: HidKeyCode, event: KeyboardEvent) {
+        // Mouse reports don't carry keyboard modifiers. When an OSM modifier
+        // is active and a mouse button is pressed, send a keyboard report with
+        // the modifier first so the host applies it to the click (Ctrl+Click).
+        if event.pressed && self.osm_state.value().is_some() {
```
HaoboGu, line 1486: OSM + mouse click sends the modifier, but the OSM state is not updated after sending it.

ldsands:
> When `activate_on_keypress` is enabled, you can tapping the modifier by double-tap the key, why an extra `tap_on_double_press` is needed for this case?

With activate_on_keypress enabled, double-tapping doesn't actually produce a clean bare modifier tap. No release report is sent between the first release and the second press — the modifier stays ON continuously from the first press. After the second press, the OSM just re-arms in Single state and waits for the timeout before releasing. So the OS sees one long modifier hold, not a tap.

For keys like GUI/Win, the OS distinguishes between a quick tap (opens Start menu) and a long hold (does nothing or opens a different UI). Without tap_on_double_press, the modifier isn't released until the timeout fires, which is too slow to register as a tap.

tap_on_double_press fixes this by immediately sending a release report on the second press, creating a clean press-release sequence.

I'm not opposed to changing the way that activate_on_keypress works but I usually prefer to double tap if I need the start menu (or similar on Linux) rather than waiting for the timeout to trigger the one-shot key. What do you think?

HaoboGu:
IIRC the current behavior is to release the one-shot modifier on the second tap when `activate_on_keypress = true`. So I thought `tap_on_double_press` was covered in current behavior.. Maybe I remember it wrong.

ldsands:
I don't know what I missed, but when I was testing the one-shot behaviors on my keyboard, I could have sworn that neither the double tap nor the send the key on timeout of the one-shot worked. Clearly, I must not have had the behavior enabled or something. I honestly don't think that this branch adds anything that is needed at all. If there is anything useful here, you can let me know and I can pull out just the changes you want.

Sorry to have wasted your time on this.

HaoboGu:
Could you please test that if `activate_on_keypress = true` works for you?

ldsands:
> Could you please test that if `activate_on_keypress = true` works for you?

Yes, I retested that (from current main branch) on my keyboard and it works fine. I still don't know how I came to the conclusion that it wasn't working as I had anticipated. Oh well.

HaoboGu:
Good to hear that! Anyway thanks a lot for your time and effort!

Issue #87, opened 2024-10-06
[Feature Request] SPI 74HC595 shift register as part of matrix scanning


ramonimbao:
ZMK has support for [74HC595 shift registers as parts of its matrix scanning](https://zmk.dev/docs/development/hardware-integration/shift-registers) as of adding additional GPIO. This works really well for small boards with few GPIO available like the [Seeed Studio XIAO NRF52840](https://www.seeedstudio.com/Seeed-XIAO-BLE-nRF52840-p-5201.html). I have boards here I can test for this. I'd like to be able to contribute but I am unfamiliar with embedded Rust, and my Rust skills are definitely rusty 😅

HaoboGu:
I'm happy to see more matrix types are supported :D

You can combine [`embedded_hal::spi`](https://docs.rs/embedded-hal/latest/embedded_hal/spi/index.html) and [`MatrixTrait`](https://github.com/HaoboGu/rmk/blob/main/rmk/src/matrix.rs#L10C1-L15C31) to write a platform diagnostic 74HC595 matrix scanning algorithm. If you have any question about implementing it, just feel free to ask

HaoboGu:
If you knew Rust before, learning embedded stuff is not that hard. I think some basic embedded and spi protocol knowledge is all you need to implement it.

ramonimbao:
Was searching around for crates to reference and I found this crate: [shiftreg-spi](https://crates.io/crates/shiftreg-spi). It seems like exactly what's needed since it implements the `OutputPin` trait. Does this mean we'll be able to mix it in with a normal matrix alongside other microcontroller GPIO pins? That's exactly how some of my boards are set up.

HaoboGu:
> Was searching around for crates to reference and I found this crate: [shiftreg-spi](https://crates.io/crates/shiftreg-spi). It seems like exactly what's needed since it implements the OutputPin trait. Does this mean we'll be able to mix it in with a normal matrix alongside other microcontroller GPIO pins? That's exactly how some of my boards are set up.

Yes, it seems that you even don't need to change any RMK code. If I understand it correctly, you can just put the `OutputPin` list to the matrix, that should work.

mingti:
Any progress on this? I'd love to use this feature in my project if it's already in the main branch.

HaoboGu:
> Any progress on this? I'd love to use this feature in my project if it's already in the main branch.

No built-in matrix for it right now, but iirc in the community there're already some implementations

PR #798, opened 2026-04-11
Add ShiftRegisterMatrix for 74HC595-based matrix scanning

Taro-Hayashi:
## Summary

Add `ShiftRegisterMatrix`, a 74HC595-based matrix implementation for RMK.

## What this adds

- a new `rmk::shift_register::ShiftRegisterMatrix`
- SPI-driven column scanning using one or more daisy-chained 74HC595 shift registers
- regular GPIO row inputs
- compatibility with RMK's existing matrix/debounce flow

## Why

This implementation allows matrix columns to be driven through a 74HC595 chain while keeping rows as normal input pins, making it practical to build larger matrices on MCUs with tight pin budgets.
This may also be relevant to #87.

## Notes

- one 74HC595 provides 8 column outputs
- multiple 74HC595s can be daisy-chained for wider matrices
- the implementation includes small settle delays around column switching to improve scan stability on real hardware

## Validation

- `cargo check -p rmk`
- verified on real hardware using a 74HC595-driven matrix


rmk/src/shift_register.rs
```diff
@@ -0,0 +1,291 @@
...
+    pub async fn init(&mut self) {
+        self.clear_columns().await;
+    }
+
+    /// Busy-wait delay (~30µs) for SPI signal settling.
+    ///
+    /// Matches QMK's `matrix_io_delay()` to allow the 595 outputs to
+    /// propagate through the key matrix before reading row pins.
+    #[inline(always)]
+    fn io_delay() {
```
HaoboGu: Remove it. Use`embassy_time::Timer::after_micros(30).await` instead. All delays in RMK should be async

rmk/src/matrix/hc595_matrix.rs
```diff
@@ -0,0 +1,291 @@
...
+use embedded_hal_async::digital::Wait;
+use embedded_hal_async::spi::SpiBus;
+
+use crate::debounce::{DebounceState, DebouncerTrait};
+use crate::event::{KeyboardEvent, publish_event_async};
+use crate::input_device::{InputDevice, Runnable};
+use crate::matrix::{KeyState, MatrixTrait};
+
+/// Maximum number of bytes in the shift register chain (supports up to 32 columns).
+const SR_MAX_BYTES: usize = 4;
```
HaoboGu, line 45: SR_MAX_BYTES = 4 caps the chain at 32 columns, but nothing enforces it. 

Add something like `const { assert!(COL <= SR_MAX_BYTES * 8, "…")`  in `new()` or `NUM_BYTES` calculation

rmk/src/lib.rs
```diff
@@ -106,6 +106,7 @@ pub mod light;
 pub mod matrix;
 pub mod morse;
 pub mod processor;
+pub mod shift_register;
```
HaoboGu: Move this file under `matrix` mod.

rmk/src/shift_register.rs
```diff
@@ -0,0 +1,291 @@
...
+            key_states: [[KeyState::new(); ROW]; COL],
+            scan_pos: (0, 0),
+            #[cfg(feature = "async_matrix")]
+            rescan_needed: false,
+        }
+    }
+
+    /// Initialize the shift register by clearing all outputs.
+    /// Call this once before starting the scan loop.
+    pub async fn init(&mut self) {
```
HaoboGu: `init` is unnecessary. it's just one-liner and can be moved to `new()`

rmk/src/shift_register.rs
```diff
@@ -0,0 +1,291 @@
...
+    SPI: SpiBus,
+    CS: OutputPin,
+    #[cfg(not(feature = "async_matrix"))] In: InputPin,
+    #[cfg(feature = "async_matrix")] In: Wait + InputPin,
+    D: DebouncerTrait<ROW, COL>,
+    const ROW: usize,
+    const COL: usize,
+    const ROW_OFFSET: usize,
+    const COL_OFFSET: usize,
+> Runnable for ShiftRegisterMatrix<SPI, CS, In, D, ROW, COL, ROW_OFFSET, COL_OFFSET>
```
HaoboGu: use `[input_device]` macro

rmk/src/shift_register.rs
```diff
@@ -0,0 +1,291 @@
...
+    D: DebouncerTrait<ROW, COL>,
+    const ROW: usize,
+    const COL: usize,
+    const ROW_OFFSET: usize,
+    const COL_OFFSET: usize,
+> MatrixTrait<ROW, COL> for ShiftRegisterMatrix<SPI, CS, In, D, ROW, COL, ROW_OFFSET, COL_OFFSET>
+{
+    #[cfg(feature = "async_matrix")]
+    async fn wait_for_key(&mut self) {
+        // A shift-register matrix does not have a generally safe passive
```
HaoboGu: Why? I think you can:

  1. Shift out all-0xFF bytes and latch — this drives all columns high simultaneously
  2. Call Wait::wait_for_high() on the row input pins (exactly like Matrix does)
  3. On wake, clear the outputs and start a normal scan

I don't see a reason that this doesn't work

Taro-Hayashi:
Thanks for the review. I updated the PR and incorporated the structural suggestions (moved it under matrix, switched to #[input_device], removed init(), and added the compile-time column limit check).

For the timing-related parts, I also tested on a real board where the shift-register matrix and PMW3610 share the same SPI lines. In that setup, making the latch sequence fully interruptible and using the all-high idle wait caused regressions, so I kept the latch section non-interruptible and used short polling while idle. Shared-SPI designs are probably not the common case, so I do not want to overstate this, but they do exist in practice, so I was a bit conservative there.

rmk/src/shift_register.rs
Taro-Hayashi: I tested the fully-async latch path on a real shared-SPI board, and that caused regressions. In particular, yielding during the latch/write/latch sequence made the shift-register transaction interruptible, which was not stable with another SPI device on the same lines. Because of that, I kept this section non-interruptible for now.

rmk/src/matrix/hc595_matrix.rs
Taro-Hayashi: Added a compile-time check so configurations above 32 columns fail during compilation.

rmk/src/lib.rs
Taro-Hayashi: Done. The implementation now lives under matrix as shift_register_matrix.

rmk/src/shift_register.rs
Taro-Hayashi: Done. init() is removed, and construction is now handled by async new() so the outputs are cleared during initialization.

rmk/src/shift_register.rs
Taro-Hayashi: Done. I switched this to #[input_device(publish = KeyboardEvent)] and removed the manual InputDevice/Runnable implementation.

rmk/src/shift_register.rs
Taro-Hayashi: I tried the all-high idle-wait approach as suggested, but on the shared-SPI test board it caused idle-side regressions in pointing behavior. Since the normal scan path remained stable while short polling also remained stable, I reverted this part back to the conservative polling-based idle path.

HaoboGu:
> For the timing-related parts, I also tested on a real board where the shift-register matrix and PMW3610 share the same SPI lines. In that setup, making the latch sequence fully interruptible and using the all-high idle wait caused regressions, so I kept the latch section non-interruptible and used short polling while idle. Shared-SPI designs are probably not the common case, so I do not want to overstate this, but they do exist in practice, so I was a bit conservative there.

How do you share the SPI? Is there any code public that I can take a look?

Taro-Hayashi:
The keyboard firmware code is here:
https://github.com/Taro-Hayashi/rmk-cannonball-ll

The RMK branch I used for this test is here:
https://github.com/Taro-Hayashi/rmk/tree/feat/74hc595-matrix

The schematic PDF is here:
https://drive.google.com/file/d/1t-5icDK8QyWdWlFEOCBERznMTQsj-sAU/view?usp=sharing

In this design, the shift-register matrix and the PMW3610 share the same SPI lines, with separate control lines for each device. The shared-SPI wrapper I used is in the RMK branch above.

I thought it might be worth mentioning because supporting this case could make the implementation a bit more flexible. But for my own keyboard, it is also fine if I just keep a local version working, so if this does not fit the direction you want for RMK, I am happy to follow your preferred approach.

HaoboGu:
🤔 I think you're using it wrong, in embedded-hal there're `SpiDevice` and `SpiBus`: https://docs.rs/embedded-hal/latest/embedded_hal/spi/index.html

And for the async version is in embassy: https://docs.embassy.dev/embassy-embedded-hal/git/default/shared_bus/asynch/spi/struct.SpiDevice.html

With the async version you can share the SPI and use it across the devices with keeping async feature. RMK depends on embassy, so just use the async version is fine.

Taro-Hayashi:
I reworked the implementation to use `SpiDevice`, and also renamed the module/type from the more generic `shift_register` to `hc595_matrix` / `Hc595Matrix`, since the current implementation is really specific to 74HC595-style usage rather than a generic shift-register abstraction.

Main changes in this update:
- switched the matrix-side SPI bound from `SpiBus` to `embedded-hal-async` `SpiDevice`
- renamed the module/type to `hc595_matrix` / `Hc595Matrix`
- kept the implementation focused on the 74HC595 case

The previously resolved items are preserved:
- `#[input_device(publish = KeyboardEvent)]` macro is still used (no manual `InputDevice` / `Runnable` impls)
- no separate `init()`; `new()` is `async` and clears the outputs during construction
- compile-time check for `COL <= 32` still in place

Tested on real hardware and the matrix scan works as expected.
I hope this is closer to the intended direction.

HaoboGu:
Could you share the SPI across the 74HC595 and PMW3610 now? And can

```rust
    #[inline(always)]
    fn latch_io_delay() {
        for _ in 0..960 {
            core::hint::spin_loop();
        }
    }
```
be removed and use `embassy_time::Timer::after_micros()` instead?

Taro-Hayashi:
Thanks for the review. The delays turned out to be unnecessary, so I removed them.

As for sharing the SPI between the 74HC595 and the PMW3610: I tried wrapping the 74HC595 side as an `SpiDevice` on a shared bus, but the matrix showed ghosting. Rather than expand the scope of this PR, I've decided to keep the shared-SPI variant as a custom implementation in my own repo.

HaoboGu:
Interesting, I thought switching to `SpiDevice` + embedded-hal's `SpiBus` might solve the SPI sharing issue, I don't know why it doesn't work. Anyway, the current implementation looks good, let's get it merged first. Thanks for your work!

PR #803, opened 2026-04-16
fix persistent split one-message-behind-state

### LINKED ISSUE #801: serial rx gets into persistent one-message-behind state
https://github.com/HaoboGu/rmk/blob/2aecdc91c9d04cefa6936c0aff1323adb6682eca/rmk/src/split/serial/mod.rs#L52-L69

Imagine that the remote has sent 2 full messages, and the local call to `self.serial.read` fits them both within `self.buffer`.  (Note: I'm not sure this can actually happen with the checked-in code today because the buffer is only `SPLIT_MESSAGE_MAX_SIZE` bytes; it depends on if more than 1 of the *minimum* message size can fit within that. I tried increasing the buffer size while diagnosing something else and may made a latent problem real in the process.)

Now `<SerialSplitDriver<S> as SplitReader>::read` will take the first one and return, leaving the second in the buffer.

But...the next call to `<SerialSplitDriver<S> as SplitReader>::read` won't actually check if there's already a complete message in the buffer before calling `read`. Thus, this message is delayed until the remote sends something more. And once it does, that call to `<SerialSplitDriver<S> as SplitReader>::read` will read one message, but again there's more data waiting. That delay will persist until it happens to exceed the buffer size (so the "there's already a full message in the buffer" condition is resolved) or the central is power cycled.

I think the fix is just to move the `if self.buffer[..self.n_bytes_part].contains(&SENTINEL)` into the `while` condition, rather than after the `read`.

Not sure what existing test harness there is for this file if any.

scottlamb:
Fixes #801

### DIFF
--- a/rmk/src/split/serial/mod.rs
+++ b/rmk/src/split/serial/mod.rs
@@ -30,13 +30,13 @@ pub(crate) async fn run_serial_peripheral_manager<
 }
 
 /// Serial driver for BOTH split central and peripheral
-pub(crate) struct SerialSplitDriver<S: Read + Write> {
+pub(crate) struct SerialSplitDriver<S> {
     serial: S,
     buffer: [u8; SPLIT_MESSAGE_MAX_SIZE],
     n_bytes_part: usize,
 }
 
-impl<S: Read + Write> SerialSplitDriver<S> {
+impl<S> SerialSplitDriver<S> {
     pub(crate) fn new(serial: S) -> Self {
         Self {
             serial,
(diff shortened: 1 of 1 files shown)


scottlamb:
> note: rustc unexpectedly overflowed its stack! this is a bug

Hmm, I wonder if this is a regression in rustc 1.95.0, released today.

HaoboGu:
I've updated main, skipped the failed examples

HaoboGu (approved):
lgtm, thanks!

PR #804, opened 2026-04-16
rust 1.95 clippy fixes

scottlamb:
Note the `src/host/via/mod.rs` is a behavior change: the `#[cfg(not(feature = "vial_lock"))]` path was within a `#[cfg(feature = "vial_lock"))]` path before.


HaoboGu:
Thanks!

PR #805, opened 2026-04-16
improve log readability

scottlamb:
Before:

```
Process Single key action: Key(Hid(J)), KeyboardEvent { pressed: true, pos: Key(KeyPos { row: 7, col: 4 }) }
Last key code changed from Hid(J) to Hid(J)(pressed: true)
Sending keyboard report, pressed: true
```

After:

```
[   33410ms DEBUG] Process Single key action: Key(Hid(J)), KeyboardEvent { pressed: true, pos: Key(KeyPos { row: 7, col: 4 }) }
[   33411ms INFO ] Sending keyboard report, modifiers: [RGui], keycodes: [J]
```

### DIFF
--- a/rmk/src/keyboard.rs
+++ b/rmk/src/keyboard.rs
@@ -1419,7 +1419,7 @@ impl<'a> Keyboard<'a> {
             }
 
             // Update last key code
-            if hid_keycode != HidKeyCode::Again {
+            if hid_keycode != HidKeyCode::Again && self.last_key_code != key {
                 debug!(
                     "Last key code changed from {:?} to {:?}(pressed: {:?})",
                     self.last_key_code, key, event.pressed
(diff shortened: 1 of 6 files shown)


rmk/src/keyboard.rs
```diff
@@ -163,6 +163,13 @@ impl Runnable for Keyboard<'_> {
     }
 }
 
+struct HeldKeyCodesDbg<'a>(&'a [HidKeyCode]);
```
HaoboGu: I don't like this, adding an extra struct just for debugging is weird

scottlamb: Would you prefer to not filter (logging `[J, No, No, No, No, No]` instead of `[J]`) or find another way to implement it?

If the latter...

Is it specifically that it's a throw-away reference used just for the duration of one formatting? If so, would you prefer `struct Keyboard { ..., held_keycodes: HeldKeyCodes, ... }` that owns the `[HidKeyCode; 6]` as well as having the debug impls?

I saw recently there's`std::fmt::from_fn` for this kind of one-off, but that doesn't seem to be available in `core`, much less for `defmt`.

There's building the output within a `if log::log_enabled!(Level::Debug)`, but that requires building an intermediate string and wouldn't play well with `defmt`.

I'm open to other ideas.

HaoboGu: > Would you prefer to not filter (logging [J, No, No, No, No, No] instead of [J]) or find another way to implement it?

Yes, I prefer to log just `[J, No, No, No, No, No]`. I don't want to introduce any extra complexity just for logging a specific item.

scottlamb: Removed `HeldKeyCodesDbg`. Does that apply to the custom `ModifierCombination` debug impl too?

HaoboGu: Keeping it is fine

HaoboGu (approved):
lgtm now, thanks!

PR #806, opened 2026-04-17
test ux improvements

scottlamb:
It looks like standard `cargo test` fails since ab3ad91c with some failures and timeouts. Improve the experience of landing in the repo and running tests:

* make `cargo test` fail-fast with instructions to use `cargo nextest` instead
* update `CLAUDE.md` instructions to use `cargo nextest`
* add symlinks so you don't need to specify the path to the nextest config in the `rmk`, `rmk-types`, and `rmk-macros` crates

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -20,15 +20,17 @@ The `scripts` directory contains useful scripts for checking/formatting code
 ## Commands
 
 ### Testing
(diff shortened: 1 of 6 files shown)


HaoboGu (approved):
Good catch, thanks a lot!

PR #808, opened 2026-04-17
fix(core): only record press timestamp for morse keys

scottlamb:
`Keyboard::process_inner` wrote every press into the bounded timer LinearMap (HOLD_BUFFER_SIZE = 16), but only the morse-press path consumes that value. Non-morse non-combo presses filled slots no one reads, so after 16 distinct positions the map saturated and `set_timer` began logging "Timer buffer full, dropping timer for ...".

Gate the write on `key_action.is_morse()`. The combo-press path in `process_combo` still reads the timer and now falls back to `Instant::now()` for non-morse inputs, which is indistinguishable within the same synchronous chain.

### DIFF
--- a/rmk/src/keyboard.rs
+++ b/rmk/src/keyboard.rs
@@ -361,27 +361,32 @@ impl<'a> Keyboard<'a> {
         #[cfg(feature = "host_security")]
         self.keymap.update_matrix_state(&event);
 
-        // Matrix should process key pressed event first, record the timestamp of key changes
-        if event.pressed {
-            self.set_timer_value(event, Some(Instant::now()));
-        }
         // Update activity time for BLE split central sleep management
         #[cfg(all(feature = "split", feature = "_ble"))]
         update_activity_time();
+
+        // Capture the event time once per event and thread it through.
+        let event_time = Instant::now();
 
         // Process key
         let key_action = &self.keymap.get_action_with_layer_cache(event);
 
         if self.combo_on {
-            if let (Some(key_action), is_combo) = self.process_combo(key_action, event).await {
-                self.process_key_action(&key_action, event, is_combo).await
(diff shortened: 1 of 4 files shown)


rmk/src/keyboard.rs
```diff
@@ -361,17 +361,22 @@ impl<'a> Keyboard<'a> {
...
         #[cfg(all(feature = "split", feature = "_ble"))]
         update_activity_time();
 
         // Process key
         let key_action = &self.keymap.get_action_with_layer_cache(event);
 
+        // Record press timestamp only for morse keys — they're the only consumers
+        // via `get_timer_value` in keyboard/morse.rs. The combo-press path also reads
+        // it but falls back to `Instant::now()`, which within the same synchronous
+        // chain is indistinguishable. Gating here keeps the bounded timer map from
```
HaoboGu: If the combo-press path also reads it but falls back to `Instant::now()`, change the combo path to use `Instant::now()`?

scottlamb: Huh. Please check me on this, as you obviously know this code much better than I do, but it looks like we can go even further and get rid of the map entirely. Even for the morse path, it looks like this was only used for the current key within `process_inner`. 

Between, press and release, it's stored in `HeldKey` instead.

HaoboGu: It looks correct, but the added `press_time` param of `process_key_action`, `process_key_action_morse`, etc. is kind of confusing, because the key is not always "pressed", and it's not always the "time of the key is pressed". It feels more like the `event_time`.

scottlamb: Good point, `event_time` it is.

scottlamb:
...and rebased to fix the changelog conflict.

PR #810, opened 2026-04-19
Fix macro keycode expansion after type change

smjothen:
Fix macro code generation in `rmk-macro` to use `HidKeyCode` instead of the old `KeyCode` path.


HaoboGu:
thanks!

PR #813, opened 2026-04-19
feat(state): expand ConnectionState from bool to u8 enum to support USB suspend

fuchskurt:
Tracking suspend as a distinct connection state preserves the Disconnected->Connected wakeup path and unlocks suspend-aware firmware features, such as turning off the backlight when the host suspends the bus.

Replace AtomicBool CONNECTION_STATE with AtomicU8 and add a Suspended variant to ConnectionState:

-   0 = Disconnected
-   1 = Connected
-   2 = Suspended


HaoboGu:
I think `ConnectionState` needs to be rewritten like `BleStatus`, for example, make the `ConnectionState` be aware of `ConnectionType`. Just adding a `Suspended` state just a temporary solution. BLE doesn't have `Suspended` state.

fuchskurt:
I agree. I created a fork to address this, since the keyboard LED backlight staying on when the host is off was driving me crazy. This was essentially a quick 'what can I do with minimal impact on the codebase' fix, and I figured it was worth creating a PR since it doesn't break anything.

That said, I think your suggestion is the right approach. If this is a desired feature, I'm happy to work on it, though the earliest I'd have time is next week. If that works, just mark this PR as draft or close it and I'll take a look.

HaoboGu:
Yes that would be great! Thanks in advance

HaoboGu:
I'm going to do a refactor about USB/BLE transport, I'll do the ConnectionStatus refactor. Just fyi in case that you're doing something similar

PR #814, opened 2026-04-23
feat: add `zsa_voyager_bl` for ZSA Voyager DFU

jpds:
I have finished enabling RMK for the ZSA Voyager. This is the only patch it required to enable the `kbctrl!(Bootloader)` key.

Once merged, I will publish the rest of the supporting code in a separate keyboard repo.

### DIFF
--- a/rmk/Cargo.toml
+++ b/rmk/Cargo.toml
@@ -201,6 +201,9 @@ rp2040 = [
 ## Enable feature if you're using Adafruit nRF52 bootloader and want bootloader jumping key
 adafruit_bl = ["_nrf_ble"]
 
+## Enable feature if you're using the ZSA Voyager ignition DFU bootloader
+zsa_voyager_bl = []
+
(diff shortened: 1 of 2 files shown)


HaoboGu:
Thanks for the PR!

`https://github.com/qmk/qmk_firmware/blob/master/keyboards/zsa/voyager/voyager.c` is licensed under GPL-2.0. I am concerned that this might not be compatible with RMK's licensing.

jpds:
@HaoboGu The only thing taken from voyager.c is from `bootloader_jump` - the knowledge of which pins to toggle and the timing, which is the hardware behavior dictated by ZSA's PCB circuit. This Rust implementation here uses raw MMIO register access (which is completely different from QMK's C GPIO abstraction layer).  I've removed the reference comment to avoid the ambiguity.

The dapboot trick didn't work as it expected the hardware boot key to be pressed for reset.

HaoboGu:
Okay, that's fine, thanks a lot for your effort!

PR #815, opened 2026-04-24
minimal Azoteq iqs5xx (TPS43, TPS65, ...) driver

scottlamb:


scottlamb:
(re-pushed to use the `#[input_device(...)]` macro rather than direct `impl Runnable` to be more iodiomatic)

HaoboGu:
I don't have the hardware on my hand, pls ping me if you think this PR is ready for merging

scottlamb:
I think so—it works reliably for me, and I've built on it locally with follow-ups to add the missing gestures.

I haven't added any user-facing documentation; I can add some here if you'd like or save it until we have full functionality.

HaoboGu:
Yeah, please add the documentation, ty

scottlamb:
Added docs. While doing so I felt silly saying it doesn't support `keyboard.toml`, so I added that too.

HaoboGu (approved):
Great work, thanks!

PR #819, opened 2026-04-25
fix(usb): shrink HID endpoint max_packet_size to fit tight PMA

jpds:
The `add_usb_writer!` and `add_usb_reader_writer!` macros hardcode `max_packet_size` 64 for every HID endpoint. Each endpoint reserves that many bytes per direction in the USB  controller's Packet Memory Area (PMA).

Whilst enabling a stenography HID on the ZSA Voyager (which has a 512-byte PMA), this new HID caused embassy-usb to panic at endpoint allocation - the four HID interfaces  at 64 bytes each exceeded the available packet memory. The PMA budget is tight because EP0 and the endpoint descriptor table already consume a number of bytes before any HID endpoints are allocated. Right-sizing each endpoint to its actual report frees enough PMA for the additional interface.

The current endpoints update like this:

  ```
  ┌───────────────────────┬─────────────┬──────────────────────┬─────────────────┐
  │       Endpoint        │ Report size │ Previous reservation │ New reservation │
  ├───────────────────────┼─────────────┼──────────────────────┼─────────────────┤
  │ Keyboard (IN+OUT)     │ 8 B         │ 128 B                │ 16 B            │
  ├───────────────────────┼─────────────┼──────────────────────┼─────────────────┤
  │ Composite (IN)        │ 16 B        │ 64 B                 │ 16 B            │
  ├───────────────────────┼─────────────┼──────────────────────┼─────────────────┤
  │ Vial (IN+OUT)         │ 32 B        │ 128 B                │ 64 B            │
  └───────────────────────┴─────────────┴──────────────────────┴─────────────────┘
  ```

The future steno HID has a report size of 9B with a 16B reservation.

The existing 3-argument call syntax is preserved with a default of 64, so no downstream breakage.

### DIFF
--- a/rmk/src/ble/mod.rs
+++ b/rmk/src/ble/mod.rs
@@ -232,14 +232,14 @@ pub(crate) async fn run_ble<
     #[cfg(not(feature = "_no_usb"))]
     let (mut _usb_builder, mut keyboard_reader, mut keyboard_writer, mut other_writer) = {
         let mut usb_builder: embassy_usb::Builder<'_, D> = new_usb_builder(usb_driver, rmk_config.device_config);
-        let keyboard_reader_writer = add_usb_reader_writer!(&mut usb_builder, KeyboardReport, 1, 8);
-        let other_writer = add_usb_writer!(&mut usb_builder, CompositeReport, 9);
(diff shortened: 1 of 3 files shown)


HaoboGu:
Nice work, thanks a lot!

PR #825, opened 2026-04-28
fix(usb): complete remote wakeup for non-OTG peripherals

jpds:
USB remote wakeup doesn't work on non-OTG STM32/GD32 parts:

Two gaps in the current implementation:

1. The USB-only suspend loop in lib.rs never calls usb_device.remote_wakeup(). Only the BLE+USB path in ble/mod.rs had the call.
2. The HID writer only signals USB_REMOTE_WAKEUP after getting EndpointError::Disabled from a failed write. On non-OTG peripherals, endpoints stay NAK/VALID during suspend, so that error never fires and the wakeup signal never gets sent.

Fix:

- Track bus suspend state via a USB_SUSPENDED atomic, updated from the Handler::suspended() callback.
- When a report is pending and the bus is suspended, signal USB_REMOTE_WAKEUP before attempting the write. This wakes the host proactively.
- Call usb_device.remote_wakeup() in the USB-only suspend loop. On failure, fall back to wait_resume() (host-initiated resume still works).
- Keep the EndpointError::Disabled retry path for OTG peripherals where Disabled is the correct suspend indicator.

Testing: Tested on a ZSA Voyager (GD32F303). Keyboard reliably wakes the host on keypress after suspend. The companion embassy-stm32 driver fix (non-OTG Bus::remote_wakeup()) is at: https://github.com/embassy-rs/embassy/pull/5978

### DIFF
--- a/rmk/src/ble/mod.rs
+++ b/rmk/src/ble/mod.rs
@@ -306,10 +306,11 @@ pub(crate) async fn run_ble<
             match select(usb_device.wait_resume(), USB_REMOTE_WAKEUP.wait()).await {
                 Either::First(_) => continue,
                 Either::Second(_) => {
-                    info!("USB wakeup remote");
-                    if let Err(e) = usb_device.remote_wakeup().await {
-                        info!("USB wakeup remote error: {:?}", e)
+                    info!("USB remote wakeup requested");
+                    if usb_device.remote_wakeup().await.is_ok() {
+                        continue;
                     }
+                    usb_device.wait_resume().await;
                 }
             }
         }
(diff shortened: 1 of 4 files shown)


rmk/src/ble/mod.rs
```diff
@@ -306,9 +306,12 @@ pub(crate) async fn run_ble<
             match select(usb_device.wait_resume(), USB_REMOTE_WAKEUP.wait()).await {
                 Either::First(_) => continue,
                 Either::Second(_) => {
-                    info!("USB wakeup remote");
-                    if let Err(e) = usb_device.remote_wakeup().await {
-                        info!("USB wakeup remote error: {:?}", e)
+                    info!("USB remote wakeup requested");
+                    match usb_device.remote_wakeup().await {
```
HaoboGu: I think it's better to have:

```rust
if let Ok(_) = usb_device.remote_wakeup().await {
  continue;
}
// Err branch
usb_device.wait_resume().await;
```

for smaller binary size

jpds: Sure, done.

HaoboGu:
Thanks a lot!

PR #827, opened 2026-04-29
feat(usb): advertise boot keyboard protocol on primary HID interface

jpds:
## Problem

Linux requires `power/wakeup=enabled` in sysfs for USB remote wakeup to work. RMK keyboards get `power/wakeup=disabled` by default, requiring a per-VID udev rule. QMK keyboards get `enabled` automatically with no udev rule.

## Root cause

The Linux `usbhid` driver auto-enables wakeup when it probes a HID interface with `bInterfaceSubClass=1` (Boot) and `bInterfaceProtocol=1` (Keyboard):

```c
// drivers/hid/usbhid/hid-core.c:1205
if (interface->desc.bInterfaceSubClass == USB_INTERFACE_SUBCLASS_BOOT &&
    interface->desc.bInterfaceProtocol == USB_INTERFACE_PROTOCOL_KEYBOARD)
    device_set_wakeup_enable(&dev->dev, 1);
```

QMK sets these fields. RMK (embassy-usb) leaves them at 0/0.

## Fix

Pass `HidSubclass::Boot` and `HidBootProtocol::Keyboard` when constructing the primary keyboard HID interface, in both the USB-only and BLE+USB paths.

Extends `add_usb_reader_writer!` with optional subclass/protocol parameters. The default (No/None) is unchanged for non-keyboard interfaces like the composite report and steno endpoints.

## Testing

Confirmed with `lsusb -v` that the keyboard interface now shows `bInterfaceSubClass=1, bInterfaceProtocol=1`, and `cat /sys/bus/usb/devices/.../power/wakeup` reads `enabled` on plug-in without any udev rule. Tested on kernel 6.18.

### DIFF
--- a/rmk/src/ble/mod.rs
+++ b/rmk/src/ble/mod.rs
@@ -215,7 +215,15 @@ pub(crate) async fn run_ble<
     #[cfg(not(feature = "_no_usb"))]
     let (mut _usb_builder, mut keyboard_reader, mut keyboard_writer, mut other_writer) = {
         let mut usb_builder: embassy_usb::Builder<'_, D> = new_usb_builder(usb_driver, rmk_config.device_config);
-        let keyboard_reader_writer = add_usb_reader_writer!(&mut usb_builder, KeyboardReport, 1, 8, 8);
+        let keyboard_reader_writer = add_usb_reader_writer!(
+            &mut usb_builder,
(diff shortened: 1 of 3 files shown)


HaoboGu:
Good catch, thanks!

PR #831, opened 2026-05-04
hid: fix system control array descriptor so system sleep key works

jpds:
I noticed that pressing the system sleep key on my keymap did nothing. The HID system control descriptor had usage_min = 0x81 with logical_min = 1, which caused the host to resolve the sleep usage ID to an out-of-range value and discard it.

Fix is a one-byte descriptor change: usage_min = 0x01. No behaviour change for any other report type.

Tested with a Linux desktop.


HaoboGu (approved):
Good catch. Thanks!

PR #833, opened 2026-05-05
watchdog: Add runner

jpds:
Adds a watchdog runner to the supported chipsets. This enables the hardware watchdog to reset the board automatically if Embassy on the device locks up.

I do not have hardware to test these other chipsets, but I used this as the basis for the watchdog on my ZSA Voyager.

Timeout it set to `8s` on RP2040 as that's a hardware limitation, the others run this at `10s`.

### DIFF
--- a/rmk-macro/src/codegen/watchdog.rs
+++ b/rmk-macro/src/codegen/watchdog.rs
@@ -14,7 +14,12 @@ use rmk_config::resolved::hardware::ChipSeries;
 #[cfg(feature = "watchdog")]
 pub(crate) fn expand_watchdog_init(hardware: &Hardware) -> (TokenStream2, Option<TokenStream2>) {
     let init = match hardware.chip.series {
-        ChipSeries::Stm32 | ChipSeries::Rp2040 | ChipSeries::Nrf52 | ChipSeries::Esp32 => quote! {},
+        ChipSeries::Rp2040 => quote! {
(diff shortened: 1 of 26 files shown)


HaoboGu:
It seems that there're some issues/warning reported, so the CI fails. Could you fix it?

jpds:
@HaoboGu Fixed now, and enhanced the CI so it didn't hide the error.

HaoboGu:
Can you make watchdog optional?

jpds:
Should be all good now.

rmk-macro/src/codegen/watchdog.rs
```diff
@@ -0,0 +1,53 @@
...
+            let mut watchdog_runner = ::rmk::watchdog::Rp2040Watchdog::default_runner(
+                ::embassy_rp::watchdog::Watchdog::new(p.WATCHDOG),
+            );
+        },
+        ChipSeries::Nrf52 => quote! {
+            let mut watchdog_runner =
+                ::rmk::watchdog::Nrf52Watchdog::default_runner(p.WDT);
+        },
+        ChipSeries::Esp32 => quote! {
+            let mut esp_wdt_inner = timg0.wdt;
```
HaoboGu: timg0 is not imported?

jpds: Great catch - adjusted.

HaoboGu:
Nice work, thanks a lot!

PR #834, opened 2026-05-06
feat: add combo.prior_idle_time cooldown window

yekingyan:
Added a configurable cooldown window (require_prior_idle_ms) that prevents combo recording when a key is pressed within the idle period after the previous key. This is equivalent to ZMK's per-combo require-prior-idle-ms, but implemented as a global setting under [behavior.combo].

Tested on nrf52840_split with require_prior_idle_ms = 150 — fast typing no longer triggers combo false-positives, and keystrokes are never swallowed.

When a key press falls within the idle window, combo update is skipped while the key still dispatches normally — fast typists won't lose keystrokes, only combo false-positives are prevented.

Configuration example:
  [behavior.combo]
  require_prior_idle_ms = 150  # 150ms cooldown

Default is no cooldown (backward compatible).

### DIFF
--- a/rmk-config/src/lib.rs
+++ b/rmk-config/src/lib.rs
@@ -629,6 +629,7 @@ pub struct OneShotModifiersConfig {
 pub(crate) struct CombosConfig {
     pub combos: Vec<ComboConfig>,
     pub timeout: Option<DurationMillis>,
+    pub require_prior_idle_ms: Option<u64>,
 }
 
 /// Configurations for combo
--- a/rmk-config/src/lib.rs
+++ b/rmk-config/src/lib.rs
@@ -629,7 +629,7 @@ pub struct OneShotModifiersConfig {
 pub(crate) struct CombosConfig {
     pub combos: Vec<ComboConfig>,
     pub timeout: Option<DurationMillis>,
-    pub require_prior_idle_ms: Option<u64>,
+    pub prior_idle_time: Option<DurationMillis>,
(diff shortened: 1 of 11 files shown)


HaoboGu:
Did you try `prior_idle_time`?

If `prior_idle_time` doesn't work for combo, please rename `combo.require_prior_idle_ms` to `combo.prior_idle_time`. Or just use `prior_idle_time` in combo as well? Removing an extra param is great if possible.

yekingyan:
> Did you try `prior_idle_time`?
> 
> If `prior_idle_time` doesn't work for combo, please rename `combo.require_prior_idle_ms` to `combo.prior_idle_time`. Or just use `prior_idle_time` in combo as well? Removing an extra param is great if possible.

Thanks for the suggestion! I checked the existing prior_idle_time and confirmed in the code — it's only used in the morse flow tap path (keyboard.rs:648), gated by enable_flow_tap and key_action.is_morse(). It detects fast typing to resolve morse keys as tap immediately, which is the opposite direction of what combo needs (suppress  combo during fast typing). I also tested it on hardware and confirmed it has no effect on combo behavior.
So reusing the same param isn't feasible, but I've renamed combo.require_prior_idle_ms to combo.prior_idle_time  to keep naming consistent.

HaoboGu:
Thanks!

HaoboGu:
And it would be great if you can add some document.

PR #837, opened 2026-05-09
keep Cargo.toml and Cargo.lock in sync

scottlamb:
My language server keeps running `cargo check`-like commands and causing whatever lockfiles it hits to be updated. So it's hard to prepare clean changes when starting from out-of-sync lockfiles.

So add a script that updates them all (conveniently, `cargo fetch` doesn't even need the target toolchain) and enforces it on CI.

Most (possibly all) of the lockfile updates in this change are to match `a885a16f`.


HaoboGu:
Great, thanks!

PR #838, opened 2026-05-10
fix delayed modifier oled updates

scottlamb:
The symptom I would see is that when I hit shift, there was no change on press, but the icon was underlined on release. The next keypress would release the underline.


HaoboGu:
oh good catch, thanks!

PR #840, opened 2026-05-10
rework RP2040 PIO UART interrupt path

scottlamb:
The TX path was broken if `PIO::uart_buffer().buf_tx.is_full()` was true on entry; it would not drain or disable the interrupt, so the it'd repeat forever, starving other work. Best guess this was supposed to be `is_empty()`, but the check is unnecessary anyway.

I accidentally reproduced this quite reliably by sending several maximum-length messages in a row. :-/

While here, I adjusted both TX and RX loops to transfer up to 8 bytes per availability check rather than 1 by using the FLEVEL register instead of FSTAT.


HaoboGu (approved):
lgtm, thanks a lot!

PR #841, opened 2026-05-10
fix(config): serde default for CombosConfig.combos

jpds:
I don't have combos defined in my `keyboard.toml`, so without this, a `[behavior.combo]` section that only sets timeout or prior_idle_time (no combos defined) fails to deserialise with "missing configuration field behavior.combo.combos".


HaoboGu:
Thanks!

PR #842, opened 2026-05-10
fix(morse): resolve same-hand HRM tap on press in Normal mode

jpds:
I noticed that when I very quickly rolled a home-row key with another key fast (such as: a+g), what I would get would be "ga" as output.

That patch solves that issue by ensuring the taps fire in the correct order.


HaoboGu:
lgtm, thanks!

PR #845, opened 2026-05-12
fix(morse): drop stale held-buffer entry before flow-tap re-insert

jpds:
I noticed that my flow-tap backspace key would occasionally jam and just start deleting text.

Traced it down to this stale entry.


HaoboGu (approved):
Great, thanks!

PR #846, opened 2026-05-12
fix(morse): keep a continuation breadcrumb after a flow-tapped tap

jpds:
Stacked on #845.

A tap that gets resolved by flow-tap (rather than by early-fire) drops the key out of the held buffer the moment it's released. On a tap-dance key I have configured as tap = Backspace / hold = Shift / hold_after_tap = Backspace - rapidly tapping and then holding to repeat-delete only worked sporadically.

Fix: flow-tapped entries get their own held-buffer state, FlowTapped, and on release they downgrade to EarlyFired if the key has a hold_after_tap action, the same thing early-fire already does. Plain tap-hold keys have no hold_after_tap, so nothing changes for them.

I put this in a separate state rather than reusing the ProcessedButReleaseNotReportedYet release arm, since that arm also fires for permissive-hold and unilateral-tap resolutions and I'd rather not special-case those.

### DIFF
--- a/rmk/src/keyboard.rs
+++ b/rmk/src/keyboard.rs
@@ -450,7 +450,7 @@ impl<'a> Keyboard<'a> {
                 self.held_buffer.push(HeldKey::new(
                     event,
                     *key_action,
-                    KeyState::ProcessedButReleaseNotReportedYet(action),
+                    KeyState::FlowTapped(action),
                     now,
                     time_out,
                 ));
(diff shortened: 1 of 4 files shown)


jpds:
@HaoboGu Good to go in spite the CI failing on a external URL (and I already tried pushing again to unclog it).

HaoboGu:
thanks!

PR #750, opened 2026-03-03
New protocol

HaoboGu:
This PR contains RMK's new host communication protocol.

-----

UPDATE: Now the PR becomes quite large, I'll split it to several smaller parts.

-----

Phase 1: ICD Types and postcard-rpc Integration
Phase 2: Feature Gate and ProtocolService Skeleton
Phase 3: Support USB Raw Bulk Transport
Phase 4: Implement core Endpoints, make the basic protocol work
Phase 5: Security (Lock/Unlock) feature
Phase 6: Implement all remaining endpoints
Phase 7: Topic(Notification from keyboard to host)
Phase 8: Support BLE Serial
Phase 9: Host Tool

PR #848, opened 2026-05-13
New Protocol - V2

HaoboGu:
The 2nd try of adding RMK's protocol. Different from previous tries(#750, #835), this time `postcard-rpc` is dropped because it's kind of overkill and introduces long compile time.

Instead, a simpler, header + payload like protocol is used, and the new protocol is called `rynk` which means `rmk` + `link`.

## Progress

All sub-PRs are merged into this branch:

- [x] #847 — Rename rmk_protocol to rynk, add basic rynk protocol, remove postcard-rpc
- [x] #850 — Add Rynk core service struct
- [x] #852 — Add rynk command handlers
- [x] #863 — Make `RynkMessage` a wrapper type
- [x] #853 — Transports for Rynk
- [x] #872 — Add rynk dispatch loopback integration test
- [x] #873 — Rynk host library
- [x] #893 — Add `RynkHidService`
- [x] #894 — Add tsify wasm/TS type generation
- [x] #895 — Refactor Rynk host side trait, add rynk-wasm and web demo
- [x] #896 — Define Rynk layout in keyboard.toml
- [x] #908 — Add lock-gate security design
- [x] #907 — Add GetDeviceInfo endpoint
- [x] #909 — Implement bulk transfer endpoints
- [x] Documentation — `docs/docs/main/docs/features/rynk.md` (wired into the features nav)


rmk-types/src/protocol/rynk/message.rs
```diff
@@ -0,0 +1,135 @@
...
+    fn cmd(&self) -> Result<Cmd, RynkError>;
+    fn seq(&self) -> Result<u8, RynkError>;
+    fn payload_len(&self) -> Result<u16, RynkError>;
+    fn payload(&self) -> Result<&[u8], RynkError>;
+    fn payload_mut(&mut self) -> Result<&mut [u8], RynkError>;
+
+    fn set_cmd(&mut self, cmd: Cmd) -> Result<(), RynkError>;
+    fn set_seq(&mut self, seq: u8) -> Result<(), RynkError>;
+    fn set_payload_len(&mut self, len: u16) -> Result<(), RynkError>;
+}
```
pcasotti: I'd prefer these to not return `Result` and instead be parsed before hand:
```rust
pub struct MyMessage;

impl RynkMessage for MyMessage {
    fn cmd(&self) -> Cmd {
        todo!()
    }
    // ..
}

impl TryFrom<&[u8]> for MyMessage {
    type Error = RynkError;

    fn try_from(value: &[u8]) -> Result<Self, Self::Error> {
        todo!()
    }
}

pub trait RynkMessage {
    fn cmd(&self) -> Cmd;
    // ..
}
```
Something like this, so the error only exists when parsing from `&[u8]` once. This will also help remove the duplicate `if self.len() < RYNK_HEADER_SIZE { return Err(RynkError::InvalidRequest); }` in every method.

We might need `RynkMessage: TryFrom<&[u8]>`, lifetimes might get in the way but i believe its possible

rmk/src/host/rynk/handlers/combo.rs
```diff
@@ -0,0 +1,39 @@
...
+
+    pub(crate) async fn handle_set_combo(&self, payload: &mut [u8]) -> Result<usize, RynkError> {
+        let (r, _) = postcard::take_from_bytes::<SetComboRequest>(payload).map_err(|_| RynkError::InvalidRequest)?;
+        self.ctx.set_combo(r.index, r.config).await;
+        Self::write_response(&(), payload)
+    }
+
+    #[cfg(feature = "bulk_transfer")]
+    pub(crate) async fn handle_get_combo_bulk(&self, _payload: &mut [u8]) -> Result<usize, RynkError> {
+        Err(RynkError::Internal)
```
pcasotti: A specific `RynkError::Unimplemented` for these cases would be nice.

rmk/src/host/rynk/handlers/system.rs
```diff
@@ -0,0 +1,85 @@
...
+        Self::write_response(&caps, payload)
+    }
+
+    pub(crate) async fn handle_reboot(&self, payload: &mut [u8]) -> Result<usize, RynkError> {
+        // Synchronous reset — function never returns on real hardware. The
+        // wire response is moot; the host treats post-Reboot disconnect as
+        // success. On std/test targets the call falls through and we send
+        // an Ok envelope so loopback tests complete.
+        crate::boot::reboot_keyboard();
+        Self::write_response(&(), payload)
```
pcasotti: Does reversing the order of the operations would allow to respond to the host before rebooting?

rmk/src/host/rynk/handlers/system.rs
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    pub(crate) async fn handle_bootloader_jump(&self, payload: &mut [u8]) -> Result<usize, RynkError> {
```
pcasotti: This probably needs a similar comment as `handle_reboot`

rmk/src/host/rynk/mod.rs
```diff
@@ -0,0 +1,185 @@
...
+impl<'a> RynkService<'a> {
+    pub fn new(keymap: &'a KeyMap<'a>) -> Self {
+        Self {
+            ctx: KeyboardContext::new(keymap),
+        }
+    }
+
+    /// Process one inbound message in place. Always writes a response
+    /// envelope (Ok or Err) into `msg`; `cmd` and `seq` are echoed verbatim.
+    pub async fn dispatch(&self, msg: &mut [u8]) {
```
pcasotti: Use `RynkMessage` instead of `&[u8]` here and in the other methods in this file.

pcasotti (changes requested):

rmk-types/src/protocol/rynk/message.rs
HaoboGu: This makes sense! I'll try

rmk/src/host/rynk/handlers/system.rs
HaoboGu: theoretically yes, but it's hard to guarantee the sending is successful, so maybe just remove it?

rmk/src/host/rynk/mod.rs
HaoboGu: It adds something like: `pub async fn dispatch<'a, M: RynkMessage<'a>>(&self, frame: &mut M)` which I feel is a little bit verbose.

I made RynkMessage a trait because of orphan rule, maybe making `RynkMessage` a wrapper of `&'a [u8]` is better?

```rust
struct RynkMessage<'a> {
  buf: &'a [u8]
}
```

HaoboGu: Or a new type like struct: `struct RynkMessage<'a>(&'a mut [u8]);`

rmk-types/src/protocol/rynk/message.rs
HaoboGu: Everything is fixed in #863

pcasotti (approved):

Schievel1:
Iirc this is also used as the protocol between central and peripheral. 
If we want to flash the peripheral via the central we can't loose any message during transfer. 

I can count a sequence number and simply not start the update if a message went missing. Or we need flow control e.g. with a sliding window for rynk. For the latter I would say it's better to have that reusable in the protocols instead of my firmware update cooking up its own flow control.

HaoboGu:
> Iirc this is also used as the protocol between central and peripheral. If we want to flash the peripheral via the central we can't loose any message during transfer.
> 
> I can count a sequence number and simply not start the update if a message went missing. Or we need flow control e.g. with a sliding window for rynk. For the latter I would say it's better to have that reusable in the protocols instead of my firmware update cooking up its own flow control.

In the current design Rynk is not used between central and peripheral, but I hope it can be used in the future. Now it has only a `SEQ` number. Could you explain further about the "flow control" mechanism?

Schievel1:
> > Iirc this is also used as the protocol between central and peripheral. If we want to flash the peripheral via the central we can't loose any message during transfer.
> 
> > 
> 
> > I can count a sequence number and simply not start the update if a message went missing. Or we need flow control e.g. with a sliding window for rynk. For the latter I would say it's better to have that reusable in the protocols instead of my firmware update cooking up its own flow control.
> 
> 
> 
> In the current design Rynk is not used between central and peripheral, but I hope it can be used in the future. Now it has only a `SEQ` number. Could you explain further about the "flow control" mechanism?

What I mean with flow control is basically what I already did for the split protocol in #886. So mostly checking if the data is correct and resending a packet if it isn't. 
But after considering this a bit longer I tend towards a solution where this is on the application layer. (What I already did in #886)
If we later on find out we use similar implementations for this across different connections (I mean firmware via ble, firmware via uart from central to split, some different thing that sends data over rynk and must have consistency etc.) we can still merge all those into a unified solution and put it into rynk.

Schievel1:
> > Iirc this is also used as the protocol between central and peripheral. If we want to flash the peripheral via the central we can't loose any message during transfer.
> 
> > 
> 
> > I can count a sequence number and simply not start the update if a message went missing. Or we need flow control e.g. with a sliding window for rynk. For the latter I would say it's better to have that reusable in the protocols instead of my firmware update cooking up its own flow control.
> 
> 
> 
> In the current design Rynk is not used between central and peripheral, but I hope it can be used in the future. Now it has only a `SEQ` number. Could you explain further about the "flow control" mechanism?

What I mean with flow control is basically what I already did for the split protocol in #886. So mostly checking if the data is correct and resending a packet if it isn't. 
But after considering this a bit longer I tend towards a solution where this is on the application layer. (What I already did in #886)
If we later on find out we use similar implementations for this across different connections (I mean firmware via ble, firmware via uart from central to split, some different thing that sends data over rynk and must have consistency etc.) we can still merge all those into a unified solution and put it into rynk.

HaoboGu:
In favor of #962

PR #853, opened 2026-05-14
Transports for Rynk

HaoboGu:


rmk-config/src/lib.rs
```diff
@@ -848,6 +852,7 @@ impl Default for HostConfig {
     fn default() -> Self {
         Self {
             vial_enabled: true,
+            rynk_enabled: false,
```
pcasotti, line 855: Maybe we should switch and prefer rynk as default.

rmk-types/src/protocol/rynk/message.rs
```diff
@@ -81,12 +83,12 @@ impl<'a> TryFrom<&'a mut [u8]> for RynkMessage<'a> {
     /// declared payload (`buf.len() >= RYNK_HEADER_SIZE + payload_len`).
     fn try_from(buf: &'a mut [u8]) -> Result<Self, RynkError> {
         if buf.len() < RYNK_HEADER_SIZE {
-            return Err(RynkError::InvalidRequest);
+            return Err(RynkError::Malformed);
```
pcasotti, line 106: `RynkError::Internal`?

HaoboGu: good catch! Yeah it should be `Internal`

HaoboGu: wait, this buffer is the inbound message, so if the inbound message's length is even smaller than header size, the message is Malformed, it seems correct.


I thought it was a local allocated buffer

pcasotti: So `try_from` is for inbound and `build` is outbound?

HaoboGu: Yes

rmk/src/host/rynk/mod.rs
```diff
@@ -160,3 +150,314 @@ impl<'a> RynkService<'a> {
...
+    ///
+    /// Owns the reassembly/dispatch buffer, parses frames as `5 + LEN`
+    /// headers, dispatches each frame in place via
+    /// [`dispatch`](RynkService::dispatch), and emits topic frames between
+    /// request/response turns.
+    ///
+    /// Transport-specific setup and reconnect both stay in the caller.
+    pub async fn run_session<R: Read, T: Write>(&self, rx: &mut R, tx: &mut T) {
+        let mut buf = [0u8; RYNK_BUFFER_SIZE];
+        let mut topics = topics::TopicSubscribers::new();
```
HaoboGu, line 162: **Boot/connection panic: `TopicSubscribers` over-subscribes events whose `subs` budget is too small.**

`TopicSubscribers::new()` (run once per `run_session`) subscribes to `BatteryStatusEvent`, `LayerChangeEvent`, `ConnectionStatusChangeEvent`, etc. The generated `subscriber()` does `PubSubChannel.subscriber().expect(...)`, i.e. it **panics** when the `subs` limit is exceeded.

Two concrete crashes:
- `event.battery_status` has `subs = 0` in `event_default.toml`, but this path subscribes to it under `#[cfg(feature = "_ble")]`. So *every* `_ble` + `rynk` build panics the first time a rynk session starts: `Failed to create subscriber for BatteryStatusEvent. The 'subs' limit has been exceeded.`
- `layer_change` and `connection_status_change` have `subs = 1`. Dual-transport boards attach the same `HostService` to **both** USB and BLE (see `examples/use_rust/nrf52840_ble/src/main.rs:217-220`), so two `run_session`s run concurrently → two layer/conn subscribers → the second panics. Split central builds already take a layer/conn subscriber in `split/driver.rs`, so split + rynk overflows too.

Fix: bump `battery_status`/`layer_change`/`connection_status_change` (and `sleep_state`/`wpm_update` for split+display) `subs` to cover the per-transport host subscribers, or share one `TopicSubscribers` across transports instead of one per session.

rmk/Cargo.toml
```diff
@@ -105,7 +105,7 @@ features = ["split", "vial", "async_matrix", "_ble"]
 cortex-m = { version = "0.7" }
 
 [features]
-default = ["defmt", "storage", "vial", "vial_lock", "watchdog"]
+default = ["defmt", "storage", "rynk", "watchdog"]
```
HaoboGu: **Default-feature change not mirrored in the macro's `get_rmk_features()` → parity validation mis-detects the active host protocol.**

This flips the default from `vial`/`vial_lock` to `rynk`, but `rmk-macro/src/codegen/feature.rs:22-25` still hardcodes the *old* default set for default-features deps:
```rust
if default_features {
    feature_set.push("defmt"); feature_set.push("storage");
    feature_set.push("vial"); feature_set.push("vial_lock");  // stale: should be "rynk"
}
```
For any keyboard that depends on `rmk` with default features, `validate_feature_config_parity` now sees a *faked* feature list (`vial` on, `rynk` off) that no longer matches the crate that actually compiled (`rynk` on, `vial` off). The macro then generates host wiring for the wrong protocol (or emits a spurious parity panic), even though the firmware was built with rynk. Update `get_rmk_features()` to push `rynk` instead of `vial`/`vial_lock`.

rmk/src/ble/mod.rs
```diff
@@ -382,23 +410,39 @@ async fn gatt_events_task(server: &Server<'_>, conn: &GattConnection<'_, '_, Def
...
-                                    warn!("Wrong host packet data: {:?}", event.data());
+                                    warn!("Wrong Vial packet data: {:?}", data);
+                                }
+
+                                #[cfg(feature = "rynk")]
+                                if !data.is_empty() {
+                                    debug!("Got Rynk packet ({} bytes)", data.len());
+                                    // Awaits the pipe's backpressure when the rynk
+                                    // consumer falls behind.
+                                    crate::channel::RYNK_BLE_RX_PIPE.write_all(data).await;
```
HaoboGu, line 435: **Backpressure here stalls the entire GATT event loop, including keyboard HID.**

`RYNK_BLE_RX_PIPE.write_all(data).await` runs inside `gatt_events_task`, *before* `gatt_event.accept()` (line ~463). The 512-byte pipe is drained only by `run_session`, which can be parked awaiting a slow dispatch (e.g. a flash-bound `SetKeyAction`). If the host pushes rynk data faster than the session drains, `write_all` blocks, and since this task serially services *all* GATT events for the connection, queued keyboard/mouse/LED/CCCD events stop being processed until the rynk consumer catches up — the BLE keyboard appears to hang during a config transfer. The old Vial path used `try_send` + drop-with-warning to keep the loop responsive. Consider a try-write/drop (or a dedicated drain) so rynk backpressure can't block HID.

rmk/src/usb/vial.rs
```diff
@@ -0,0 +1,119 @@
...
+    writer: &'a mut HidWriter<'static, D, 32>,
+}
+
+impl<D: Driver<'static>> ErrorType for VialUsbTx<'_, D> {
+    type Error = VialUsbError;
+}
+
+impl<D: Driver<'static>> Write for VialUsbTx<'_, D> {
+    async fn write(&mut self, buf: &[u8]) -> Result<usize, Self::Error> {
+        let n = buf.len().min(32);
```
HaoboGu: **`write` silently truncates to 32 bytes instead of erroring on a size mismatch.**

`let n = buf.len().min(32); ... Ok(n)` means a `buf` larger than 32 returns `Ok(32)` and drops the tail. `embedded_io_async::Write::write_all` only stops when the returned count equals `buf.len()`, so a >32-byte write would loop re-sending overlapping 32-byte windows, corrupting the HID report stream. Today callers always pass exactly 32 bytes so it's benign, but this is inconsistent with `VialUsbRx` and `VialBleTx`, which reject a size mismatch. Prefer returning an error (or asserting) when `buf.len() != 32` rather than masking it with `min(32)`.

rmk/src/host/rynk/mod.rs
```diff
@@ -160,3 +150,314 @@ impl<'a> RynkService<'a> {
...
+
+        loop {
+            // 1. Read the fixed header or a topic
+            match select(rx.read(&mut buf[..RYNK_HEADER_SIZE]), topics.next_event()).await {
```
HaoboGu, line 166: **`select` cancels the in-flight `rx.read` whenever a topic fires; on cancellation-unsafe transports this drops received bytes and desyncs the frame stream.**

Each loop iteration races `rx.read(&buf[..RYNK_HEADER_SIZE])` against `topics.next_event()`. When a topic event wins, the partially-completed read future is dropped. For the BLE `Pipe` this is safe (bytes stay buffered), but `run_rynk_uart` drives `run_session` over an arbitrary embassy `Read` whose `read` future may have already consumed bytes from a DMA/ring buffer when cancelled. Those bytes are lost; the next iteration reads a fresh header from the wrong offset, mis-parses `LEN`, and triggers the oversize-drain path — dropping a real request frame. Worth documenting the cancel-safety requirement on the transport, or buffering the header read so a topic interrupt can't discard partial input.

rmk/src/usb/vial.rs
```diff
@@ -0,0 +1,119 @@
...
+/// Reconnect loop. Awaits host endpoint readiness, runs one Vial session,
+/// then loops back to wait again. The Rx/Tx adapter pair is recreated each
+/// iteration — they're zero-cost `&mut` borrows of `reader`/`writer`.
+pub async fn run_host_usb<D: Driver<'static>>(
+    reader: &mut HidReader<'static, D, 32>,
+    writer: &mut HidWriter<'static, D, 32>,
+    service: &VialService<'_>,
+) -> ! {
+    loop {
+        reader.ready().await;
```
HaoboGu, line 28: **Reuse: `VialUsbError`, `VialBleError`, and `RynkTransportError` are three byte-identical `embedded_io_async::Error` newtypes.**

All three define the same `Display` / `core::error::Error` / `Error::kind() -> ConnectionReset` / `defmt::Format` impls (~30 lines each). `RynkTransportError` already lives in the shared `host/rynk/transport.rs`. The two Vial copies (`usb/vial.rs:36`, `ble/vial.rs:32`) duplicate it verbatim, so any change to error semantics must be hand-applied in three places and can silently drift. Consider hoisting one shared transport-error type (it needn't be `_ble`-gated) and reusing it across all adapters.

rmk/src/host/mod.rs
```diff
@@ -1,21 +1,31 @@
...
+/// UART-backed rynk transport helper.
+#[cfg(feature = "rynk")]
+pub use rynk::run_rynk_uart;
 #[cfg(feature = "vial")]
 pub use via::VialService as HostService;
+
+/// Run one host-protocol session over a BLE connection. Resolves to the
+/// rynk transport under `rynk` and the Vial transport under `vial`.
+#[cfg(all(feature = "rynk", feature = "_ble"))]
+pub(crate) use crate::ble::rynk::run_host_ble;
```
HaoboGu, line 31: **Project-rule violation: in-crate `pub(crate) use` for convenience.**

`run_host_ble` / `run_host_usb` / `build_host_usb` are re-exported here purely so call sites can spell `crate::host::...` instead of the real `crate::ble::rynk::` / `crate::usb::rynk::` path. CLAUDE.md: *"Don't use `pub use` for convenient usage within the crate."* This indirection also hides where the function actually lives and which protocol impl is selected. Import the concrete feature-gated paths at the two call sites (`usb/mod.rs`, `ble/mod.rs`) instead.

rmk-config/src/lib.rs
HaoboGu: I'd rather keep vial as default for compatibility now.

PR #854, opened 2026-05-15
feat(osm): add configurable quick_release option for one-shot modifiers

yekingyan:
## What

  Add a `quick_release` option to `[behavior.one_shot_modifiers]` that controls when one-shot modifiers are released.

  - `quick_release = false` (default): modifier releases on next key **release** (chain mode, equivalent to ZMK `&skn`)
  - `quick_release = true`: modifier releases on next key **press** (equivalent to ZMK `&skq`)

  ## Problem

  With the default chain mode, the one-shot modifier stays active until the next key is released. During fast typing, key presses often overlap (the
  next key is pressed before the previous one is fully released), causing the modifier to affect multiple keys unintentionally. For example, tapping OSM
   Shift then quickly typing "ab" produces "AB" instead of the expected "Ab".

  Setting `quick_release = true` solves this by releasing the modifier as soon as the next key is pressed, guaranteeing only one character is modified
  regardless of typing speed.

  ## Backward compatibility

  Default is `false`, preserving existing behavior. The option is `Option<bool>` in TOML parsing, so existing `keyboard.toml` files without this field
  work without any changes.

  ## Testing

  Tested both `quick_release = true` and `quick_release = false` on hardware:
  - Fast typing with key overlap: only one character modified in quick-release mode
  - Key repeat (hold a key): `Aaaa...` in quick-release, `AAAA...` in chain mode
  - OSM chaining (multiple OSM taps): works in both modes
  - Combo + OSM interaction: no premature modifier clearing

### DIFF
--- a/rmk/src/keyboard.rs
+++ b/rmk/src/keyboard.rs
@@ -1533,7 +1533,11 @@ impl<'a> Keyboard<'a> {
             _ => warn!("KeyCode variant not supported: {:?}", key),
         }
 
-        self.update_osm(event);
+        let quick_release = self.keymap.one_shot_modifiers_config().quick_release;
+        let osm_consumed = self.update_osm(event);
+        if quick_release && osm_consumed && key.is_basic_keyboard_key() && event.pressed {
+            self.send_keyboard_report_with_resolved_modifiers(true).await;
+        }
         self.update_osl(event);
     }
 
--- a/rmk/tests/keyboard_combo_test.rs
+++ b/rmk/tests/keyboard_combo_test.rs
@@ -1,7 +1,7 @@
 pub mod common;
 
 use embassy_time::Duration;
-use rmk::config::{BehaviorConfig, CombosConfig, MorsesConfig, OneShotConfig};
+use rmk::config::{BehaviorConfig, CombosConfig, MorsesConfig, OneShotConfig, OneShotModifiersConfig};
 use rmk::keyboard::combo::{Combo, ComboConfig};
 use rmk::types::keycode::HidKeyCode;
 use rmk::types::modifier::ModifierCombination;
@@ -445,6 +445,66 @@ fn test_overlapping_triggered_combos_release_all_outputs() {
             // Order of the two release reports depends on combo iteration order;
             // `M+,` is index 0 so its output (RightBracket) releases first.
             [0, [kc_to_u8!(Equal), 0, 0, 0, 0, 0]],
+            [0, [0; 6]],
+        ]
+    }
+}
+
+#[test]
+fn test_combo_with_one_shot_modifier_quick_release() {
+    key_sequence_test! {
+        keyboard: create_test_keyboard_with_config(BehaviorConfig {
+            combo: get_combos_config(),
+            one_shot: OneShotConfig {
+                timeout: Duration::from_millis(300),
+                ..Default::default()
+            },
+            one_shot_modifiers: OneShotModifiersConfig {
+                quick_release: true,
+                ..Default::default()
+            },
+            ..Default::default()
+        }),
+        sequence: [
+            [1, 3, true, 10],
+            [1, 5, true, 10],
+            [1, 3, false, 50],
+            [1, 5, false, 70],
+            [1, 3, true, 50],
+            [1, 3, false, 110],
+        ],
+        expected_reports: [
+            [KC_LSHIFT, [HidKeyCode::E as u8, 0, 0, 0, 0, 0]],
+            [0, [HidKeyCode::E as u8, 0, 0, 0, 0, 0]],
+            [0, [0; 6]],
+        ]
+    }
+}
+
+#[test]
+fn test_overlapped_combo_quick_release() {
+    key_sequence_test! {
+        keyboard: create_test_keyboard_with_config(BehaviorConfig {
+            combo: get_combos_config(),
+            one_shot_modifiers: OneShotModifiersConfig {
+                quick_release: true,
+                ..Default::default()
+            },
+            ..Default::default()
+        }),
+        sequence: [
+            [1, 3, true, 10],
+            [1, 5, true, 10],
+            [1, 3, false, 50],
+            [1, 5, false, 10],
+            [1, 4, true, 100],
+            [1, 3, true, 10],
+            [1, 4, false, 50],
+            [1, 3, false, 10],
+        ],
+        expected_reports: [
+            [KC_LSHIFT, [HidKeyCode::A as u8, 0, 0, 0, 0, 0]],
+            [0, [HidKeyCode::A as u8, 0, 0, 0, 0, 0]],
             [0, [0; 6]],
         ]
     }
--- a/rmk/tests/keyboard_one_shot_test.rs
+++ b/rmk/tests/keyboard_one_shot_test.rs
@@ -540,4 +540,219 @@ mod one_shot_test {
             ]
         };
     }
+
+    /// Chain mode (quick_release = false): modifier released on key RELEASE
+    #[test]
+    fn test_osm_chain_mode_basic() {
+        key_sequence_test! {
+            keyboard: create_test_keyboard_with_one_shot_modifiers_config(OneShotModifiersConfig {
+                quick_release: false,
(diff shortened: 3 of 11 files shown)


ldsands:
I like this idea but if I may I have a quick thought. What about chaining modifiers (e.g. ctrl+shift+p)? Might there be a need to have an additional config to go with your proposed quick_release that exempts the quick_release from working on other modifier keys?

yekingyan:
> I like this idea but if I may I have a quick thought. What about chaining modifiers (e.g. ctrl+shift+p)? Might there be a need to have an additional config to go with your proposed quick_release that exempts the quick_release from working on other modifier keys?

Thanks for the question! OSM chaining already works correctly with `quick_release = true`.

  When you tap multiple OSM keys in sequence (e.g. OSM Shift → OSM Ctrl → P), each OSM tap goes through
  `process_action_osm` which accumulates modifiers (Single(Shift) → Single(Shift|Ctrl)). It does not
  trigger `update_osm`, so the previous OSM state is preserved.

  The quick-release only fires when a non-OSM key is pressed — at that point all accumulated modifiers are
   applied to that single key and then cleared.

  I've tested this on hardware with a home-row OSM layout:

  ```toml
  [behavior.one_shot_modifiers]
  activate_on_keypress = true
  quick_release = true

  Keymap (layer 1 home row):
  ["OSM(LGui)", "OSM(LAlt)", "OSM(LCtrl)", "OSM(LShift)", ...]

  Tap OSM Shift → tap OSM Ctrl → press P correctly produces Ctrl+Shift+P.
  ```

rmk/src/keyboard.rs
```diff
@@ -1533,7 +1533,12 @@ impl<'a> Keyboard<'a> {
             _ => warn!("KeyCode variant not supported: {:?}", key),
         }
 
+        let quick_release = self.keymap.one_shot_modifiers_config().quick_release;
+        let osm_active_before = self.osm_state.value().is_some();
         self.update_osm(event);
+        if quick_release && osm_active_before && self.osm_state.value().is_none() && event.pressed {
+            self.send_keyboard_report_with_resolved_modifiers(true).await;
```
HaoboGu, line 1539: Should reports sent where be limited to HID keys, excluding keys like media/consumer-control/system keys?

rmk/tests/keyboard_one_shot_test.rs
```diff
@@ -39,7 +39,13 @@ mod one_shot_test {
     ];
 
     fn create_test_keyboard() -> Keyboard<'static> {
-        let behavior_config: &'static mut BehaviorConfig = Box::leak(Box::new(BehaviorConfig::default()));
+        let behavior_config: &'static mut BehaviorConfig = Box::leak(Box::new(BehaviorConfig {
+            one_shot_modifiers: OneShotModifiersConfig {
+                quick_release: true,
```
HaoboGu: The previous tests should be kept

rmk/src/keyboard.rs
yekingyan: > Should reports sent where be limited to HID keys, excluding keys like media/consumer-control/system keys?

Good point. Added an `is_basic_keyboard_key` check — the follow-up report is now only sent for basic keyboard keycodes, skipping  media/consumer/system/mouse keys.

rmk/tests/keyboard_combo_test.rs
```diff
@@ -154,6 +154,10 @@ fn test_combo_with_one_shot_modifier() {
                 timeout: Duration::from_millis(300),
                 ..Default::default()
             },
+            one_shot_modifiers: OneShotModifiersConfig {
```
HaoboGu: Don't modify current tests, add new tests with new configurations.

rmk/src/keyboard.rs
```diff
@@ -1004,21 +1004,21 @@ impl<'a> Keyboard<'a> {
             self.process_key_action(&action, new_event, true, Instant::now()).await;
             debug!("[Combo] {:?} triggered", action);
             embassy_time::Timer::after_millis(20).await;
-            // Reset other combos' state
-            self.reset_combo(key_action);
+            // Reset non-triggered sub-combos of the combo that just won.
+            self.reset_sub_combos(&combo_actions);
```
HaoboGu: Is this combo change related to the osm change? If not, make it a separate PR please, thanks

rmk/src/keyboard.rs
```diff
@@ -1533,7 +1533,22 @@ impl<'a> Keyboard<'a> {
             _ => warn!("KeyCode variant not supported: {:?}", key),
         }
 
+        let quick_release = self.keymap.one_shot_modifiers_config().quick_release;
+        let osm_active_before = self.osm_state.value().is_some();
         self.update_osm(event);
+        let is_basic_keyboard_key = matches!(key,
```
HaoboGu: This should be extracted as a helper in `impl KeyCode`.

rmk/src/keyboard.rs
```diff
@@ -1533,7 +1533,22 @@ impl<'a> Keyboard<'a> {
...
+            KeyCode::Hid(hid) if hid.process_as_consumer().is_none()
+                && hid.process_as_system_control().is_none()
+                && !hid.is_mouse_key()
+        );
+        if quick_release
```
HaoboGu: This check is verbose.. is there a better way doing it? For example, make `update_osm` return whether the `osm` is triggered, and then check here?

rmk/tests/keyboard_combo_test.rs
yekingyan: > Don't modify current tests, add new tests with new configurations.

Done. Original tests are untouched, new quick_release tests added separately.

rmk/src/keyboard.rs
yekingyan: The combo change (reset_sub_combos) is independent of the OSM feature — I've removed it from this PR.

rmk/src/keyboard.rs
yekingyan: > This should be extracted as a helper in `impl KeyCode`.

Extracted as KeyCode::is_basic_keyboard_key() in rmk-types/src/keycode/mod.rs.

rmk/src/keyboard.rs
yekingyan: > This check is verbose.. is there a better way doing it? For example, make `update_osm` return whether the `osm` is triggered, and then check here?

update_osm() now returns bool indicating whether the OSM was consumed. Call site simplified to:
 let osm_consumed = self.update_osm(event);
 if quick_release && osm_consumed && key.is_basic_keyboard_key() && event.pressed { ... }

rmk/tests/keyboard_one_shot_test.rs
```diff
@@ -540,4 +540,219 @@ mod one_shot_test {
...
+                [KC_LSHIFT, [kc_to_u8!(A), 0, 0, 0, 0, 0]], // A with LShift
+                [0, [kc_to_u8!(A), 0, 0, 0, 0, 0]], // Quick-release: modifier removed
+                [0, [0, 0, 0, 0, 0, 0]], // All released
+                [0, [kc_to_u8!(B), 0, 0, 0, 0, 0]], // B without LShift
+                [0, [0, 0, 0, 0, 0, 0]], // All released
+            ]
+        };
+    }
+
+    // TODO: test_osm_quick_release_rolling removed — OSM + morse/tap-hold interaction
```
HaoboGu, line 661: Is the bug still here?

HaoboGu: @yekingyan could you also fix this? we can get this PR merged then

yekingyan: > Is the bug still here?

Sorry for the long silence — thanks for merging it anyway and for your patience.

Yes, the bug is still there on main: the OSM/OSL timeout uses a single `select`, so a non-consuming event (e.g. a layer-key release) cancels the wait and the OSM gets stuck. I'll open a separate PR that switches it to a deadline loop and adds the rolling test.

HaoboGu: Thanks a lot!

PR #856, opened 2026-05-16
fix: flush buffered keys after early-fired morse tap

### LINKED ISSUE #855: bug: EarlyFired tap dance swallows immediately following key press is their press and release sequences overlap
I have following layout on my main/default layer:
```
keys = """
B                F               L              P           Q              Quote     W              O              U               Y
TD(2)            TD(3)           TD(4)          TD(5)       K              J         TD(6)          TD(7)          TD(8)           TD(9)
X                V               M              D           Z              Minus     G              Comma          Dot             Slash
                                 TD(0)          LT(1,R,PH)  Enter          Backspace TD(1)          GraveEscape
"""
```

With morse behaviors configured like following:
```
[behavior.morse]
hold_timeout = "250ms"
gap_timeout = "250ms"

[behavior.morse.profiles]
# This profile is recommended on the home row, when enable_flow_tap = true, and the hold action activates a layer or acts as a modifier (aka home row mod)
HRM = { permissive_hold = true, enable_flow_tap = true, unilateral_tap = true, hold_timeout = "400ms", prior_idle_time = "120ms" }
FH = { hold_on_other_press = true }
PH = { permissive_hold = true }
TH = { normal_mode = true }

.....

[[behavior.morse.morses]]
tap = "C"
hold = "MO(3)"
hold_after_tap = "C"
profile = "HRM"
```

With this when I type word "could" I usually do roll in such a way that I release 'C' before after I press 'O' and release 'O' after I press 'U'. 

Expected output should be "cou", but instead I get "cuo".

alxzh:
When a tap-dance/morse key early-fires its tap action, keep the EarlyFired breadcrumb for hold-after-tap continuation but flush buffered non-morse keys if no unresolved morse key still blocks them.

Add a dusk-layout C/O/U rollover regression that reproduces the prior C-U-O ordering, plus gap-timeout and tap-hold controls.

This fixes #855

### DIFF
--- a/rmk/src/keyboard/morse.rs
+++ b/rmk/src/keyboard/morse.rs
@@ -47,15 +47,9 @@ impl<'a> Keyboard<'a> {
             _ => unreachable!(),
         };
 
-        // If there's still morse key in the held buffer, don't fire normal keys
-        // FIXME? is |Holding needed here?
-        if self
-            .held_buffer
-            .keys
-            .iter()
-            .any(|k| k.action.is_morse() && matches!(k.state, KeyState::Pressed(_)))
-        {
-            return; //?
+        // If there's still an unresolved morse key in the held buffer, don't fire normal keys.
+        if self.has_unresolved_morse_key() {
+            return;
         }
 
         self.fire_held_non_morse_keys().await;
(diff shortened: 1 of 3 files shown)


HaoboGu (approved):

rmk/src/keyboard/morse.rs
HaoboGu: Maybe the check here needs to be updated similarly?

alxzh: good point, updated

HaoboGu (approved):
Thanks for your work!

PR #859, opened 2026-05-21
feat: add StickyMod (SM) action for Alt+Tab-style modifier cycling

ldsands:
## Summary

Adds a new `SM(key, modifier)` action — StickyMod — that holds a modifier
across repeated presses of the same key, then releases it automatically when
any non-SM, non-modifier key is pressed, the active layer changes, or an
optional timeout expires.

**Primary use case:** Alt+Tab window/tab cycling. Bind `SM(Tab, LAlt)` to a
key; the first press sends Alt+Tab, subsequent presses send Tab (Alt stays
held), and Alt releases as soon as you press any other key or switch layers.

## Motivation

This is a direct port of the `KC.SK()` (Sticky Key) behavior from KMK firmware.
For those migrating from KMK, `SM(Tab, LAlt)` replicates `KC.SK(KC.LALT)` used
in conjunction with Tab for Alt+Tab cycling. This was the last regularly-used
KMK feature I needed in order to fully replicate my KMK keymap in RMK
(though others may find additional gaps).

This feature covers similar ground to #724 (Tabber), which I was aware of
before writing this implementation but found didn't quite fit my needs — I
preferred this approach because it generalizes to any key+modifier combination
rather than being Tab-specific, and includes timeout support. That said, I
have no attachment to the name `SM` or `StickyMod` — happy to rename this to
whatever fits best if this is merged.

### How it differs from OSM

| Behavior | `OSM(mod)` | `SM(key, mod)` |
|---|---|---|
| Modifier release | After the next single keypress | After any non-SM/non-modifier press, or layer change, or timeout |
| Key bundled | No — modifier only | Yes — modifier + key in one action |
| Repeatable cycling | No | Yes |

### Behavior details

- **First press:** activates SM state, sends `modifier + key`
- **Release:** unregisters key, modifier stays held
- **Subsequent presses:** modifier already active, sends `key` again; timeout resets
- **Release triggers:** any non-SM, non-modifier keypress; any layer deactivation; timeout expiry
- **Modifier exclusion:** `Modifier` actions and HID modifier keycodes (Shift, Ctrl, etc.)
  do *not* release SM — this lets Shift+Tab work for reverse cycling

### Optional timeout

Timeout is measured from the last SM **press** — repeated presses extend the
hold window. Implemented via the main `run()` loop deadline (same pattern as
mouse repeat), so it fires reliably regardless of how many press/release
cycles have occurred.

```toml
[behavior.sticky_mod]
timeout = "5s"   # auto-release modifier after 5s since last SM press
```

Default: no timeout (modifier held until released by keypress or layer change).

### TOML syntax

```toml
SM(Tab, LAlt)           # Alt+Tab cycling
SM(Tab, LCtrl)          # Ctrl+Tab cycling
SM(Tab, LCtrl|LShift)   # Ctrl+Shift+Tab (reverse cycling)
```

## Changes

- `rmk-types`: new `Action::StickyMod(KeyCode, ModifierCombination)` variant
- `rmk`: new `keyboard/sticky_mod.rs` module — `StickyModState` with optional deadline, state machine, and processing logic
- `rmk`: integrated into `keyboard.rs` — dispatch, modifier resolution, release triggers on layer deactivation, deadline-based timeout in `run()` loop
- `rmk-config`: TOML grammar (`keymap.pest`) and parser for `SM(key, mod)` syntax
- `rmk-macro`: codegen support for `SM` in `action_parser.rs` and `behavior.rs`
- `rmk`: `sm!()` macro in `layout_macro.rs`
- `docs`: `behavior.md` (Sticky Modifiers section), `layout.md` (SM syntax entry)

## Tests

7 integration tests in `rmk/tests/keyboard_sticky_mod_test.rs`:
1. Basic flow: press SM twice while MO held
2. Layer change cleanup: MO release triggers SM release
3. Shift integration: Shift key does not release SM (enables reverse cycling)
4. Rapid presses: 3× SM press/release cycle
5. Combined modifiers: `LCtrl|LShift` combination
6. Timeout: modifier auto-releases after inactivity
7. Timeout reset: pressing SM again extends the timeout window


HaoboGu:
There is [One Shot Sticky Modifier](https://rmk.rs/main/docs/configuration/behavior#one-shot-modifiers) in main branch, what's the difference?

ldsands:
> There is [One Shot Sticky Modifier](https://rmk.rs/main/docs/configuration/behavior#one-shot-modifiers) in main branch, what's the difference?

OSM(LAlt) releases the modifier after one keypress — press OSM, release, press Tab, Alt+Tab fires once and Alt is gone. SM(Tab, LAlt) bundles the modifier and key together and keeps the modifier held across repeated presses of the same SM key: first press sends Alt+Tab, second press sends Alt+Tab again (Alt still held), and Alt only releases when you press something else entirely. OSM is for one-shot use; SM is specifically for cycling (Alt+Tab, Ctrl+Tab) where you need the modifier to persist across multiple presses of the same key.

HaoboGu:
> OSM(LAlt) releases the modifier after one keypress — press OSM, release, press Tab, Alt+Tab fires once and Alt is gone. SM(Tab, LAlt) bundles the modifier and key together and keeps the modifier held across repeated presses of the same SM key: first press sends Alt+Tab, second press sends Alt+Tab again (Alt still held), and Alt only releases when you press something else entirely. OSM is for one-shot use; SM is specifically for cycling (Alt+Tab, Ctrl+Tab) where you need the modifier to persist across multiple presses of the same key.

Great, maybe the two types be merged into a single type of behavior, i.e. a general "Sticky Key"?

I'm imaging some like `SK(key, keep, max_repeat)`:

- When this SK is triggered, `key` is activated
- `key` keeps to be activated if keys in `keep` list is pressed
- `key` is released when `keep` is pressed for `max_repeat` times, or any other keys is triggered

With this, one-shot mod can be represented as `SK(mod, [], 1)`, and `SM(Tab, LAlt)` can be represented as `SK(Tab, [LAlt], MAX_REPEAT)`

What do you think?

ldsands:
> > OSM(LAlt) releases the modifier after one keypress — press OSM, release, press Tab, Alt+Tab fires once and Alt is gone. SM(Tab, LAlt) bundles the modifier and key together and keeps the modifier held across repeated presses of the same SM key: first press sends Alt+Tab, second press sends Alt+Tab again (Alt still held), and Alt only releases when you press something else entirely. OSM is for one-shot use; SM is specifically for cycling (Alt+Tab, Ctrl+Tab) where you need the modifier to persist across multiple presses of the same key.
> 
> Great, maybe the two types be merged into a single type of behavior, i.e. a general "Sticky Key"?
> 
> I'm imaging some like `SK(key, keep, max_repeat)`:
> 
> * When this SK is triggered, `key` is activated
> * `key` keeps to be activated if keys in `keep` list is pressed
> * `key` is released when `keep` is pressed for `max_repeat` times, or any other keys is triggered
> 
> With this, one-shot mod can be represented as `SK(mod, [], 1)`, and `SM(Tab, LAlt)` can be represented as `SK(Tab, [LAlt], MAX_REPEAT)`
> 
> What do you think?

I like that idea, since the two are conceptually similar. Let me think through what implementing this in the one-shot would need (and what I'd want from it).

Max repeat, on the other hand, isn't something I'd personally use, though I see the utility. Right now I'm thinking of the browser tabs I have open. As long as we can set "no max repeat" or "infinite" as an option, that works for me.

I'd also like a timeout that's independent of the one-shot timeout. This probably isn't strictly necessary, but I suspect most people who use this would want a different timeout than the one for one-shot keys. I usually exit a sticky mod with my layer button, but when cycling through browser tabs I'll sometimes go through several, pause to look at the screen, then continue. Again, a personal preference I could make work with a shared timeout, but worth mentioning.

I also wonder how this interacts with layer changes. I use this key on another layer (via MO) so that returning to the base layer automatically exits the sticky mod and I can resume typing immediately. So I'd want an optional per-key feature to exit on layer change. I wouldn't want this on my other one-shot keys, though, since I use those across layers constantly; I added it specifically to this sticky mod implementation.

Forgive the verbosity; talking through it helped. I'm happy to fold this into the current one-shot keys implementation, but if you want to go that route, I would prefer a per-key timeout option and a per-key "exit on layer change" option. Adding max retries is a great idea. I don't know what you'd want as the default, but I'd like to have the max retries configurable to have infinite, or to assume infinite until the timeout from the last sticky mod key press.

What do you think?

HaoboGu:
> Max repeat, on the other hand, isn't something I'd personally use, though I see the utility. Right now I'm thinking of the browser tabs I have open. As long as we can set "no max repeat" or "infinite" as an option, that works for me.

Yes, I agree. Omitting it means "infinite", i.e. `SK(Tab, [LAlt], MAX_REPEAT) == SK(Tab, [LAlt])`.
But I'm not sure what the default behavior should be: max_repeat == 1 or infinite?

About the per-key timeout, I have no strong opinion on it. Using the shared timeout as the default and overriding it when a per-key timeout exists is fine with me. But it does increase the complexity and RAM/Flash usage.

> I also wonder how this interacts with layer changes.

Should it be included in the keep list?

Also, the length of keep list is a bit tricky. I think it should be calculated at compile-time and applied to the type. TBH I don't know if the idea works, but I think it's worth a try at least.

ldsands:
Just a quick update, I think I'll be done with this by the end of the week or earlier for you to look at. I may also wait until https://github.com/HaoboGu/rmk/pull/854 is done as well and make sure that I have it working with that merge.

HaoboGu:
Great!

Only a minor comment left in #854, I think we can get it merged first.

HaoboGu:
#854 is merged

ldsands:
Thanks for the feedback on consolidating SM and OSM! I've rebased this branch onto the latest `main` (which includes PR #854's `quick_release` feature) and reworked the implementation to address your request.

**What changed:**
- Removed `StickyMod` entirely — replaced by the unified `StickyKey` (`SK`) action
- `SK(key, [modifier], max_repeat, timeout_ms, exit_on_layer_change)` — all args after `key` are optional
  - `max_repeat = 0` → infinite repeats (default)
  - `timeout_ms = 0` → uses global `[behavior.sticky_key] timeout` (default; no timeout if unset)
  - `exit_on_layer_change` → defaults to `false` (modifier survives layer changes)
- `OSM` remains as-is for the one-shot use case; `SK` handles the "hold modifier across repeated presses" use case
- 11 integration tests added for SK behavior; all 450 tests passing
- Docs updated in `behavior.md` and `layout.md`

Example usage: `SK(Tab, [LAlt])` for Alt+Tab window cycling — first press sends Alt+Tab, each subsequent press sends Tab again while holding Alt, releases when you press anything else.

HaoboGu:
I was imaging `SK` would replace one-shot, but it seems in current implementation one-shot key is kept.

ldsands:
> I was imaging `SK` would replace one-shot, but it seems in current implementation one-shot key is kept.

@HaoboGu

I'm glad you brought this up. So as I mentioned earlier, I'm relying heavily on Claude, and so I clearly did not quite define well enough what I wanted it to do. Also, I somehow forgot that you were okay with completely replacing the one shot with sticky key, and so I was aiming to be completely backwards compatible. But re reading our earlier conversation, I realized that is not exactly what you wanted. I decided to merge in what I already had anyway because there were some changes that I thought were worth it to get into this branch and PR, and then I can now take another step of completely eliminating the one shot and migrating all those features into the sticky keys.

So I do still have a couple of questions for you. Mostly it's about how we want this formatted. I did make a couple of changes based on your recommendations from earlier, but you didn't quite cover everything in terms of how they should look for the new sticky key only migration. Below are examples of everything we'll need (I think).

```txt
old: OSM(LGui)
new: SK(LGui)

old: OSL(1)
new: SK(1) or SK(MO(1)) or SK(TO(1)) or SK("L1") or SK(Layer = 1) or SK(Layer, [1])

old: SK(Tab, [LAlt], 0, 0, true) or SK(Tab, [LCtrl])
new: SK(Tab, [LAlt])
```

So the OSM is very straightforward I assume no issues there though you had previously suggested `SK(LGui [], 1)` I think this simpler version is probably the way to go.

For OSL There are many different options here. I'm not really sure which one you'd prefer to take. If we just use SK(1) that'll be a very straightforward translation. However, I'm not sure that this is the way I would prefer. I think I prefer one of these two: `SK(MO(1)) or SK(TO(1))`. Do you have a strong preference for any of the options I put above, or if you have some other option you would prefer?

Finally, for the sticky mod I think just simplifying from your suggested `SK(Tab, [LAlt], MAX_REPEAT)` I would prefer to set the max repeat in the behavior profile.

Which brings us to the other major area where I would like your opinion.

As of right now (for the most part) we have three Toml profiles that handle the three different sets of one shot/sticky keys. Do note that the [behavior.sticky_key] options are implemented on the key definition itself but I will change that to just use the toml profile instead (more on this below).

old:
```toml
[behavior.one_shot]
timeout = "1s"                  # default: 1s   (Duration::from_secs(1))

[behavior.one_shot_modifiers]
activate_on_keypress = false    # default: false
quick_release = false           # default: false

[behavior.sticky_key]
timeout = "5s"                  # default: no timeout
max_repeat = 0                  # default: 0 = infinite (fires key on presses 1..=max_repeat, releases on max_repeat+1)
exit_on_layer_change = false    # default: false = SK survives layer changes (release only on key/timeout)
```

new:
```toml
[behavior.sticky_key]
timeout = "1s"                  # default: 1s   (Duration::from_secs(1))
activate_on_keypress = false    # default: false
quick_release = false           # default: false
max_repeat = 0                  # default: 0 = infinite (fires key on presses 1..=max_repeat, releases on max_repeat+1)
exit_on_layer_change = false    # default: false = SK survives layer changes (release only on key/timeout)
```

So I'm fairly certain that the behavior.one_shot and the behavior.one_shot_modifiers can easily be combined and just move the timeout into the behavior one shot modifiers and apply that default timeout to the one shot modifiers. If you have any issues with that, let me know. Now the behavior.sticky_key is a bit more complicated. I don't see any large issue with timeout being combined and applied to all one-shot/sticky keys, especially if down the road I implement an individual key override (which I do want so that I can increase the timeout of the alt tab sticky key/mod). The max repeat can be set there and does not interfere with the current one shot modifiers, but will only apply to the current sticky keys, so as I see it no issue combining there. Now the exit_on_layer_change could be more complicated. I could see some people not wanting a one-shot modifier to be held across a layer change. However, I would personally prefer that, but for my sticky mods, I do not want that. So once again, I would like to see an individual key override for this, but, I do not think it's a bad idea to keep this as a default that you set in the toml profile. Also, I'm not sure I actually like exit_on_layer_change as the name for this parameter. So if you have other suggestions for what to change this name to, let me know.

For now I don't think I'm going to implement any key level overrides to rather keep the combining as simple as possible. Then once I've confirmed that the consolidation is working well, then I would like to see what kind of effect adding these individual key overrides has on the RAM and flash.

What do you think?

HaoboGu:
Thanks for your detailed explanation! Yes some details needs to be clarified.

1. For OSM/OSL, I prefer `SK(LGui)` and `SK(Mo(1))`. And if we do this, the default value of `keep` is `[]` and `max_repeat` is 1. It requires a change of default `max_repeat` value from current implementation.
2. Yes all configurations should be merged into a single `behavior.sticky_key`, but I prefer to keep `max_repeat` in per-key config, because the `max_repeat` value should be always 1 for one-shot keys. It's impossible to have other values if we don't do per-key `max_repeat` config.
3. I'm fine with `exit_on_layer_change` and current behavior.

Thanks again for your effort! Please feel free to discuss anything you're unclear about.

ldsands:
@HaoboGu what do you think?

Claude's summary of all the changes below:

## Summary: unify One-Shot (OSM/OSL) and Sticky Key into a single `SK` feature

This consolidates the previously separate **one-shot modifier (`OSM`)**, **one-shot layer (`OSL`)**, and **sticky-key** mechanisms into one `SK(...)` action driven by a
single engine. The old `oneshot.rs` runtime is deleted and folded into `sticky_key.rs`; one config table and one keymap action now cover everything.

### The three `SK` shapes
`SK` picks its behavior from the shape of its argument:

| Shape | Syntax | Behavior |
|-------|--------|----------|
| **Pure-mod** | `SK(LGui)`, `SK(LCtrl\|LShift)` | One-shot modifier — held for the next key press, then auto-released (former `OSM`). |
| **Layer** | `SK(MO(n))` | One-shot layer — layer `n` active for the next key press, then released (former `OSL`). |
| **Tap-key** | `SK(Tab, [LAlt])`, `SK(Tab, [LCtrl\|LShift])` | Modifier held across **repeated presses of the same key** (Alt+Tab–style cycling); releases on any other key.
|

### Unified config: `[behavior.sticky_key]`
A single table replaces `[behavior.one_shot]` and `[behavior.one_shot_modifiers]`:

| Field | Default | Notes |
|-------|---------|-------|
| `timeout` | `1s` | Idle auto-release; applies to all shapes. |
| `activate_on_keypress` | `false` | Pure-mod only (OSSM) — send modifier on the SK press itself. |
| `quick_release` | `false` | Pure-mod only — release modifier on next key *press* vs *release*. |
| `max_repeat` | `0` | Tap-key cycling cap; `0` = unlimited. |
| `release_on_layer_change` | `false` | Whether a layer change releases the SK. |

(`activate_on_keypress`/`quick_release` are pure-mod–only and silently ignored for tap-key/layer shapes.)

### Breaking changes / migration
- `OSM(mod)` → `SK(mod)`, `OSL(n)` → `SK(MO(n))` — old actions now error at build time.
- `[behavior.one_shot]` / `[behavior.one_shot_modifiers]` tables → `[behavior.sticky_key]`.
- Old 5-positional `SK(key, [mod], max_repeat, timeout, exit_on_layer_change)` → `SK(key, [mod])` + the config table.
- `exit_on_layer_change` (per-key) → `release_on_layer_change` (global).
- Tap-key SKs now have a 1s default timeout (previously none).

A full migration table is in `docs/configuration/behavior.md`, which has been rewritten to document the shapes, every config field, and the migration.

### Tests
- `keyboard_sticky_key_test.rs` — 17 tests covering all three shapes, mixed-shape replacement, timeout/reset, `max_repeat`, `release_on_layer_change` survive/exit, and that
the pure-mod–only flags (`activate_on_keypress`/`quick_release`) are correctly *ignored* on tap-key SKs.
- `keyboard_one_shot_test.rs` — pure-mod path through the new engine, covering `activate_on_keypress` and `quick_release` (chain + quick-release modes, combined modifiers).
- All five `[behavior.sticky_key]` fields are exercised; the in-repo `use_config` examples are migrated to the new syntax.

### Internals
Wire format, storage, and Vial keycode handling updated for the unified action; the removed OSM/OSL keycodes degrade gracefully to `No` rather than panicking.

HaoboGu:
Sorry for the delay, I'm quite busy these days. 

A few comments/questions:

1. For `pure mod` and `layer` mode, the `max_repeat` is set to `1` and for `tap key` mode, the `max_repeat` is set to `0` automatically? If so this should be added to the doc I think
2. I don't want to remove `OSM`/`OSL` from the user space(users are already familiar with those concepts), only change the internal implementation.
3. Why `activate_on_keypress`/`quick_release` don't work on normal SK?

ldsands:
> Sorry for the delay, I'm quite busy these days.

@HaoboGu, No problem, I'm in the same boat, way too much to do, far too little time.

> 
> A few comments/questions:
> 
> 1. For `pure mod` and `layer` mode, the `max_repeat` is set to `1` and for `tap key` mode, the `max_repeat` is set to `0` automatically? If so this should be added to the doc I think

The mechanism is slightly different from what you described. max_repeat is a single global value (default 0), and it's read in exactly one place — the tap-key path:

```txt
┌───────────────────────────┬───────────────────┬──────────────────────────────────────────────────────────┐
│           Shape           │ Reads max_repeat? │                         Behavior                         │
├───────────────────────────┼───────────────────┼──────────────────────────────────────────────────────────┤
│ Tap-key (SK(Tab, [LAlt])) │ Yes               │ 0 (default) = unlimited; N > 0 = release after N presses │
├───────────────────────────┼───────────────────┼──────────────────────────────────────────────────────────┤
│ Pure-mod (SK(LGui) / OSM) │ No                │ Always applies to exactly one following key (hardcoded)  │
├───────────────────────────┼───────────────────┼──────────────────────────────────────────────────────────┤
│ Layer (SK(MO(n)) / OSL)   │ No                │ Always applies to exactly one following key (hardcoded)  │
└───────────────────────────┴───────────────────┴──────────────────────────────────────────────────────────┘
```

So the value isn't auto-set to 1 for one-shots and 0 for tap-key — it's one global that defaults to 0, and the one-shot shapes simply never read it (their "one key only" behavior is baked into their code path). I'll document this and fix the current max_repeat row in behavior.md, which misleadingly implies it applies to one-shot modifiers.

Proposed wording:

max_repeat — Tap-key SKs only. Caps how many repeated presses keep the modifier held; 0 (default) = unlimited. Pure-mod (SK(LGui)) and layer (SK(MO(n))) SKs ignore this — they always apply to exactly one following key.

I can't think of a way that Max Repeat would be useful for anything other than the sticky mod behavior. But it's basically essential for the sticky mod behavior to have more than one (and ideally many/infante max repeat). If you can think of a use for using Max Repeat on other one shot keys, I can add that, that's fine, but I just don't think it is needed. Again, feel free to fill me in on a use that you think this will be used for and I'm more than willing to add it, but then I think we will want to have separate max repeats for those two different behaviors.

> 2. I don't want to remove `OSM`/`OSL` from the user space(users are already familiar with those concepts), only change the internal implementation.

My bad, I'll add them back as aliases so it'll all use as much of the same code as possible.

> 3. Why `activate_on_keypress`/`quick_release` don't work on normal SK?

These are one-shot-modifier semantics, so they only apply to the pure-mod shape (OSM, equivalently SK(mod)):

- activate_on_keypress controls whether the modifier is sent to the host the instant the SK key itself is pressed (true), versus deferring it until the next key (false,
default). The tap-key shape already sends mod + key immediately on every press, so the flag would be a redundant no-op there.
- quick_release controls whether the modifier releases on the next key's press (true) or its release (false, default chain mode). The tap-key shape has no single terminating
key — it cycles the same key and releases on any other key — so there's nothing for that distinction to act on.
- The layer shape sends no modifier at all, so neither applies.

I'll make sure this is clearly documented alongside the fields.

Do those changes sound good (I'll get working on them and I'll probably be done with them soon).

HaoboGu:
Yes it looks good, thanks for your effort!

ldsands:
@HaoboGu 

I just fixed the conflicts with the latest commits and all checks passed. Let me know if you need anything else.

HaoboGu:
It takes some time to review, thanks in advance

rmk-config/src/resolved/behavior.rs
```diff
@@ -3,17 +3,15 @@ use std::collections::HashMap;
...
     pub forks: Option<Forks>,
     pub morse: Option<Morse>,
-}
-
-pub struct OneShot {
-    pub activate_on_keypress: Option<bool>,
-    pub quick_release: Option<bool>,
+    pub sticky_key_timeout_ms: Option<u64>,
+    pub sticky_key_activate_on_keypress: Option<bool>,
+    pub sticky_key_quick_release: Option<bool>,
```
HaoboGu: What about merging these into a `StickyKey` config struct

rmk/src/host/via/keycode_convert.rs
```diff
@@ -186,16 +177,6 @@ pub(crate) fn from_via_keycode(via_keycode: u16) -> KeyAction {
             let layer = via_keycode as u8 & 0x0F;
             KeyAction::Single(Action::LayerToggle(layer))
         }
-        0x5280..=0x529F => {
-            // One-shot layer
-            let layer = via_keycode as u8 & 0xF;
-            KeyAction::Single(Action::OneShotLayer(layer))
-        }
```
HaoboGu, line 196: Keep Vial OSM/OSL supported, just migrate to SK backend

rmk/src/layout_macro.rs
```diff
@@ -311,47 +311,72 @@ macro_rules! thp {
...
+/// - `$keep`: `ModifierCombination` held between presses
 ///
 /// # Example
 /// ```ignore
-/// osl!(1)  // Next key will be from layer 1, then return to current layer
-/// osl!(2)  // Next key will be from layer 2, then return to current layer
+/// sk!(Tab, ModifierCombination::LALT)  // SK(Tab, [LAlt])
 /// ```
 #[macro_export]
-macro_rules! osl {
```
HaoboGu, line 324: keep osl!/osm! macro

rmk/Cargo.toml
```diff
@@ -94,6 +94,7 @@ critical-section = { version = "1.2", features = ["std"] }
 env_logger = "0.11"
 ctor = "1.0"
 embedded-hal-mock = { version = "0.11.1", features = ["embedded-hal-async"] }
+rusty-fork = "0.3"
```
HaoboGu: This dep is not needed anymore with `cargo-nextest`

rmk/src/keyboard/sticky_key.rs
```diff
@@ -0,0 +1,402 @@
...
+
+        // For the layer shape, deactivate the active layer before clearing the latch.
+        if let StickyKeyState::Active {
+            layer: Some(layer_num), ..
+        } = self.sticky_key_state
+        {
+            self.keymap.deactivate_layer(layer_num);
+        }
+
+        self.sticky_key_state = StickyKeyState::None;
```
HaoboGu: If the state is set to None while a sticky key remains held down, it will cause the key to get stuck

rmk/src/keyboard/sticky_key.rs
```diff
@@ -0,0 +1,402 @@
...
+        } else {
+            self.process_sticky_tap_key(params, event).await;
+        }
+    }
+
+    /// Pure-mod (OSM) shape: accumulate the modifier across taps, apply it through the
+    /// terminating key, honor `activate_on_keypress`/`quick_release`.
+    async fn process_sticky_pure_mod(&mut self, params: StickyKeyAction, event: KeyboardEvent) {
+        let config = self.keymap.sticky_key_config();
+        let deadline = (config.timeout != Duration::MAX).then(|| Instant::now() + config.timeout);
```
HaoboGu: For one-shot keys, the deadline should be the `release_time` + timeout. For in current implementation, the deadline is set to the `press_time` + timeout for one-shot keys.

rmk/src/keyboard.rs
```diff
@@ -152,18 +155,29 @@ impl Runnable for Keyboard<'_> {
                 // Process buffered held key
                 self.process_buffered_key(key).await
             } else {
-                // If mouse repeat is pending, race subscriber against deadline
-                let event = if let Some(deadline) = self.mouse.next_deadline() {
+                // Race subscriber against any pending deadlines (mouse repeat, SK timeout)
+                let sk_deadline = self.sticky_key_state.deadline();
```
HaoboGu: Is it possible to avoid racing sk + mouse here? I want to keep the main run loop as simple as possible. I want to remove mouse deadline in the future as well.

rmk/src/keyboard.rs
```diff
@@ -1544,21 +1584,23 @@ impl<'a> Keyboard<'a> {
...
         }
-        self.update_osl(event);
     }
 
     /// Process layer switch action.
-    fn process_action_layer_switch(&mut self, layer_num: u8, event: KeyboardEvent) {
+    async fn process_action_layer_switch(&mut self, layer_num: u8, event: KeyboardEvent) {
         // Change layer state only when the key's state is changed
         if event.pressed {
             self.keymap.activate_layer(layer_num);
```
HaoboGu: Add 

```rust
if self.keymap.sticky_key_config().release_on_layer_change {
    self.release_sticky_key_if_active().await;
}
```
when activating layer?

HaoboGu:
Should we merge #866 first before merging this? Or the fix in #866 is already included in this PR?

ldsands:
> Should we merge #866 first before merging this? Or the fix in #866 is already included in this PR?

I can probably include it in here. I'll let you know for sure as I work through your comments.

Thanks again for all your comments I hope to be able to get everything worked out over the next day or two.

ldsands:
Okay I think I got it all. I also did some testing on my own keyboard and everything worked as intended except for "stacking" OSM and OSL I'm still working on that right now.

All 9 review items from HaoboGu's Jun 27-29 review on commit ebfe286 are resolved. Changes are on branch `feat/sticky-mod` at these commits:

| # | Item | Commit |
|---|------|--------|
| 1 | Group resolved Behavior fields into StickyKeyConfig struct | `34a65599` |
| 2 | Vial OSM/OSL → SK backend | `5ac8d57f` (pre-existing) |
| 3 | Add `osl!`/`osm!` macros | `d2d35f49` |
| 4 | Remove `rusty-fork` dev-dependency | `42dec113` |
| 5 | Guard timeout while key held (Pressed phase) | `acbe4f35` |
| 6 | Refresh deadline on pure-mod release (release_time + timeout) | `acbe4f35` |
| 7 | Simplify run loop — single `with_deadline` race + post-check | `f7bf7903` |
| 8 | `release_on_layer_change` on layer activation | `f7bf7903` |

### Post-review fixes (self-discovered during code review)

| Fix | Description | Commit |
|-----|-------------|--------|
| A | Trailing newline in keyboard_sticky_key_test.rs | `2f30a0c2` |
| B | `with_deadline` restructured to only wrap subscriber (not process_inner) to avoid spurious timeout during processing | `2f30a0c2` |
| C | Added `test_sk_timeout_while_held` — exercises the Item 5 guard path | `2f30a0c2` |

ldsands:
Okay I think I got the OSL and OSM "stacking" working. I've manually confirmed that any combination of tapping and/or holding either of those in any sequence worked as intended (brief summary of changes below).

OSM + OSL Coexistence (commit 231df5cc)

The single-latch design forced mutual exclusion between OSM and OSL. Fixed by allowing pure-mod (modifiers) and layer shapes to compose — they affect different dimensions and are naturally compatible. Tap-key remains exclusive with both.

**Changes:**
- `process_sticky_pure_mod` — skip release guard when active state is a layer (compatible)
- `process_sticky_layer` — capture existing mods from prior state and carry them forward
- `update_sticky_key` — new `quick_release` arm deactivates layer before clearing state when both are active
- `resolve_explicit_modifiers` — treat layer shapes same as pure-mod (only add modifiers on press/Held)
- Tests updated for coexistence behavior

HaoboGu:
I just checked, some comments are still unresolved?

rmk/src/keyboard.rs
HaoboGu: The racing is not removed.

rmk-config/src/resolved/behavior.rs
ldsands: "Done in 34a65599 — the five flat sticky_key_* fields on Behavior are now grouped into a StickyKeyConfig struct matching the runtime one."

rmk/Cargo.toml
ldsands: Done in 42dec113 — removed the `rusty-fork` dev-dependency from Cargo.toml and unwrapped the `rusty_fork_test!{}` block in the sticky key tests. All test functions are now regular top-level `#[test]` functions. Verified with `cargo build` and `cargo test`.

ldsands:
> I just checked, some comments are still unresolved?

You're right, my bad, I'm working on resolving the last few and I'll post another comment when I think I got all of them.

Thanks for your patience!

rmk/src/keyboard/sticky_key.rs
ldsands: Done in acbe4f35 — added a guard in `release_sticky_key_if_active()` that checks if the phase is still `Pressed` (key physically held) and returns early, deferring cleanup to the Held→None transition on physical release. The unconditional `self.sticky_key_state = StickyKeyState::None` is no longer reachable while the key is pressed.

rmk/src/keyboard/sticky_key.rs
ldsands: Done in acbe4f35 — `process_sticky_pure_mod()` now refreshes the deadline on release-to-Latched, matching the existing behavior in `process_sticky_layer()`. The timeout is now correctly measured from release time for one-shot (OSM) keys.

rmk/src/host/via/keycode_convert.rs
ldsands: Done in 6385513a — keycode_convert.rs now properly maps Action::StickyKey ↔ VIA OSM/OSL ranges. Uses VIA's own 0x5280..0x529F (OSL) and 0x52A0..0x52BF (OSM) ranges — the same ranges the old OneShotLayer/OneShotModifier actions used — instead of QMK's 0x7C02/0x7C03 ranges, because VIA's ranges have no prefix-vs-payload bit overlap and all modifier combinations round-trip cleanly through the 5-bit packed bits format. Round-trip tests added for OSM(LCtrl/LShift/LAlt) and OSL(0/5). Note: the previous claim that this was done in 5ac8d57f was incorrect — that commit only added the osl!/osm! macro aliases, not the Vial conversion layer.

rmk/src/layout_macro.rs
ldsands: Done in d2d35f49 — `osl!` and `osm!` added as thin macro aliases that delegate to `sk_layer!` and `sk_mod!` respectively.

rmk/src/keyboard.rs
ldsands: Done in f7bf7903 — the `release_on_layer_change` check is now applied on layer activation too, so pressing MO(n) will release the sticky key if configured. Verified with all 519 tests passing.

rmk/src/keyboard.rs
ldsands: Finally figured out what happened this change got lost in a rebase. The commit that fixed this is [07b2ac6](https://github.com/HaoboGu/rmk/pull/859/commits/07b2ac6286a211c132f43c51ae46e013de8ce184)

ldsands:
Okay I think that is everything Not sure why that one test failed it says what I copied below so just rerunning it should work.

[warn]: Failed to get latest Xtensa Rust version: HTTP GET Error: GitHub API returned status code: 403 Forbidden

Let me know if I missed something again.

.github/workflows/user_build.yml
```diff
@@ -113,47 +111,15 @@ jobs:
...
-      - name: Upload bin artifacts
-        # espflash produces .bin images for esp32 chips instead of uf2/hex
-        if: ${{ startsWith(needs.get_chip_name.outputs.chip_name, 'esp32') }}
-        uses: actions/upload-artifact@v7
-        with:
-          name: ${{ needs.get_chip_name.outputs.artifact_prefix }}${{ needs.get_chip_name.outputs.project_name }}-firmware_bin
-          path: rmk/*.bin
-      - name: Stage ELF
-        # Expose the raw ELF(s) for debugging (probe-rs/gdb); esp32 flashes it directly.
-        # Name them like the other firmware: <name>.elf, or <name>-central/-peripheral.elf
```
HaoboGu, line 141: Revert those changes

.vscode/settings.json
```diff
@@ -57,5 +57,22 @@
...
     // "examples/use_rust/stm32h7/Cargo.toml",
-  ],
+],
+"workbench.colorCustomizations": {
+    "activityBar.activeBackground": "#fdae33",
+    "activityBar.background": "#fdae33",
+    "activityBar.foreground": "#15202b",
+    "activityBar.inactiveForeground": "#15202b99",
+    "activityBarBadge.background": "#017e4d",
+    "activityBarBadge.foreground": "#e7e7e7",
```
HaoboGu: Please revert those changes as well

HaoboGu:
That's because the unstable network? of the github action server I think. Reran it.

rmk/src/host/via/keycode_convert.rs
```diff
@@ -96,6 +87,12 @@ pub(crate) fn to_via_keycode(key_action: KeyAction) -> u16 {
                 }
             },
             Action::User(id) => (id as u16 & 0xF) | 0x7E00,
+            Action::StickyKey(sk) => {
+                match sk.layer {
+                    Some(l) => 0x5280 | (l as u16), // OSL, VIA range (same as old OneShotLayer)
+                    None => 0x52A0 | ((sk.keep.into_packed_bits() & 0x1F) as u16), // OSM, VIA range (same as old OneShotModifier)
```
HaoboGu: need to check whether the current StickyKey is `OSM/OSL`? Otherwise the info will be lost.

rmk/src/keyboard/sticky_key.rs
HaoboGu: I don't see the code

rmk/src/keyboard/sticky_key.rs
```diff
@@ -0,0 +1,443 @@
...
+            // Single mutually-exclusive latch: a tap-key press cycles (repeat_count) an existing
+            // tap-key latch, but REPLACES any other shape. Release the foreign latch first so only
+            // a same-shape tap-key latch can reach the cycle arm below.
+            if self.sticky_key_state.is_active() && !self.sticky_key_state.is_tap_key() {
+                self.release_sticky_key_if_active().await;
+            }
+
+            let mut should_deactivate = false;
+
+            match &mut self.sticky_key_state {
```
HaoboGu: If there are 2 stickykeys are pressed, the second key's state is wrong.

.github/workflows/user_build.yml
ldsands: Fixed in [3386146](https://github.com/HaoboGu/rmk/pull/859/commits/33861467ef57091add8de8f59af8174309900802)

.vscode/settings.json
ldsands: Fixed in [3386146](https://github.com/HaoboGu/rmk/pull/859/commits/33861467ef57091add8de8f59af8174309900802)

rmk/src/host/via/keycode_convert.rs
ldsands: Fixed in [1a46b791](https://github.com/HaoboGu/rmk/pull/859/commits/1a46b791).

VIA conversion now only encodes true OSM (`key == No`) and valid OSL actions. Tap-key StickyKeys are rejected instead of being converted to OSM and losing their tap-key information.

rmk/src/keyboard/sticky_key.rs
ldsands: Fixed in [a05a9942](https://github.com/HaoboGu/rmk/pull/859/commits/a05a9942).

StickyKey state now tracks the owning physical key, so a second tap-key StickyKey replaces the first cleanly and the first key’s later release does not affect the second.

ldsands:
Resolved in [301c6352](https://github.com/HaoboGu/rmk/pull/859/commits/301c6352).

Merged the latest upstream changes and resolved the combo-test conflict by retaining the upstream combo regressions while migrating the one-shot test setup to StickyKey.
Also adjusted pure-mod StickyKey releases so combo-generated press/release events from different trigger keys are handled correctly.

HaoboGu:
The overall implementation is nice, but I'm worried about the increased FLASH/RAM.

ldsands:
Size-focused follow-up in two independent commits:

- `2088d64f` compacts the sticky-key timeout state by replacing `Option<Instant>` with an `Instant::MAX` sentinel. The first Size Report improved by an average of 203.6 B of `.text` and 7.6 B of `.bss` per tracked binary versus the preceding commit.
- `5b339859` moves sticky-key configuration/deadline setup into the common dispatcher and passes the result into the OSM, OSL, and tap-key paths. This removes repeated setup logic while keeping behavior and public action/config/VIA representations unchanged.

Both changes are limited to `rmk/src/keyboard/sticky_key.rs`. Formatting and the full local CI test matrix passed before the second commit; the next Size Report will measure the second change independently.

ldsands:
Reverted `5b339859` because its Size Report regressed relative to `2088d64f`: across the 20 tracked binaries it added an average of 115.8 B of `.text` and 3.6 B of `.bss`. The earlier timeout-state compaction remains, as it produced the best measured result.

rmk-types/src/action/mod.rs
```diff
@@ -31,6 +31,22 @@ use crate::modifier::ModifierCombination;
...
 use crate::steno::StenoKey;
 
+/// Parameters for the StickyKey action.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Serialize, Deserialize, MaxSize)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[cfg_attr(feature = "rmk_protocol", derive(Schema))]
+pub struct StickyKeyAction {
+    /// Key sent on each SK press. `KeyCode::Hid(HidKeyCode::No)` selects the pure-mod (OSM) shape
+    /// when `layer` is `None`; otherwise it's the tap-key (alt-tab) shape.
+    pub key: KeyCode,
```
HaoboGu: Use `HidKeyCode` here for saving Flash/RAM. In `keyboard.rs` only Hid keycodes are processed now.

braindefender:
@ldsands 

> exit_on_layer_change (per-key) → release_on_layer_change (global).

Need some clarification on the `release_on_layer_change`. While it is useful for `SK(Key, Mod)`, it becomes quite tricky when it comes to `SK(Mod)` (OSM behavior).

In my current [ZMK keymap](https://github.com/braindefender/wellum/blob/master/README_EN.md#how-one-shot-sticky-modifiers-work), I use similar to `activate_on_keypress` and release OSM's via layer change tap (`MO(1)` or else).

«Layer change» can happen multiple ways. For example, if you press `MO(1)`, then tap`SK(Mod)`, and then release `MO(1)` — it changes layer (back to layer 0). 

In this case, if the `release_on_layer_change` is `true`, will it release `Mod`? And if not, will it release it on next `MO(1)` press?

Or `SK(Mod)` completely ignores `release_on_layer_change` and there is no way to release OSM except for a timeout or by pressing any normal key (therefore, triggering unwanted shortcut)?

rmk-types/src/action/mod.rs
ldsands: Implemented in 603ad068. `StickyKeyAction.key` now uses `HidKeyCode` directly, and I updated the sticky-key runtime, layout macros, and VIA conversion paths accordingly. Verified with: `rmk-types` tests (19 passed), sticky-key nextest suite (20 passed), VIA sticky-key conversion tests (3 passed), formatting, and `cargo check --no-default-features`.

ldsands:
@braindefender Good catch—the global `release_on_layer_change` also affects the OSM and OSL shapes. I added optional per-shape overrides:

```toml
[behavior.sticky_key]
release_on_layer_change = false
tap_key_release_on_layer_change = true
one_shot_mod_release_on_layer_change = false
layer_release_on_layer_change = false
```

braindefender:
> @braindefender Good catch—the global `release_on_layer_change` also affects the OSM and OSL shapes. I added optional per-shape overrides:
> 
> ```toml
> [behavior.sticky_key]
> release_on_layer_change = false
> tap_key_release_on_layer_change = true
> one_shot_mod_release_on_layer_change = false
> layer_release_on_layer_change = false
> ```

It is still not very clear what exactly is `layer_change`?
If I enable `one_shot_mod_release_on_layer_change` will it be impossible to keep sticky mods between layers?

`Case #1` (plain OSSM, activate_on_keypress: true):
- press `MO(1)` → tap `OSM(LCtrl)` → release `MO(1)` → tap letter `C`
=>
- `LCtrl+C`

`Case #2` (release OSSM with `MO(1)`, activate_on_keypress: true):
- press `MO(1)` → tap `OSM(LCtrl)` → release `MO(1)`
=>
- `LCtrl` is held
then
- tap `MO(1)`
=> 
- `LCtrl` is released

If I enable `one_shot_mod_release_on_layer_change`, will it break both `Case #1` and `Case #2` because "release `MO(1)`" is a "layer_change"?

I don't see the case where OSM should be released on "layer change". Only in "layer enter". So, maybe we change `one_shot_mod_release_on_layer_change` to `one_shot_mod_release_on_layer_enter` then?

rmk-types/src/action/mod.rs
HaoboGu: Now it's much better

HaoboGu:
@codex review

ldsands:
Thanks @braindefender for your feedback it is very helpful!

So I use my keyboard in the way shown below fairly often where I'll tap an OSM(<mod>) then hold my MO(1) then tap a key so I want it to survive that first transition then be consumed by the key in the other layer.

So I toyed with the idea of doing OSM releasing on layer_exit rather then layer_enter.

Below is the difference between what I'm used to (top the one_shot_mod_release_on_layer_exit) vs your proposed (one_shot_mod_release_on_layer_enter) for your two cases and a couple cases of my own.

```
Case                                                         Policy       Setting  Outcome
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━  ━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
braindefender 1: MO(1) → OSM(Ctrl) → release MO(1) → C       layer_enter  false    Ctrl+C
                                                             layer_enter  true     Ctrl+C
                                                             layer_exit   false    Ctrl+C
                                                             layer_exit   true     C
───────────────────────────────────────────────────────────  ───────────  ───────  ─────────────────────────────────────────────
braindefender 2: MO(1) → OSM(Ctrl) → release MO(1) → MO(1)   layer_enter  false    Ctrl remains held after both transitions
                                                             layer_enter  true     Ctrl is released when MO(1) is pressed again
                                                             layer_exit   false    Ctrl remains held after both transitions
                                                             layer_exit   true     Ctrl is released on the first MO(1) release
───────────────────────────────────────────────────────────  ───────────  ───────  ─────────────────────────────────────────────
ldsands 1: OSM(Ctrl) → MO(1) → 1                             layer_enter  false    Ctrl+1
                                                             layer_enter  true     1
                                                             layer_exit   false    Ctrl+1
                                                             layer_exit   true     Ctrl+1
───────────────────────────────────────────────────────────  ───────────  ───────  ─────────────────────────────────────────────
ldsands 2: OSM(Ctrl) → MO(1) → release MO(1) → T             layer_enter  false    Ctrl+T
                                                             layer_enter  true     T
                                                             layer_exit   false    Ctrl+T
                                                             layer_exit   true     T
```

I'm not opposed to doing both but for the backend, especially with the RAM and FLASH concerns I don't know if this is a good idea and I haven't thought at all about both simultaneously would do. Are you partial to the on layer enter over exit? I suppose we could also have a layer_change be set to on_enter or on_exit or true for both, maybe that is the better option?

What are your thoughts?

(Obviously if @HaoboGu has any thoughts, please weigh in)

rmk/src/keyboard/sticky_key.rs
```diff
@@ -0,0 +1,584 @@
...
+                StickyKeyState::Active {
+                    layer: Some(prev_layer),
+                    phase,
+                    mods,
+                    ..
+                } => {
+                    self.keymap.deactivate_layer(prev_layer);
+                    (phase, mods)
+                }
+                StickyKeyState::Active { mods, phase, .. } => (phase, mods),
```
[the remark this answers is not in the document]
ldsands: Fixed in 3c2a0448. A layer-shaped StickyKey now releases a held tap-key before taking over the shared latch; added a regression test for the displaced-key sequence.

rmk/src/keyboard/sticky_key.rs
```diff
@@ -0,0 +1,584 @@
...
+        // only happen when the key is held and idle. For layer and tap-key shapes, the
+        // deadline fires in the same scenario.
+        // Clear the deadline to avoid busy-looping on every iteration.
+        if matches!(self.sticky_key_state, StickyKeyState::Active { pressed: true, .. }) {
+            debug!(
+                "StickyKey timeout fired while key is still held — clearing deadline, deferring to physical release"
+            );
+            if let StickyKeyState::Active { deadline, .. } = &mut self.sticky_key_state {
+                deadline.clear();
+            }
```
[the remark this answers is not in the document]
ldsands: Fixed in 3c2a0448. A tap-key now re-arms its timeout on physical release when that deadline fired while it was held; added regression coverage.

HaoboGu:
Yeah I know what the issue is -- if we do the per-key config, the Flash/RAM consumption bloats a lot, because we have to reserve the space for ALL keys which are possible to set to `SK`. 

Another question is the configurations. TBH I'm not a fan to make the configurations`SK Mod/SK Layer` only. I think the default configuration should be applied to all SKs, and for each SK, we can override it. And this hits the Flash/RAM issue again.

My current idea is to have a list of `profile`s, like #958.

In this case, the configuration can be simplified like:

```toml
[behavior.sticky_key.profiles]
p = {
  timeout = "1s",
  activate_on_keypress = true,
  max_repeat = 10,
  release_mode = "other_key_press | layer_exit | layer_enter | etc",
}
[behavior.sticky_key] # default
timeout = "1s",
activate_on_keypress = false,
max_repeat = 0,
release_mode = "other_key_press | layer_exit | layer_enter | etc",
```

We can have multiple release mode in different profiles.

Then in keymap, we can use `SK(mod, @p)` to specify the profile and if the profile is not specific, the default one is used.

This is convenient for keyboard.toml users, for Rust API users, they can just use the id of the profile maybe?

braindefender:
Yeah, profiles look good, customizable and right in the middle between global config and per-key config (in terms of memory footprint too).

ldsands:
> Yeah I know what the issue is -- if we do the per-key config, the Flash/RAM consumption bloats a lot, because we have to reserve the space for ALL keys which are possible to set to `SK`.
> 
> Another question is the configurations. TBH I'm not a fan to make the configurations`SK Mod/SK Layer` only. I think the default configuration should be applied to all SKs, and for each SK, we can override it. And this hits the Flash/RAM issue again.
> 
> My current idea is to have a list of `profile`s, like #958.
> 
> In this case, the configuration can be simplified like:
> 
> ```toml
> [behavior.sticky_key.profiles]
> p = {
>   timeout = "1s",
>   activate_on_keypress = true,
>   max_repeat = 10,
>   release_mode = "other_key_press | layer_exit | layer_enter | etc",
> }
> [behavior.sticky_key] # default
> timeout = "1s",
> activate_on_keypress = false,
> max_repeat = 0,
> release_mode = "other_key_press | layer_exit | layer_enter | etc",
> ```
> 
> We can have multiple release mode in different profiles.
> 
> Then in keymap, we can use `SK(mod, @p)` to specify the profile and if the profile is not specific, the default one is used.
> 
> This is convenient for keyboard.toml users, for Rust API users, they can just use the id of the profile maybe?

Thanks for the feedback, I agree this seems like a much better way to go. I'll get started and hopefully I'll be able to finish it today or tomorrow.

docs/docs/main/docs/configuration/behavior.md
```diff
@@ -34,51 +31,107 @@ In this example, when both layers 1 (`upper`) and 2 (`lower`) are active, layer
...
+For example, an Alt+Tab profile can release on another key press or either
+direction of a layer transition:
+
+```toml
+[behavior.sticky_key]
+timeout = "1s"
+
+[behavior.sticky_key.profiles.alt_tab]
+timeout = "5s"
+release_mode = "other_key_press | layer_enter | layer_exit"
```
HaoboGu: Add a "double tap to release" mode?

rmk-config/src/resolved/behavior.rs
```diff
@@ -1,14 +1,61 @@
...
+pub struct StickyKeyConfig {
+    pub timeout_ms: Option<u64>,
+    pub activate_on_keypress: Option<bool>,
+    pub max_repeat: Option<u16>,
+    pub release_mode: Option<StickyKeyReleaseMode>,
+    pub profiles: HashMap<String, StickyKeyProfile>,
+}
+
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub struct StickyKeyReleaseMode(pub u8);
```
HaoboGu: Use `bitfield` for release mode? Just like `ModifierCombination`

rmk/src/keyboard/sticky_key.rs
```diff
@@ -0,0 +1,656 @@
...
+use rmk_types::action::StickyKeyAction;
+use rmk_types::keycode::HidKeyCode;
+use rmk_types::modifier::ModifierCombination;
+
+use crate::config::StickyKeyReleaseMode;
+use crate::event::{KeyboardEvent, KeyboardEventPos};
+use crate::keyboard::Keyboard;
+use crate::keymap::StickyKeyShape;
+
+static STICKY_LAYER_EVENT_CHANNEL: Channel<crate::RawMutex, StickyKeyReleaseMode, 4> = Channel::new();
```
HaoboGu: This channel is not a preferred way.

rmk/src/keyboard/sticky_key.rs
```diff
@@ -0,0 +1,656 @@
...
+        profile: u8,
+        phase: SkPhase,
+        /// Whether the physical StickyKey switch is currently held down.
+        pressed: bool,
+        repeat_count: u16,
+        deadline: StickyKeyDeadline,
+    },
+}
+
+#[cfg(test)]
```
HaoboGu: Don't place this test block here?

rmk/src/keyboard/sticky_key.rs
```diff
@@ -0,0 +1,656 @@
...
+    /// modifier until the SK is released. OSM `Held`.
+    Held,
+}
+
+/// An optional deadline stored without `Option<Instant>`'s extra discriminant.
+///
+/// `Duration::MAX` already means "no timeout" in sticky-key configuration, so
+/// `Instant::MAX` is reserved as the inactive sentinel here.
+#[derive(Clone, Copy, Debug)]
+pub(crate) struct StickyKeyDeadline(Instant);
```
HaoboGu: Remove this, `Instant` is fine.

rmk/src/keyboard/sticky_key.rs
```diff
@@ -0,0 +1,656 @@
...
+        /// Physical key that owns this latch.
+        source: KeyboardEventPos,
+        mods: ModifierCombination,
+        /// `HidKeyCode::No` = pure-mod or layer shape; any other key = tap-key shape.
+        key: HidKeyCode,
+        /// `Some(n)` = OSL shape; `None` = pure-mod or tap-key shape.
+        layer: Option<u8>,
+        /// Selected Sticky Key profile (`u8::MAX` means default profile).
+        profile: u8,
+        phase: SkPhase,
```
HaoboGu: What's the difference between SkPhase and StickyKeyState? It's kind of confusing

rmk/src/keyboard/sticky_key.rs
```diff
@@ -0,0 +1,656 @@
...
+                _ => {}
+            }
+        }
+    }
+
+    /// Layer (OSL) shape: activate the layer for the next foreign key. Mirrors the former
+    /// `process_action_osl`. The layer carries no modifier, so consuming it emits no HID
+    /// report — the foreign key resolves on the active layer in `process_action_key` before
+    /// the latch is consumed.
+    async fn process_sticky_layer(&mut self, params: StickyKeyAction, event: KeyboardEvent) {
```
HaoboGu: Why the sticky mod and sticky layer are processed differently?

HaoboGu:
Did a rough review, there are lots of design/implementation smells in the PR. I tagged some of them. The overall direction is good and I really like current behavior design, but the implementation is not very clean. Could you please refactor it to make it better?

ldsands:
> Did a rough review, there are lots of design/implementation smells in the PR. I tagged some of them. The overall direction is good and I really like current behavior design, but the implementation is not very clean. Could you please refactor it to make it better?

I'll do my best, thanks!

rmk/src/keyboard/sticky_key.rs
ldsands: Removed the deadline wrapper; the state now stores Option<Instant> directly.

rmk/src/keyboard/sticky_key.rs
ldsands: Replaced the private channel with the typed StickyKeyReleaseEvent.

rmk/src/keyboard/sticky_key.rs
ldsands: Removed the inline size test.

rmk/src/keyboard/sticky_key.rs
ldsands: Removed SkPhase. StickyKeyState now directly represents None, Pressed, Latched, and Held while carrying the shared active data.

docs/docs/main/docs/configuration/behavior.md
ldsands: Added double_tap as an opt-in release mode, with tests for modifier, layer, and tap-key shapes.

rmk/src/keyboard/sticky_key.rs
ldsands: Consolidated the shared lifecycle and release transitions. An explicit effect enum now keeps only the actual modifier, layer, and tap-key side effects separate.

rmk-config/src/resolved/behavior.rs
ldsands: Changed StickyKeyReleaseMode to use bitfield-struct.

ldsands:
Okay I think that this implementation is much better (at least from a code smell perspective). I've done some of my normal manual testing in part using my own keyboard and I've been running it with profiles since yesterday afternoon (before the conflicts that came up) and I haven't had any issues as far as I can tell it works as intended.

Again, thanks for your patience with me on this pr. I have never done anything else in rust and I'm much more of a data scientist and thus my training is really limited in this area.

Let me know what changes (if any 🤞) are needed.

HaoboGu:
Thanks a lot! The current implementation looks great! I'll review it today and I'm really excited to get it merged soon. Thanks again for your effort on this.

rmk/src/keyboard.rs
```diff
@@ -1235,27 +1315,46 @@ impl<'a> Keyboard<'a> {
                 KeyCode::Consumer(consumer) => {
                     self.process_action_consumer_control(consumer, event).await;
                     self.update_osm(event);
-                    self.update_osl(event);
+                    self.update_osl(event).await;
```
HaoboGu: The old osl/osm implementation should be removed?

And the vial converter should be fixed as well

rmk/src/config/behavior.rs
```diff
@@ -148,6 +149,63 @@ impl Default for MorsesConfig {
     }
 }
 
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub struct StickyKeyReleaseMode(u8);
+
+impl StickyKeyReleaseMode {
```
HaoboGu: This should be moved to rmk-types

rmk-config/src/resolved/behavior.rs
```diff
@@ -1,5 +1,58 @@
 use std::collections::HashMap;
 
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub struct StickyKeyReleaseMode(u8);
+
+impl StickyKeyReleaseMode {
```
HaoboGu, line 12: In rmk-config, this can be a struct, not a compressed u8. Because rmk-config is used only in keyboard.toml parsing. rmk-macro converts it to compressed types in rmk-types

rmk/src/event/state.rs
```diff
@@ -17,6 +20,56 @@ impl LayerChangeEvent {
...
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(crate) enum LayerTransition {
+    Enter,
+    Exit,
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(crate) struct LayerTransitionGeneration(u64);
+
+static LAYER_TRANSITION_GENERATION: Mutex<crate::RawMutex, Cell<u64>> = Mutex::new(Cell::new(0));
```
HaoboGu: I think the `LayerTransitionGeneration`, `LayerTransitionEvent`, and all auto mouse layer changes should be removed. Just process them as normal layer change is fine

rmk/src/keymap.rs
```diff
@@ -29,6 +46,8 @@ pub struct KeymapData<const ROW: usize, const COL: usize, const NUM_LAYER: usize
     pub(crate) encoder_map: [[EncoderAction; NUM_ENCODER]; NUM_LAYER],
     /// Per-layer activation flags
     layer_state: [bool; NUM_LAYER],
+    /// Per-layer mutation generations used to avoid releasing another owner's layer.
+    layer_generation: [u64; NUM_LAYER],
```
HaoboGu: It should be removed too. There's already `layte_state`. I think the layer trigger source shouldn't be counted, the simplest solution is: the latest command counts. That avoid many many complexities. ZMK's layer is a bitmap, ZMK doesn't count the layer generation as well

rmk-types/src/action/mod.rs
HaoboGu: OneShot actions can be removed now

rmk-types/src/action/mod.rs
```diff
@@ -29,6 +29,34 @@ use crate::modifier::ModifierCombination;
 #[cfg(feature = "steno")]
 use crate::steno::StenoKey;
 
+/// Effect produced by a sticky-key action.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Serialize, Deserialize, MaxSize)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+#[cfg_attr(feature = "wasm", derive(tsify::Tsify))]
+#[cfg_attr(feature = "wasm", tsify(into_wasm_abi, from_wasm_abi))]
+pub enum StickyKeyEffect {
```
HaoboGu: I think the StickyKey should not be an `Action`, it should be added to `KeyAction` as a variant:

```rust
enum KeyAction {
  ..
  // Sticky activate `Action`, the u8 is the profile number.
  Sticky(Action, u8)
}
```

In this case:

- `OSM`: `Sticky(Action::Modifier(x), profile_num)`
- `OSL`: `Sticky(Action::LayerOn(1), profile_num)`
- `SM(Tab, LAlt)`: `Sticky(Action::KeyWithModifier(Tab, LAlt), profile_num)`

I think it's cleaner.

HaoboGu (commented):
Overall this version is much better 👍. 

I think it can be even cleaner by removing layer_generations/layer_transition_generation/event, making this PR more focused

rmk/src/keyboard.rs
ldsands: Done — I removed the legacy OneShot runtime/action path, and OSL/OSM now use `KeyAction::Sticky`. I also updated the Vial converter and added coverage for compatible OSL/OSM encodings.

rmk/src/config/behavior.rs
ldsands: Done — `StickyKeyReleaseMode` now lives in `rmk-types::sticky_key`. RMK’s config module only re-exports it for API compatibility.

rmk-config/src/resolved/behavior.rs
ldsands: Done — rmk-config now represents release mode as a normal bool-field struct. rmk-macro performs the compact conversion only when generating the rmk-types value.

rmk/src/event/state.rs
ldsands: Done — I removed `LayerTransitionGeneration`, `LayerTransitionEvent`, their configuration, and the associated auto-mouse publishing. Sticky cleanup now uses normal layer changes.

rmk/src/keymap.rs
ldsands: Done — I removed the layer-generation storage and APIs. Sticky layers now use normal boolean layer state, so the latest activation or deactivation command wins.

rmk-types/src/action/mod.rs
ldsands: Done — the OneShot `Action` variants and old `keyboard/oneshot.rs` state machine have been removed. OSL/OSM syntax and legacy settings remain only as compatibility inputs to Sticky.

rmk-types/src/action/mod.rs
ldsands: Done — Sticky is now `KeyAction::Sticky(Action, u8)`. OSM, OSL, and SM map to `Action::Modifier`, `Action::LayerOn`, and `Action::KeyWithModifier`, respectively.

ldsands:
I removed the layer generation/transition machinery and unified the old one-shot paths under `KeyAction::Sticky`, substantially reducing the implementation scope. The full RMK feature matrices and focused Sticky, OSL/OSM, and Vial tests pass.

ldsands:
Maybe I should have just created a separate pr for this but I won't be around for a couple weeks starting in a couple of days. I realized that after I used OSM keys (e.g. ctrl+shift+arrow the timeout would start as I released those keys but then it causes issues when using the mouse or other keys right after I'm done with holding any of these OSM keys. Below is the explanation for what I've added.

Added an optional Sticky Key behavior, release_on_keyup_after_timeout.

When enabled, holding an OSM or OSL beyond its configured timeout and then releasing it immediately releases the modifier or layer. Previously, releasing the key would start a new timeout, which could leave modifiers or layers unexpectedly active.

[behavior.sticky_key]
release_on_keyup_after_timeout = true

The option defaults to false for backward compatibility and does not affect modified tap keys. Tests cover OSM and OSL behavior, default compatibility, and timeout/key-up ordering.

ldsands:
I made a follow-up change to `release_on_keyup_after_timeout` that in actual use I liked better.

Using the normal Sticky Key timeout as the hold threshold turned out to be too restrictive. I still need a relatively long timeout for tapping and stacking OSM/OSL keys, but
I want a physically held modifier to stop immediately when I release it—even if it was held for less than that normal timeout.

I replaced the boolean option with an optional, independently configurable duration:

```toml
[behavior.sticky_key]
timeout = "1s"
release_on_keyup_after = "300ms"
```

The resulting behavior is:

- If the Sticky Key is released before `300ms`, it is treated as a tap and receives the complete normal `1s` timeout starting from key-up.
- If it is held for at least `300ms`, releasing it immediately releases the modifier or layer.
- Omitting `release_on_keyup_after` preserves the existing behavior.
- Modified tap keys are still excluded from this behavior.

The primary use case is holding an OSM such as Ctrl, clicking a link with the mouse, and then releasing Ctrl. RMK cannot observe the external mouse click, so physical hold
duration is used to distinguish this from a tap. Separating the hold threshold from the normal timeout lets Ctrl release promptly without shortening the timeout needed for
tapping or stacking Sticky Keys.

The implementation also preserves the original physical press time when combo processing delays dispatch and handles overlapping Sticky Modifier producers consistently. I
added coverage for short taps, the exact threshold boundary, independent timeout/hold-threshold durations, timeout/event ordering, combo-buffered input, overlapping release
orders, and mixed profiles.

rmk/src/keyboard.rs
```diff
@@ -325,37 +335,51 @@ impl<'a> Keyboard<'a> {
...
+                            self.handle_morse_timeout(&key).await;
+                        }
                     }
                 }
             }
             _ => (),
         }
     }
 
+    fn runtime_deadline(&self, buffered_deadline: Instant) -> Instant {
```
HaoboGu, line 375: Why the sticky key state is checked in so many different places? I feel that this is not correct.

rmk/src/keyboard/sticky_key.rs
```diff
@@ -0,0 +1,1141 @@
...
+    pub(crate) modifier_consumed: bool,
+    pub(crate) modifier_was_host_visible: bool,
+}
+
+/// Runtime composition for sticky effects.
+///
+/// Modifier and layer effects can coexist and therefore own distinct policies,
+/// phases, sources, and deadlines. A tap key is exclusive with both.
+#[derive(Clone, Copy, Debug, Default)]
+pub(crate) struct StickyKeyState {
```
HaoboGu, line 368: I think sticky key should be general enough to process all held `Action`s. Distinguishing between modifier/layer/keys makes the current implementation not general at all. That means, this state should not be the current shape.

HaoboGu (commented):
For the `release_on_keyup_after`, I feel it's more like a Morse/Tap-Hold behavior, not Sticky Key's. But since we've promoted the StickyKey as a `KeyAction`, I don't know if there's other solutions, so your code might be right and the only way to achieve this. 

I prefer to have a name like `release_after_hold` and set a longer threshold?

ldsands:
> For the `release_on_keyup_after`, I feel it's more like a Morse/Tap-Hold behavior, not Sticky Key's. But since we've promoted the StickyKey as a `KeyAction`, I don't know if there's other solutions, so your code might be right and the only way to achieve this.
> 
> I prefer to have a name like `release_after_hold` and set a longer threshold?

Works for me.

- Renamed release_on_keyup_after to release_after_hold throughout.
- Changed the documented example threshold from 300ms to 500ms.
- Kept the threshold optional and disabled by default.

rmk/src/keyboard.rs
ldsands: Thanks — I consolidated this in `03eb1a91`. `keyboard.rs` no longer inspects separate modifier, layer, or tap-key sticky states. It now interacts with sticky keys only through shape-independent integration hooks for deadlines, buffered claims, modifier report contribution, and the prepare/finish boundaries around action dispatch. The sticky lifecycle and active-entry inspection are owned by `sticky_key.rs`. The remaining call sites correspond to distinct keyboard processing boundaries rather than separate implementations of each sticky-key type. I also added focused behavioral coverage in `46673893`.

rmk/src/keyboard/sticky_key.rs
ldsands: Thanks — I reworked the state around that model in `03eb1a91`. `StickyKeyState` now owns a uniform bounded collection of `StickyEntry` values, and every entry stores its canonical held `Action` together with the shared phase, policy, deadline, repetition, and buffered-claim lifecycle. OSM, OSL, and modified tap keys no longer have separate top-level state slots or lifecycle types.

Action-specific handling is now limited to applying and releasing the concrete effect. The small amount of modifier-only metadata remains because modifiers can have multiple physical producers and must remember whether their HID effect has already been reported; layer and tap-key effects are represented completely by their stored `Action`. The OSM and OSL aliases remain unchanged. `46673893` adds parity and edge-case coverage to ensure the consolidation does not remove existing behavior.

HaoboGu: sorry maybe I didn't make it clear, the current code still has modifier/layer/tap-key processing logic in sticky key, which leads to the binary size bloats -- these actions' processing logic is duplicate in normal path and sticky key path. 

Actually sticky key itself should not execute/classify any Action variant, all "effect" should be removed. Only a thin state like sticky: (Action, u8) is needed.

braindefender:
@HaoboGu @ldsands What is the current status of this PR?

ldsands:
> @HaoboGu @ldsands What is the current status of this PR?

Hi @braindefender

Sorry, I've been busy with a conference, getting ready to move soon, starting a new job and more. It has been a crazy month. Thus I've been unable to keep up on this PR.

I resolved all the conflicts and now I'm working through anything that code reviews bring up (better models mean more identified potential issues) and I'm working through them right now. I'll also see if I cannot reduce the flash/ram usage as well one more time. Once they're all done (hopefully today) I'll tag you and HaoboGu to let you know that I think I got everything.

Also, as I've been using this branch daily since July I did come across another minor issue that does come up on occasion. I fixed the issue and now (confirmed on my keyboard).

For the record, the issue I mentioned above is explained with an example below:

If I hold OSM(ctrl)+OSM(shift)+leftarrow (again just an example) and then I lift off of OSM(ctrl) and continue to hold OSM(shift)+leftarrow the OSM(ctrl) is still being sent. To get the desired behavior, I must release all keys on the keyboard before I can use OSM(shift)+leftarrow correctly. Also, note that I can lift up on the leftarrow while holding down the OSM(shift), but the keyboard is still sending the control and the shift simultaneously. If I left up on the OSM(ctrl) while still holding any other key. I want it to just send only the keys (be they OSM or not) that I'm holding down. I also want when I release that, OSM(ctrl) for the keyboard to immediately stop sending that OSM(ctrl) to the OS but continue sending the OSM(shift) which I am holding.

ldsands:
@HaoboGu 

Okay, I think I've gone about as far as I can but in one of the final reviews of the code codex found an issue with Auto Mouse Layer below is the small summary of the issue along with the recommended fix after exploring a few different possibilities.

> Auto Mouse Layer currently sees SK(LCtrl) as an ordinary Ctrl action, so it can incorrectly deactivate the mouse layer when deactivate_on_key is enabled. A possible fix is to retain a small “came from Sticky” marker in the action event, allowing Auto Mouse Layer to keep the layer active and reset its timeout as documented. This would be a small change, but may slightly increase event size and affect dongle serialization.

I've tried as much as possible (though not always successfully) not to touch any other features, so I thought I would ask on this one. I've committed the fix that it came up with, but I'll undo it and tackle it some other way if you think there is a better approach or if you think we can just leave it as is.

Other recent changes (mostly just to deal with merges, or to address failed tests) as summarized by codex.

- Added immediate live release for held Sticky modifiers.
- Fixed Sticky combo buffering, identity, ownership, and claim timing.
- Preserved modifier ownership across Caps Word and other interactions.
- Added compile-time elimination of unused Sticky handlers to reduce firmware flash.
- Strengthened Sticky configuration, capacity documentation, and tests.

Let me know if you need anything else.

HaoboGu:
Thanks @ldsands ！

I have an idea about sharing state between processors. Would you mind me editing code in this PR?

ldsands:
> Thanks @ldsands ！
> 
> I have an idea about sharing state between processors. Would you mind me editing code in this PR?

@HaoboGu Absolutely feel free.

edit: I think I had said something about this before, but I am mostly self-taught in Python for data science. I know enough to be able to muddle my way through other programming languages and now with AI where it is, I'm able to make contributions that work, but as you have clearly observed, my ability to do programming of this kind "well" is limited. Thanks again for your patience with me on this pr.

PR #864, opened 2026-05-27
feat: embed version and git hash in USB serial number

jpds:
This PR changes the serial number from the static string (`vial:f64c2b3c:000001`) and replaces it with a build-time stamp of the form `rmk:<version>-<git-hash>` (with or without `vial:...` based on feature flags). Allowing users to know what they have flashed on their devices.

Vial's `util.py` looks for the string, anywhere in the serial. 

Downstream firmware wanting to append their own build info can use `const_format::concatcp!(rmk::config::RMK_BUILD_INFO, ";my-fw:", ...)`

### DIFF
--- a/docs/docs/main/docs/configuration/keyboard_device.mdx
+++ b/docs/docs/main/docs/configuration/keyboard_device.mdx
@@ -20,7 +20,7 @@ vendor_id = 0x4c4b                # USB vendor ID (16-bit hex)
 product_id = 0x4643               # USB product ID (16-bit hex)
 manufacturer = "YourName"         # (Optional) Manufacturer string
 product_name = "KeyboardName"     # (Optional) Keyboard name, if not set, it will be same as `name` field
-serial_number = "vial:f64c2b3c"   # (Optional) Serial number of the keyboard
+serial_number = "my-firmware-v1"  # (Optional) Serial number; defaults to rmk:<version> (with a vial: prefix for Vial builds)
 chip = "nrf52840"                 # Target microcontroller
 usb_enable = true                 # (Optional) Enable USB functionality. If omitted, RMK uses the board/chip default.
 ```
@@ -29,20 +29,22 @@ usb_enable = true                 # (Optional) Enable USB functionality. If omit
 <Tab label={<Rust />}>
 
 ```rust title="main.rs"
+use rmk::config::{DeviceConfig, RmkConfig};
 
 let keyboard_device_config = DeviceConfig {
     vid: 0x4c4b,
     pid: 0x4643,
     manufacturer: "Haobo",
     product_name: "RMK Keyboard",
-    serial_number: "vial:f64c2b3c:000001",
+    // Omit serial_number to use RMK's default version stamp (Vial builds get the
+    // vial: marker automatically).
+    ..Default::default()
 };
 
 let rmk_config = RmkConfig {
     device_config: keyboard_device_config,
     // .. Other configurations
 };
-
 ```
 
 </Tab>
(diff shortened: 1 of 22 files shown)


HaoboGu:
iirc Vial needs `vial:f64c2b3c` to work, could you double check whether it works on both web and Vial app?

jpds:
@HaoboGu I had only checked the web UI which worked fine. Adjusted now.

docs/docs/main/docs/configuration/keyboard_device.mdx
```diff
@@ -27,28 +27,33 @@ usb_enable = true                 # (Optional) Enable USB functionality. If omit
 <Tab label={<Rust />}>
 
 ```rust title="main.rs"
+use const_format::concatcp;
```
HaoboGu: This introduces extra dependency, I think it's unnecessary?

docs/docs/main/docs/configuration/keyboard_device.mdx
```diff
@@ -27,28 +27,33 @@ usb_enable = true                 # (Optional) Enable USB functionality. If omit
...
 ```
 
 </Tab>
 </Tabs>
 
 ::: info
 
-To make your keyboard to be recognized by Vial, the `serial_number` shall start with `vial:f64c2b3c:`
+If `serial_number` is omitted, RMK embeds a version stamp automatically. For Vial builds (the
+default), this is `rmk:<version>-<git-hash>;vial:f64c2b3c`, which lets Vial discover the
```
HaoboGu: I prefer to have `vial:f64c2b3c` at the beginning(or  because it might be dropped in BLE. There's length limitation in BLE's serial.

rmk/src/config/device.rs
```diff
@@ -13,14 +13,33 @@ pub struct DeviceConfig<'a> {
...
+///
+/// When the `vial` feature is enabled the string also includes `;vial:f64c2b3c` so that
+/// vial-gui can discover the device without requiring a hardcoded serial number override.
+///
+/// Downstream firmware can extend this with their own build info using
+/// `const_format::concatcp!(rmk::RMK_BUILD_INFO, ";my-firmware:", env!("CARGO_PKG_VERSION"), "-", env!("MY_GIT_HASH"))`.
+#[cfg(feature = "vial")]
+pub const RMK_BUILD_INFO: &str = concat!(
+    "rmk:",
+    env!("CARGO_PKG_VERSION"),
```
HaoboGu: There are several thoughts about the BUILD_INFO, I think it can be improved to cover:

1. When using RMK version from git or local path, the `CARGO_PKG_VERSION` will point to the version in `Cargo.toml` of used RMK, which is not accurate.
2. When the workspace is dirty(for example, using a local RMK with edits), the RMK_GIT_HASH is not accurate too.
3. When using crates.io version, `CARGO_PKG_VERSION` and `RMK_GIT_HASH` are always fixed -- that means we can remove either of them.

rmk/build.rs
```diff
@@ -11,6 +11,11 @@ fn main() {
     common::set_target_cfgs(&mut cfgs);
 
     println!("cargo:rerun-if-changed=build.rs");
+    println!("cargo:rerun-if-changed=.git/HEAD");
```
HaoboGu: This means ALWAYS rebuild RMK in user space -- which is not correct. This is because that there's no `.git/HEAD` under `rmk` folder. 

And I don't think use this to monitoring commit changes is correct, `.git/HEAD` won't change when committing code.

jpds: Right, I had only tested this with my local copy which is all Git based. Simplified it to the crate version.

HaoboGu:
Thanks!

PR #866, opened 2026-05-29
fix(osm): keep OSM/OSL timeout alive across non-consuming events

yekingyan:
## Problem

  The OSM/OSL release path waited with a single `select(timeout, next_event)`.
  The first event arriving during the timeout cancelled the wait permanently,
  so the one-shot got stuck in `Single` and never auto-released. A typical
  trigger is the release event of a layer key (a non-consuming event).

  Additionally, when a consuming event was already sitting in `unprocessed_events`
  (dequeued from the channel before entering the wait), the subscriber would
  never deliver it again, causing the OSM to hang indefinitely (e.g. fast
  Ctrl+Shift+P sequences).

  ## Fix

  Replace the single `select` with a deadline loop: non-consuming events are
  pushed to the queue and we keep waiting for the remaining time, while a
  consuming event breaks out (OSM: press in quick_release mode, release
  otherwise; OSL: press).

  Before entering the loop, check whether `unprocessed_events` already contains
  a consuming event and skip the await entirely if so.

  ## Test

  Added `test_osm_timeout_not_cancelled_by_non_consuming_event`: an OSL press
  interrupts the OSM timeout; after the timeout elapses the OSM must have
  auto-released, so the later layer-1 key is sent without the modifier.
  Red before the fix, green after.

### DIFF
--- a/rmk/src/keyboard/oneshot.rs
+++ b/rmk/src/keyboard/oneshot.rs
@@ -1,5 +1,5 @@
 use embassy_futures::select::{Either, select};
-use embassy_time::Timer;
+use embassy_time::{Instant, Timer};
 use rmk_types::modifier::ModifierCombination;
 
 use crate::event::KeyboardEvent;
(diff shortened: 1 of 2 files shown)


rmk/tests/keyboard_one_shot_test.rs
```diff
@@ -658,8 +658,40 @@ mod one_shot_test {
         };
     }
 
-    // TODO: test_osm_quick_release_rolling removed — OSM + morse/tap-hold interaction
-    // has a known bug where the OSM deadline loop times out before the tap resolves.
+    // NOTE: test_osm_quick_release_rolling is intentionally omitted. With a
+    // tap-hold key the OSM modifier is cleared before the buffered morse tap
+    // resolves, so the tapped key is emitted without the modifier. That is a
+    // separate OSM + morse/tap-hold timing issue, unrelated to the timeout
+    // fix in this change.
```
jpds: I feel like "temporally bounded" LLM comments like this should be removed. It's referring to "this change" but - 50 commits and code changes later - I'm not going to know what that change is without having to dig it up.

jpds (changes requested):

rmk/tests/keyboard_one_shot_test.rs
yekingyan: Thanks for the review! Removed the temporal reference in the comment.

Also rebased on latest main and fixed a related edge case: when a consuming event is already in `unprocessed_events` before entering the await loop, the OSM would hang forever. Added a pre-check to skip the loop in that case.

jpds (approved):

rmk/src/keyboard/oneshot.rs
```diff
@@ -136,17 +162,33 @@ impl<'a> Keyboard<'a> {
...
                         }
-                        Either::Second(e) => {
-                            // New event, send it to queue
-                            if self.unprocessed_events.push(e).is_err() {
-                                warn!("Unprocessed event queue is full, dropping event");
+                        let timeout = Timer::after(deadline - now);
+                        match select(timeout, self.keyboard_event_subscriber.next_message_pure()).await {
+                            Either::First(_) => {
+                                // Timeout, deactivate layer
+                                self.keymap.deactivate_layer(layer_num);
```
HaoboGu: Should `layer_num` be `l` in `OneShotState::Initial(l) | OneShotState::Single(l)`?

yekingyan: > Should `layer_num` be `l` in `OneShotState::Initial(l) | OneShotState::Single(l)`?

Good catch — should be `l`.

Just pushed a fix (`0610cb4c`) that uses the match-bound `l` instead of the outer `layer_num` parameter.

They're numerically equal on every current call path, so there's no behavior change today, but `l` is the semantically correct value and it's more robust if that invariant ever changes.

jpds:
#1083 supersedes this by having intervening keys neither cancel nor extend the deadline.

HaoboGu:
Yeah, closing this since 1083 has been merged

PR #871, opened 2026-06-01
Add an option in chip_config.frequenecy to configure CPU clock of ESP32.

farooqkz:
>Currently it can be only 160 and 80 (values in MHz) as in par with esp-hal. As a note, when configuration of XTAL clock in
> esp-hal has been stabilized, we could also have that to further lower CPU clock to save battery.

The patch is ready for review. However, I think it could be a good idea if `rmk` prints a warning to the console if the chip is not ESP32 yet `frequency` is present.

Cheers!

### DIFF
--- a/rmk-config/src/lib.rs
+++ b/rmk-config/src/lib.rs
@@ -513,6 +513,9 @@ pub struct ChipConfig {
     /// DCDC regulator 0 voltage (for nrf52840)
     /// Values: "3V3" or "1V8"
     pub dcdc_reg0_voltage: Option<String>,
+    /// Frequency of the CPU. Currently supported only on
+    /// ESP32 and supports only 160 and 80(which are in MHz)
+    pub frequency: Option<String>
 }
 
 /// Config for lights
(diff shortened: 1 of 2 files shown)


HaoboGu:
Thanks for the PR! Did you try to override the initialization code in the proc-macro using `#[Override(chip_config)]`, like https://github.com/HaoboGu/rmk/blob/main/examples/use_config/stm32h7/src/main.rs

I think setting values like frequency to esp32 only is more like a "patch", it's not a common solution for all chips

farooqkz:
> Thanks for the PR! Did you try to override the initialization code in the proc-macro using `#[Override(chip_config)]`, like https://github.com/HaoboGu/rmk/blob/main/examples/use_config/stm32h7/src/main.rs
> 
> I think setting values like frequency to esp32 only is more like a "patch", it's not a common solution for all chips

Many chips work with different frequencies. So I guess the frequency should be configurable for those which support it. However for ESP32, only 80 and 160 will be valid. I think it makes sense to change the value type to `f32` and get this PR merged. Then open issues for others like STM32 for subsequent PRs if anyone is interested.

farooqkz:
History cleaned. `frequency` is now `f64`

farooqkz:
Checking HAL of stm32f1, I'm unsure if f64 fits.

HaoboGu:
Could you pls add some doc for this change?

farooqkz:
Sorry I'm busy with other stuff. How long can this wait?

HaoboGu:
No hurry. Just take your time

PR #874, opened 2026-06-04
feat(pointing): add pointing modes (Cursor/Scroll/Sniper)

Schievel1:
## Summary                                                                                                                                                                                                                                                                      
Thanks to Raymond8196  for most of the work on this. See also #728               
                                                                                                                                                                                                                                                                  Add pointing device modes to RMK, allowing trackballs to behave differently on different modes (e.g., scroll on layer 1, precision mode on layer 2).                                                                                                                 
                                                                                                                                                                                                                                                                                  
  ## Key Features                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                  
  **Pointing Modes:**                                                                                                                                                                                                                                                   
  - **Cursor**: Normal trackball mode (default)
  - **Scroll**: Convert motion to scroll wheel events
  - **Sniper**: Reduced sensitivity for precision control
  - **Caret**: Emit keytaps with trackball movement (mostly arrow keys, but can be configured)

  **Usage:**
  ```rust
use rmk::{
    event::{LayerChangeEvent, PointingProcessorEvent, publish_event},
    input_device::pointing::PointingMode, types::keycode::HidKeyCode,
};
use rmk_macro::processor;

#[processor(subscribe = [LayerChangeEvent])]
#[derive(Default)]
pub struct PointingProcessorController;

impl PointingProcessorController {
    pub fn new() -> Self {
        Self
    }

    async fn on_layer_change_event(&mut self, event: LayerChangeEvent) {
        match event.0 {
            0 => {
                publish_event(PointingProcessorEvent {
                    device_id: 255,
                    mode: PointingMode::Cursor(rmk::input_device::pointing::CursorConfig::default()),
                });
            }
            1 => {
                publish_event(PointingProcessorEvent {
                    device_id: 255,
                    mode: PointingMode::Sniper(rmk::input_device::pointing::SniperConfig {
                        divisor: 8,
                        invert_x: false,
                        invert_y: false,
                    }),
                });
            }
            2 => {
                publish_event(PointingProcessorEvent {
                    device_id: 255,
                    mode: PointingMode::Scroll(rmk::input_device::pointing::ScrollConfig {
                        divisor_x: 16,
                        divisor_y: 16,
                        invert_x: false,
                        invert_y: false,
                    }),
                });
            }
            3 => {
                publish_event(PointingProcessorEvent {
                    device_id: 255,
                    mode: PointingMode::Caret(rmk::input_device::pointing::CaretConfig {
                        disable_x: false,
                        disable_y: false,
                        invert_x: false,
                        invert_y: false,
                        threshold: 100,
                        keycode_up: HidKeyCode::Up,
                        keycode_down: HidKeyCode::Down,
                        keycode_left: HidKeyCode::Left,
                        keycode_right: HidKeyCode::Right,
                    }),
                });
            }

            _ => {}
        }
    }
}

```

  Technical Details:
  - Sub-pixel motion accumulation prevents precision loss
  - Configurable sensitivity divisors for each mode
  - Memory overhead: ~17 bytes (4-layer keyboard)
  - Backward compatible: default behavior unchanged

  Test Plan

  - unit tests for mode switching, accumulation, and edge cases
  - All existing tests pass, zero clippy warnings
  - Tested with PMW33xx/PMW3610 examples
  - Verified Vial compatibility

  Related Issues

  Addresses #703

TODO:
- [x] docs for toml configuration 
- [x] Docs on caret mode
- [x] Tests for caret mode
- [x] Make caret mode less jittery (although maybe thats just my sensor)
- [x] integration mode? (https://codeberg.org/Schievel/dactyl_manuform_r_track#headline-26) -> no

### CODE REVIEW

**Comment by Schievel1** [comment]
alrighty. Test fail, I will take care of this later. 

In qmk I used to have  integration scrolling mode. When in this mode, the trackball not just acts like a scroll wheel, but instead when  you move it in one direction, and leave it there, it keeps scrolling. (even though the trackball is not rolled anymore)
I never got the hang of it, but some people found it extremely useful for having autoscroll in programs that do not automatically support it. They set up a PDF, activate this slow scroll in their reading pace and start reading. 
Shall we implement that as well?

Also there's caret mode, which basically emits arrow keys when moving the trackball. Useful to move the cursor in text documents. 

More info on those:
https://codeberg.org/Schievel/dactyl_manuform_r_track#headline-26

File: docs/docs/main/docs/configuration/input_device/pmw33xx.mdx [hunk 1]
```diff
@@ -157,3 +188,97 @@ This should be added to the `central.rs`-File even if the sensor is on split per
...
+
+### Example Configuration
+Different modes are activated by sending a `PointingDeviceEvent` to the pointing processor.
+To control different processors independently, the event contains the device id to address the specific processor (255 to address all processors), the mode and its configuration.
+
+An example configuration with different layers switching on different modes could look like this:
+
+import { Tab, Tabs, Rust, Toml } from '@theme'
+
+<Tabs>
```
Schievel1: can I just rename the file to `pmw33xx.mdx` an then this is fine, or do I have to do something else?

rmk/src/event/input.rs
```diff
@@ -152,3 +157,13 @@ pub struct PointingSetCpiEvent {
     pub device_id: u8,
     pub cpi: u16,
 }
+
+/// Pointing device event
+/// TODO: Make the channel size configurable
+#[event(channel_size = 8, pubs = 2, subs = 2)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct PointingDeviceEvent {
```
Schievel1: Let's rename this to PointingProcessorEvent

docs/docs/main/docs/configuration/input_device/pmw33xx.mdx
HaoboGu: Yes, renaming it to `.mdx` is fine.

Schievel1:
> alrighty. Test fail, I will take care of this later.
> 
> In qmk I used to have integration scrolling mode. When in this mode, the trackball not just acts like a scroll wheel, but instead when you move it in one direction, and leave it there, it keeps scrolling. (even though the trackball is not rolled anymore) I never got the hang of it, but some people found it extremely useful for having autoscroll in programs that do not automatically support it. They set up a PDF, activate this slow scroll in their reading pace and start reading. Shall we implement that as well?
> 
> Also there's caret mode, which basically emits arrow keys when moving the trackball. Useful to move the cursor in text documents.
> 
> More info on those: https://codeberg.org/Schievel/dactyl_manuform_r_track#headline-26

I looked into this, and caret makes sense to implement. But integration mode is a thing that kind of just falls into place for QMK, because of the way QMK does things with polling and mouse reports. 
But for event based RMK it is way too complicated to implement for such a seldom used feature. Therefore I dropped the idea. If someone needs integration mode, they can do their own PointingProcessor and use that.

docs/docs/main/docs/configuration/input_device/pmw3610.mdx
```diff
@@ -0,0 +1,300 @@
...
+::: warning
+
+This should be added to the `central.rs`-File even if the sensor is on split peripheral.
+
+:::
+
+```rust
+    use rmk::input_device::pointing::{ PointingProcessor, PointingProcessorConfig };
+
+    let pmw3360_proc_config = PointingProcessorConfig {
```
kot149: 3360 is mixed in here

docs/docs/main/docs/configuration/input_device/pmw3610.mdx
```diff
@@ -0,0 +1,300 @@
...
+let pmw3610_sck = Output::new(p.P0_05, Level::High, OutputDrive::Standard);
+let pmw3610_sdio = Flex::new(p.P0_04);
+let pmw3610_cs = Output::new(p.P0_09, Level::High, OutputDrive::Standard);
+let pmw3610_motion = Some(Input::new(p.P0_02, Pull::Up));
+// or if you want to omit the motion pin:
+// let pmw3610_motion = Option::<Input<'static>>::None;
+
+let pmw3610_spi = BitBangSpiBus::new(pmw3610_sck, pmw3610_sdio);
+
+const POINTING_DEV_ID: u8 = 0 // this ID can be anything form 0-255. Just make sure you don't use the same number twice for different sensors to avoid confusion.
```
kot149: - missing `;` here
- typo: `from` -> `form`

docs/docs/main/docs/configuration/input_device/pmw33xx.mdx
```diff
@@ -157,3 +188,137 @@ This should be added to the `central.rs`-File even if the sensor is on split per
...
+```
+
+Create the controller and run it in the run_all! macro (rust configuration) or register it (toml configuration).
+
+import { Tab, Tabs, Rust, Toml } from '@theme'
+
+<Tabs>
+<Tab label={<Toml />}>
+
+```toml title="src/central.rs"
```
kot149: wrong file format

docs/docs/main/docs/configuration/input_device/pmw3610.mdx
```diff
@@ -0,0 +1,300 @@
...
+```
+
+Create the controller and run it in the run_all! macro (rust configuration) or register it (toml configuration).
+
+import { Tab, Tabs, Rust, Toml } from '@theme'
+
+<Tabs>
+<Tab label={<Toml />}>
+
+```toml title="src/central.rs"
```
kot149: wrong file format

rmk/src/input_device/pointing.rs
```diff
@@ -249,42 +253,267 @@ impl<S: PointingDriver> PointingDevice<S> {
...
+    }
+}
+/// Configuration for scroll mode
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
+pub struct ScrollConfig {
+    /// Divisor for X axis (→ pan). Higher = slower. 0 disables horizontal pan.
+    pub divisor_x: u8,
+    /// Divisor for Y axis (→ wheel). Higher = slower. 0 disables vertical scroll.
+    pub divisor_y: u8,
```
kot149, line 352: Same comment as https://github.com/HaoboGu/rmk/pull/728#discussion_r2862313070: I think it's better to have multipliers in addition to divisors [like ZMK](https://zmk.dev/docs/keymaps/input-processors/scaler); It allows more flexible configurations, and multiplier = 0 for disabling the axis.
Also applies to sniper mode.

docs/docs/main/docs/configuration/input_device/pmw3610.mdx
```diff
@@ -0,0 +1,300 @@
...
+    POINTING_DEV_ID
+    pmw3610_spi,
+    pmw3610_cs,
+    pmw3610_motion,
+    pmw3610_config,
+);
+
+// If you want to customize the report rate (Hz):
+// let mut pmw3610_device = Pmw3610Device::with_report_hz(
+//     POINTING_DEV_ID
```
kot149: missing `,` here

kot149 (changes requested):
Thanks for working on this. A few comments from me:

rmk/src/input_device/pointing.rs
Schievel1: Just adding multiplier with default = 1 and in accumulator.accumulate() we do "x / divisor * multiplier" or what do you have in mind?

Schievel1:
```

pub fn accumulate(&mut self, dx: i16, dy: i16, ratio_x: (u8, u8), ratio_y: (u8, u8)) -> (i16, i16) {
    let out_x = if ratio_x.1 == 0 {
        self.remainder_x = 0;
        0
    } else {
        let total_x = self.remainder_x.saturating_add(dx * ratio_x.0 as i16);
        let out = total_x / ratio_x.1 as i16;
        self.remainder_x = total_x - out * ratio_x.1 as i16;
        out
    };
    // ... 
```

Like this?

docs/docs/main/docs/configuration/input_device/pmw33xx.mdx
kot149: `warning` must be lowercase

rmk/src/input_device/pointing.rs
```diff
@@ -296,15 +560,178 @@ impl<'a> PointingProcessor<'a> {
...
+            }
+            Axis::Y => {
+                let r = if dy > 0 { -cfg.threshold } else { cfg.threshold };
+                (0, r)
+            }
         };
-        send_hid_report(Report::MouseReport(mouse_report)).await;
+        (dx, dy) = accumulator.accumulate_persistent(reduce_x, reduce_y, (1, divisor_x), (1, divisor_y));
+        count = count.saturating_add(1);
+    }
```
kot149, line 738: This loop causes an infinite loop and hangs the keyboard during standard diagonal trackball movements. Current tests only cover cases where the non-dominant axis is small, such as (150, 30). However, common diagonal inputs like (150, 150) or (120, 120) will fail.

Inside the loop, only the dominant axis is reduced while the non-dominant axis remains unchanged. Furthermore, because `accumulate_persistent` is used, the dominant axis value does not consume its remainder but instead flips signs and oscillates around the threshold (e.g., `150 -> 50 -> -50 -> 50 -> ...`). As a result, if the non-dominant axis value is large enough, `dx.abs() + dy.abs()` will never drop below the threshold. 
Also, even though `count` uses `saturating_add`, the loop condition does not check the `count`, meaning the loop never terminates once it hits `255`.

To fix this, we should either reset the non-dominant axis inside the loop so it does not block termination, or change the loop condition to check only the dominant axis. Additionally, we should add a defensive guard with a hard upper limit to the loop iterations to guarantee termination, and introduce new test cases with large values on both axes.

rmk/src/input_device/pointing.rs
```diff
@@ -296,15 +560,178 @@ impl<'a> PointingProcessor<'a> {
...
+    // Release
+    send_hid_report(Report::KeyboardReport(KeyboardReport {
+        modifier: 0,
+        reserved: 0,
+        leds: 0,
+        keycodes: [0, 0, 0, 0, 0, 0],
+    }))
+    .await;
+    Timer::after_millis(5).await;
+}
```
kot149, line 673: This report sends a raw `KeyboardReport` directly and clears all keycodes and modifiers to zero upon release. If the user is pressing other keys or modifier keys (such as Shift) at the same time, that state will be cleared. This breaks common use cases like Shift + Arrow for selection in Caret mode.

rmk/src/input_device/pointing.rs
```diff
@@ -296,15 +560,178 @@ impl<'a> PointingProcessor<'a> {
...
-            y: y.clamp(i8::MIN as i16, i8::MAX as i16) as i8,
-            wheel: 0,
-            pan: 0,
+        match self.current_mode {
+            PointingMode::Cursor(_) | PointingMode::Scroll(_) | PointingMode::Sniper(_) => {
+                // modes that generate mouse reports
+                let mouse_report = match self.current_mode {
+                    PointingMode::Cursor(cursor_config) => {
+                        let out_x = x * cursor_config.multiplier_x as i16;
+                        let out_y = y * cursor_config.multiplier_y as i16;
```
kot149: We should use `saturating_mul()`

rmk/src/input_device/pointing.rs
```diff
@@ -249,42 +253,302 @@ impl<S: PointingDriver> PointingDevice<S> {
...
-        Self { keymap, config }
+        Self {
+            keymap,
+            config,
+            accumulator: MotionAccumulator::default(),
+            current_mode: PointingMode::default(),
+        }
+    }
+
+    /// Set the pointing mode for a specific layer
```
kot149: Incorrect doc comment: this method is not layer-specific

rmk/src/input_device/pointing.rs
```diff
@@ -296,15 +560,178 @@ impl<'a> PointingProcessor<'a> {
...
+    let (mut dx, mut dy) = accumulator.accumulate_persistent(x, y, (1, divisor_x), (1, divisor_y));
+
+    if (dx.abs() + dy.abs()) <= cfg.threshold {
+        return None;
+    }
+
+    enum Axis {
+        X,
+        Y,
+    }
```
kot149: The name is confusingly similar to `crate::event::Axis`, so it would be better to distinguish the names.

HaoboGu:
I'll need for time to review it

docs/docs/main/docs/configuration/input_device/pointing_processor.mdx
```diff
@@ -0,0 +1,231 @@
...
+```
+
+Uses `MotionAccumulator` in persistent mode: `total = remainder + delta`, output = `total / divisor`, but the **full total** stays in the accumulator (not just the remainder). The consumer (caret logic) decides when to reset the accumulator via `reset_x()`/`reset_y()` after a tap has been triggered.
+
+## Mode switching
+
+Users can define a processor, in this case called `PointingProcessorController`, which subscribes to events (e.g. layer changes) and updates the active mode. The example below switches to Cursor on layer 0, Sniper on layer 1, Scroll on layer 2, and Caret on layer 3 or when key bound to User0 is pressed:
+
+```rust
+```rust title="src/pointingproccontroller.rs"
```
kot149: duplicate ` ```rust `

kot149 (changes requested):
Thanks for addressing the feedback on the multiplier. I have a few additional comments:

rmk/src/input_device/pointing.rs
Schievel1: Yes, I wasn't sure about this in the first place. 
But how would I send the hid keycode down to keyboard.rs to get it processed with every modifier?

rmk/src/input_device/pointing.rs
Schievel1: I added both, resetting the non-dominant axis and added a safety break when count gets too high

rmk/src/input_device/pointing.rs
Schievel1: Or even a KeyAction, that would be even better if I could send this. 
I would suggest if you don't have an idea, we keep it as is for now and in a follow up PR we implement some mechanism to trigger KeyActions programmatically.

docs/docs/main/docs/configuration/input_device/pointing_processor.mdx
```diff
@@ -0,0 +1,230 @@
...
+// ...
+    let mut pointing_controller = PointingProcessorController::default();
+    run_all!(
+		/* other processors go here */
+        pointing_controller,
+    ),
+```
+
+</Tab>
+</Tabs>
```
kot149, line 238: This `<Tabs>` causes the following error; add `import { Rust, Toml, Tab, Tabs } from '@theme'` to fix it.
```
error   [browser] Uncaught Error: Expected component `Rust` to be defined: you likely forgot to import, pass, or provide it. at _missingMdxReference (docs\main\docs\configuration\input_device\pointing_processor.mdx:3996:0)
```

Not sure `<Tabs>` is actually needed here, both tabs are actualy Rust.

rmk/src/input_device/pointing.rs
```diff
@@ -605,4 +1041,777 @@ mod tests {
...
+        assert_eq!(a.remainder_x, 50);
+        assert_eq!(a.remainder_y, 50);
+    }
+
+    #[test]
+    fn test_compute_caret_taps_x_dominant_default_right() {
+        let mut a = acc();
+        let result = compute_caret_taps(150, 30, &mut a, &cfg());
+        // |150|+|30| = 180 > 100. X dominant (150 >= 30). dx>0 + !invert → Right.
+        // Loop: reduce_x=-100 → total=(50, 30), |50|+|30|=80<=100 → stop.
```
kot149: Stale test comment: After `reset_y()`, it's actually `total=(50, 0), |50|+|0|=50`. The assert is correct, just the comment needs updating.

rmk/src/input_device/pointing.rs
kot149: Thanks for the fix, looks good to me

rmk/src/input_device/pointing.rs
kot149: I'm actually not sure about this too. I think it's fine to leave it as is for now, but could we at least add a doc comment on tap_key noting this limitation (modifiers are not preserved), and mention it in the pointing processor docs?

docs/docs/main/docs/configuration/input_device/pointing_processor.mdx
Schievel1: the tabs are both Rust, but one is for toml based config, one is for rust based config.

kot149 (approved):
LGTM 👍

Schievel1:
Thanks for reviewing this.

HaoboGu (approved):
Thanks a lot!!

PR #877, opened 2026-06-08
feat(dfu): Embassy boot and dfu flashing

Schievel1:
This PR adds DFU (Device Firmware Upgrade) support via embassy-boot for RP2040-based and NRF52840-based keyboards. Once set up, firmware updates can be done over USB with dfu-util — no need to press BOOTSEL/ RESET or use a debug probe.

# What's implemented
rmk crate — new dfu_rp feature gate:
- dfu.rs: init_flash() initializes the blocking flash with configurable partition offsets and returns a storage partition for RMK's keymap storage
- set_led() / mark_booted() — optional DFU activity LED and boot-success signaling to prevent bootloader rollback
- register_dfu_interface() — registers a USB DFU class (embassy-usb-dfu) with a handler that blinks the LED during transfers
- DfuLock (behind dfu_lock feature) — physical key unlock for DFU: pressing the configured matrix keys unlocks DFU for 10 s, with Morse "DFU" LED blinking
- Automatic DFU interface registration in UsbTransport::new() when the manager is initialized

# rmk-config crate — new [dfu] TOML section:
- flash_size, page_size — auto-calculate partition addresses using the bootymcbootface formula (state at 0x6000, active at 0x7000, DFU slot follows) (bootymcbootface (https://codeberg.org/Schievel/bootymcbootface/) is the matching bootloader for this, but people could roll their own and use whatever partitioning suits them)
- led — optional GPIO pin for DFU activity
- unlock_keys — matrix positions for dfu_lock
- Manual override of state_offset, state_size, dfu_offset, dfu_size for custom layouts

# flash formula
embassy-boot needs a DFU and an ACTIVE partition. ACTIVE is where the actual firmware sits, DFU is where we can put firmware using dfu-utils for example and embassy-boot then copies DFU to ACTIVE. 
Apart from that me have:
- BOOT2, where the RP2040 bootloader lives, address 0x10000000 (beginning of flash), length fixed 0x100 (for nRF52840 this does not exist and the embassy-boot partition is expanded to start at 0x0) 
- embassy-boot, where embassy-boots code lives, after BOOT2, address fixed 0x10000100, length fixed to 24K-0x100
- BOOTLOADER_STATE, where embassy-boot puts it's own state, and where we set a flag from RMK that we want to update DFU to ACTIVE, after embassy-boot, address fixed 0x10006000, length fixed 4K
- ACTIVE, where the currently running RMK lives, address fixed 0x10007000, variable length
- DFU, where we put the firmware we want embassy-boot to flash, address variable (depends on ACTIVE length), length = ACTIVE length + 1 page
- STORAGE, address variable (depends on ACTIVE + DFU), length fixed 128K

As a diagram:
 ```
  ┌─────────────────────────────────┐
  │ 0x10000000                      │
  │  +───────────────+  0x10000100  │
  │  │ BOOT2 (256B)  │              │
  │  +───────────────+  0x10000100  │
  │  │               │              │
  │  │ EMBASSY-BOOT  │              │
  │  │ (24K - 256B)  │              │
  │  │               │              │
  │  +───────────────+  0x10006000  │
  │  │ BOOTLOADER    │              │
  │  │ STATE (4K)    │              │
  │  +───────────────+  0x10007000  │
  │  │               │              │
  │  │ ACTIVE        │              │
  │  │ (variable)    │              │
  │  │               │              │
  │  ├───────────────┤  ACTIVE_END  │
  │  │               │              │
  │  │ DFU           │              │
  │  │ (ACTIVE+4K)   │              │
  │  │               │              │
  │  ├───────────────┤  DFU_END     │
  │  │               │              │
  │  │ STORAGE       │              │
  │  │ (128K)        │              │
  │  │               │              │
  │  +───────────────+  STORAGE_END │
  │ FLASH_END                       │
  └─────────────────────────────────┘
```

In order to partition the flash I use the following constraints:
- we want ACTIVE partition to be a big as possible while also allowing a bit of storage
- DFU needs to be as big as active + 1 page that embassy-boot uses for the state during flashing. This way it known where it was after continuing after a power loss and can continue flashing DFU to ACTIVE
- the partitions before ACTIVE (BOOT2, embassy-boot, BOOTLOADER_STATE) have fixed addresses and lengths. Therefore ACTIVE has a fixed address as well. (but not a fixed length
- since ACTIVE has a variable length, the address  of DFU must be calculatd. 
- STORAGE comes after ACTIVE and DFU, which have a variable length, therefore the address of STORAGE must be calculated.

To calculate the length of ACTIVE, I calculate the remaining flash after subtracting all the fixed size partitions (BOOT2, EMBASSY-BOOT, BOOTLOADER-STATE, STORAGE):
`remaining = flash_size - 28K (BOOT2, EMBASSY-BOOT, BOOTLOADER-STATE) - 128K (STORAGE size)`
The remaining flash must then be equally sized between ACTIVE and DFU, where DFU gets one more page:
`ACTIVE size = (remaining - 4K (page size)) / 2`
`DFU size = ACTIVE size + 4K (page size)`

Then we only need to get the start addresses of DFU and storage:
`DFU offset =  ACTIVE offset (always 0x10007000) + ACTIVE size`
and 
`STORAGE offset = DFU offset + DFU size`

# Examples:
- examples/use_rust/rp2040_embassy_boot/ — full Rust-API example
- Example configs updated for use_config path

# Documentation:
- New user_guide/flash_firmware/use_embassy_boot.mdx with setup guide (bootymcbootface, compilation, step-by-step flashing, DFU usage, dfu_lock, LED behavior)
- RP2040 + embassy-boot section in configuration/chip_config.mdx with partition table and TOML/Rust tabs

# Feature flags
- dfu — base DFU support (USB DFU interface)
- dfu_rp — RP2040-specific DFU with embassy-boot (implies dfu + rp2040)
- dfu_lock — physical key unlock for DFU (implies dfu_rp + host_security)
# Usage flow
1. Flash bootymcbootface (once, via UF2 or probe-rs)
2. Compile RMK with dfu_rp feature and matching flash size
3. Flash RMK firmware (via UF2 or probe-rs)
4. All future updates: cargo make bin && dfu-util -D firmware.bin

# Perspective / Roadmap
Since embassy-boot is agnostic towards how the updated firmware gets into the DFU partition, this PR lays the foundation for future features like flashing via bluetooth or flashing the peripheral from the central using the serial connection between them.
I think the next steps after merging this should be:
- flash nrf/ pico w using BLE (maybe with our own nrf ble dfu implementation. The protocol is public.
- flash peripheral from central (see #886)

# Needs: 
- [ ] https://github.com/HaoboGu/rmk-template/pull/20
# Optional:
- [ ] https://github.com/HaoboGu/rmkit/pull/11
- [ ] https://github.com/HaoboGu/rmk-template/pull/21

### CODE REVIEW

**Comment by Schievel1** [comment]
please review the addresses part thoroughly. I find the flash partitioning and the addressing (inside and outside the partition) quite confusing, as you can see with my bugfix commit for the storage start address.

**Comment by Schievel1** [comment]
Shall I squash the commits or do you want to keep the history?

**Comment by HaoboGu** [comment]
Yes, please squash the commit

File: rmk/src/dfu/mod.rs [hunk 1]
```diff
@@ -0,0 +1,471 @@
...
+type FlashType = Flash<'static, FLASH, Blocking, FLASH_SIZE>;
+#[cfg(feature = "dfu_nrf")]
+type FlashType = Nvmc<'static>;
+
+#[cfg(any(feature = "dfu_rp", feature = "dfu_nrf"))]
+type MutexType = Mutex<CriticalSectionRawMutex, RefCell<FlashType>>;
+#[cfg(any(feature = "dfu_rp", feature = "dfu_nrf"))]
+type PartitionType = BlockingPartition<'static, CriticalSectionRawMutex, FlashType>;
+
+#[cfg(any(feature = "dfu_rp", feature = "dfu_nrf"))]
```
HaoboGu, line 51: Make dfu a module and add `rp` & `nrf` separately? many feature gates can be eliminates this way.

rmk/src/dfu.rs
```diff
@@ -0,0 +1,471 @@
...
+
+/// Store an optional DFU LED pin globally.
+#[cfg(any(feature = "dfu_rp", feature = "dfu_nrf"))]
+pub fn set_led(led: Option<Output<'static>>) {
+    let _ = LED.init(Mutex::new(RefCell::new(led)));
+}
+
+/// Run a closure with the global DFU LED, if configured (platform-specific impl).
+#[cfg(any(feature = "dfu_rp", feature = "dfu_nrf"))]
+fn with_led<F, R>(f: F) -> Option<R>
```
HaoboGu: I prefer to send events, instead of hardcoding a LED in dfu mode

rmk/src/dfu.rs
```diff
@@ -0,0 +1,471 @@
...
+    }
+
+    fn system_reset(&mut self) {
+        with_led(|led| led.set_low());
+        self.inner.system_reset()
+    }
+}
+
+#[cfg(feature = "dfu_rp")]
+type FlashPeri = Peri<'static, FLASH>;
```
HaoboGu: We can use trait in embedded-storage instead of hardcode `FlashPeri`

rmk/src/dfu/mod.rs
Schievel1: I wanted to stick to DRY, but you're right, it is getting out of hand. I'm making modules

rmk/src/dfu.rs
Schievel1: I think this there is no need for either of that once I refactored into modules

rmk/src/dfu.rs
Schievel1: true that, I try and use the battery led as a template

docs/docs/main/docs/user_guide/flash_firmware/use_embassy_boot.mdx
```diff
@@ -0,0 +1,213 @@
...
+cargo run --release --target thumbv7em-none-eabihf --features nrf52840 # nRF52840
+```
+
+#### 2. Flash RMK firmware
+
+Now flash the actual RMK firmware. The bootloader stays in place.
+
+**Method A – via UF2:**
+- **RP2040**: Put the board into bootloader mode (hold BOOTSEL button, plug in USB, release), then copy the RMK firmware `.uf2` file to the appearing `RPI-RP2` drive.
+- **nRF52840**: Since the Adafruit bootloader was overwritten when flashing the bootymcbootface bootloader, this method does not work for nRF52840 anymore.
```
HaoboGu, line 125: So for nRF52840, the first RMK firmware flashing requires a debug probe? Am I correct?

docs/docs/main/docs/user_guide/flash_firmware/use_embassy_boot.mdx
```diff
@@ -0,0 +1,213 @@
+# Use embassy-boot bootloader
+
+[embassy-boot](https://github.com/embassy-rs/embassy/tree/main/embassy-boot) is a bootloader framework from the Embassy project. It splits the flash into two firmware slots (active and DFU download), so updating is **safe even if power is lost or the transfer fails** — the old firmware remains intact and the bootloader can roll back automatically. You can build your own bootloader, or use a pre-built one like **bootymcbootface**.
+
+RMK supports DFU for **RP2040** (via `dfu_rp`) and **nRF52840** (via `dfu_nrf`).
+
+[bootymcbootface](https://codeberg.org/Schievel/bootymcbootface/) is a pre-built embassy-boot bootloader for both platforms. It sits at the beginning of flash and handles the dual-slot firmware switching and rollback on boot. RMK itself provides the DFU USB interface (via `embassy-usb-dfu`) for subsequent updates, so once bootymcbootface and RMK are flashed, you never need to press BOOTSEL again.
```
HaoboGu, line 7: Would you mind adding `bootymcbootface` to rmk project as `rmk-bootloader`? Just curious about it because I do plan to add RMK's own bootloader. It seems that `bootymcbootface` has done everything well.

HaoboGu (approved):
Very nice, thanks for your effort!

docs/docs/main/docs/user_guide/flash_firmware/use_embassy_boot.mdx
Schievel1: It depends. I recently bought some nrf52840 that were clones of the sparkfun pro micro nrf and they had the adafruit uf2 bootloader preinstalled. I guess the official one and the nrfs from adafruit have that as well.
In this case you can flash the embassy boot bootloader using the uf2 bootloader, given the correct memory.x

docs/docs/main/docs/user_guide/flash_firmware/use_embassy_boot.mdx
Schievel1: I had this in mind while writing it.
What needs to be done to do that? Do we need to canonicalize the structure of it? Change the name? 

Name wise I also have:
- Das Boot (famous German movie from tue 80s)
- SirBootsalot 
- Bootique 
- BootyAndTheBeast

HaoboGu: I don't have options about the naming. What I thought about it was to make the bootloader maintained in rmk project, so that we can keep everything updated.

docs/docs/main/docs/user_guide/flash_firmware/use_embassy_boot.mdx
HaoboGu: yeah, I just wanted to ensure that users can flash the bootymcbootface bootloader without the debug probe -- not everyone has it.

Schievel1: Oh right.
So what they could do is flash the RMK firmware with dfu feature and right memory.x, this won't overwrite the adafruit because it's smaller than 0x7000 where the RMK firmware starts. This won't boot then of course because the adafruit bl expects it at a different address. Then flash the bootymcbootface using the adafruit bootloader. 
I haven't tested this and I think it's very brittle but it could work. 
Still I would recommend getting a debug probe. The moment you do something in the wrong order, have a wrong memory.x you accidentally overwrite the adafruit bl at the wrong time and the chip is bricked (until you get a debug probe) 

I think I left enough warnings that this overwrites the adafruit bootloader in the docs and also in the readme of bootmcbootface.

docs/docs/main/docs/user_guide/flash_firmware/use_embassy_boot.mdx
Schievel1: Ok then we stick to bootymcbootface as it's silly enough for my taste already. 

But let's do this in a separate PR

docs/docs/main/docs/user_guide/flash_firmware/use_embassy_boot.mdx
HaoboGu: What about to include the DFU flashing in bootloader? After flashing bootymcbootface bootloader, users can just flash the keyboard firmware using DFU.

docs/docs/main/docs/user_guide/flash_firmware/use_embassy_boot.mdx
HaoboGu: Great!

docs/docs/main/docs/user_guide/flash_firmware/use_embassy_boot.mdx
Schievel1: Yes I thought about that as well. But we have to make sure the bootloader fits in 28kB (0x0-0x7000) 
But that's a different PR then.

HaoboGu: feels like it's a little bit hard lol. Okay let's put it in another PR

HaoboGu (approved):
It looks great now

HaoboGu:
Thanks again for your effort! Having DFU support is great

docs/docs/main/docs/user_guide/flash_firmware/use_embassy_boot.mdx
Schievel1: Yes I mean it's not only the DFU protocol but the whole USB stack. 
On the other hand the rp2040 stage 1 bl has usb with mass storage and fits that in 16kB iirc so it's probably possible

Schievel1:
Great, thanks for reviewing and merging. I am preparing the follow up with dfu_split support now

Issue #883, opened 2026-06-12
feature: Add ability to override the `enable_tap_flow` on profile level


alxzh:
Today the enable_tap_flow can only be set on the default morse config, which makes it all or nothing. This setting works very well for HRM, especially for GUI, Ctrl, Alt and layer switch buttons on the home row. But if one have Shift with the tap_flow it forces one to slow down a lot when typing something like "NaNs" tap_flow makes one to slow down very noticeably.

Would be nice to have a way to have overrides for tap_flow on Morse profile level. This would allow to have better control over where user wants tap_flow and where they are OK with hold preferred behavior.

I did a bit of digging in the code and looks like the profile configs are encoded as u32 holding flags and the timeout overrides.  Adding tap_flow override is going to require expanding the the type size to u64 or shrinking the timeout value representable in these overrides. Today the timeout overrides allow to set values up to 16s. I would question if this is really required? I have never found a use to set these timeouts to any thing above 1s. So shrinking the timeout override value scale is probably OK, because users would be able to still set to something up to 8s, which should be more than enough for most cases.

PR #885, opened 2026-06-13
feat: Add support for overriding flow_tap per profile

alxzh:
This adds a feature that allows to override the tap_flow configuration option per Morse profile. This effectively addresses https://github.com/HaoboGu/rmk/issues/883. The trade of is that it shrinks the max possible timeout overrides from 16s to 8s per-profile. Given this change doesn't regress the profile storage size it seems like an acceptable tradeoff. In most cases users don't really need to set these timeout to something higher than 1 or 2 seconds.

### DIFF
--- a/rmk-types/src/morse.rs
+++ b/rmk-types/src/morse.rs
@@ -41,19 +41,41 @@ pub enum MorseMode {
 ///
 /// Bit layout of the inner `u32`:
 /// ```text
-/// 31  30 | 29      16 | 15  14 | 13       0
-/// mode   | gap_timeout | uni_tap| hold_timeout
-///  (2b)  |   (14b ms)  |  (2b)  |  (14b ms)
+/// 31  30 | 29       17 | 16  15 | 14  13   | 12       0
+/// mode   | gap_timeout | uni_tap| flow_tap | hold_timeout
+///  (2b)  |   (13b ms)  |  (2b)  |   (2b)   |  (13b ms)
 /// ```
 ///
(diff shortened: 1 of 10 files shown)


rmk-types/src/morse.rs
```diff
@@ -41,49 +41,88 @@ pub enum MorseMode {
 ///
 /// Bit layout of the inner `u32`:
 /// ```text
-/// 31  30 | 29      16 | 15  14 | 13       0
-/// mode   | gap_timeout | uni_tap| hold_timeout
-///  (2b)  |   (14b ms)  |  (2b)  |  (14b ms)
+/// 31  30 | 29       | 28      16 | 15  14 | 13       | 12       0
+/// mode   | flow_tap | gap_timeout | uni_tap| flow_tap | hold_timeout
+///  (2b)  |   (1b)   |   (13b ms)  |  (2b)  |   (1b)   |  (13b ms)
```
HaoboGu: Why `flow_tap`'s bit is at 29 and 13? I think it's fine to make it adjacent bits, the next version will be breaking change.

alxzh: good point! I was thinking about preserving backward compatibility too much. Should be fixed now

alxzh:
Rebased and fixed conflicts

alxzh:
@HaoboGu does the change make sense?

HaoboGu (approved):
It looks good, thanks a lot!

PR #886, opened 2026-06-15
feat(dfu_split): Embassy boot flash peripheral from central 

Schievel1:
This is a follow up PR for #877 which uses embassy-boot to flash the peripheral through the central.

The idea is to minimize the effort of flashing. #877 ensures, that we don't need to open up the keyboard and press the boot0 button to get the RP2040 into bootloader. It is enough to run dfu-util on the host to flash it. 
This PR takes this idea further. We embed the peripheral's firmware into the centrals firmware, then during startup of the central, we ask peripheral for it's current firmware CRC. Central calculates the CRC of it's embedded peripheral firmware. If they don't match, peripheral has an old firmware and central needs to flash it. 
It then proceeds sending chunks which peripheral puts into its DFU partition, when done central makes this known to the peripheral. Then peripheral resets. 
After that this is normal embassy-boot bootloader flash, it copies DFU to ACTIVE, boots it, if it booted, good, if not embassy-boot does a rollback. (That rollback thing is important, because it means we can try again flashing a working firmware without the need to take the keyboard apart). 
After boot the peripherals firmware CRCs on central and peripheral (hopefully) match, and we don't flash the peripheral again. 

If peripheral misses the central asking for its firmwares CRC (e.g. starts after central) it would not get an update. Therefore we let peripheral send it's firmares CRC proactively at every start and central reacts to it accordingly.

This obviously needs enough flash space on the central to hold the peripherals firmware. 

additionally we can use the central as a relay for updating the peripheral via dfu. This uses the altsetting of DFU. (usually this is used to flash different areas of the flash of a chip instead of different chips)
When we run `dfu-util -D ./central.bin -a 0` we flash the central. If we run `dfu-util -D ./peripheral.bin -a 1` we can flash the peripheral even though only the central is plugged in the host's USB. 
For several peripherals we can specify `-a 2`, `-a 3` etc. 

The diagram below shows that we check the CRC of every chunk we send, send it to central and wait for an ack before continuing. Of course this could be optimized (for example with a window of 20 chunks that we check the CRC of and so on) but this flashes the examples peripheral firmware (~34kB) via the central as a relay in <4s. I think this is good enough for now. 

Currently only supports serial connection. I experiment with bluetooth still, but couldn't get in to work properly.  

Example logs: 
Peripheral running, central starts up and has newer peripheral firmware:
```
[INFO ] RMK start! (central RMKRP2040embassy-split/src/central.rs:44)
[INFO ] Flash layout: state @ 0x6000 (4K), active @ 0x7000 (944K), dfu @ 0xF3000 (948K), storage @ 0x1E0000 (128K) (central RMKRP2040embassy-split/src/central.rs:83)
[INFO ] Flash capacity 128 KB, RMK use 128 KB(32 sectors) starting from 0x0 as storage (rmk src/storage/mod.rs:434)
[INFO ] Device enabled (rmk src/usb/mod.rs:451)
[INFO ] Running peripheral manager 0 (rmk split/serial/mod.rs:27)
[INFO ] dfu_split: checking peripheral firmware... (rmk src/split/driver.rs:201)
[INFO ] Device suspended, the Vbus current limit is 500µA (or 2.5mA for high-power devices with remote wakeup enabled). (rmk src/usb/mod.rs:490)
[INFO ] dfu_split: firmware hash mismatch (peripheral=0x2112eded, expected=0x6938687c), starting update (31360 bytes) (rmk src/split/driver.rs:245) <-- new peripheral firmware in central, needs to flash 
[INFO ] dfu_split: sending chunk at offset 0 (256 bytes) (rmk src/split/driver.rs:276)
[INFO ] dfu_split: sending chunk at offset 256 (256 bytes) (rmk src/split/driver.rs:276)
[INFO ] dfu_split: sending chunk at offset 512 (256 bytes) (rmk src/split/driver.rs:276)
... chunks keep coming
[INFO ] dfu_split: sending chunk at offset 30976 (256 bytes) (rmk src/split/driver.rs:276)
[INFO ] dfu_split: sending chunk at offset 31232 (64 bytes) (rmk src/split/driver.rs:276)
[INFO ] dfu_split: all chunks sent, signaling update complete (rmk src/split/driver.rs:312)
[INFO ] dfu_split: peripheral confirmed, update complete (rmk src/split/driver.rs:323)
[WARN ] Framing error (embassy_rp src/uart/buffered.rs:538) <- Peripheral resets itself
[WARN ] Break error (embassy_rp src/uart/buffered.rs:544)
[ERROR] Peripheral message read error: SerialError (rmk src/split/driver.rs:146)
[INFO ] dfu_split: proactive hash from peripheral (0x2112eded), checking... (rmk src/split/driver.rs:171) <- peripheral starts up with new firmware
[INFO ] dfu_split: peripheral firmware hash matches (0x2112eded), no update needed (rmk src/split/driver.rs:176) <- firmware CRCs match now, no update. 
```

Central runnnig, peripheral starts up:
```
[INFO ] Device suspended, the Vbus current limit is 500µA (or 2.5mA for high-power devices with remote wakeup enabled). (rmk src/usb/mod.rs:490)
[ERROR] Peripheral message read error: SerialError (rmk src/split/driver.rs:146)
[INFO ] dfu_split: proactive hash from peripheral (0x6938687c), checking... (rmk src/split/driver.rs:171) <- peripheral starts up
[INFO ] dfu_split: firmware hash mismatch (peripheral=0x6938687c, expected=0x67363949), starting update (31264 bytes) (rmk src/split/driver.rs:181) <- firmware hashes don't match, update peripheral with embedded firmware in central
[INFO ] dfu_split: sending chunk at offset 0 (256 bytes) (rmk src/split/driver.rs:276)
[INFO ] dfu_split: sending chunk at offset 256 (256 bytes) (rmk src/split/driver.rs:276)
[INFO ] dfu_split: sending chunk at offset 512 (256 bytes) (rmk src/split/driver.rs:276)
... chunks keep coming
[INFO ] dfu_split: sending chunk at offset 30976 (256 bytes) (rmk src/split/driver.rs:276)
[INFO ] dfu_split: sending chunk at offset 31232 (32 bytes) (rmk src/split/driver.rs:276)
[INFO ] dfu_split: all chunks sent, signaling update complete (rmk src/split/driver.rs:312)
[INFO ] dfu_split: peripheral confirmed, update complete (rmk src/split/driver.rs:323)
[WARN ] Framing error (embassy_rp src/uart/buffered.rs:538) <- peripheral restart
[WARN ] Break error (embassy_rp src/uart/buffered.rs:544)
[ERROR] Peripheral message read error: SerialError (rmk src/split/driver.rs:146)
[INFO ] dfu_split: proactive hash from peripheral (0x67363949), checking... (rmk src/split/driver.rs:171) <- peripheral starts up, sends its firmwares CRC
[INFO ] dfu_split: peripheral firmware hash matches (0x67363949), no update needed (rmk src/split/driver.rs:176) <- firmware flashed successfully, no update needed
```

The AI overlords drew this for me, all hail to the AI overlords:
```
┌─────────────────────────────────────────────────────────────────────┐
│                      SPLIT DFU UPDATE                               │
│   (peripheral connects / is connected)                              │
└─────────────────────────────────────────────────────────────────────┘


═══ PHASE 0: HANDSHAKE ── hash comparison ───────────────────────────
Do we need to update the peripheral? 

  Central                           Peripheral
    │                                   │
    │                                   ├── (on boot) FirmwareHashResponse(hash)
    │                                   │   <- read_embedded_firmware_hash()
    │                                   │   = CRC32(__vector_table..__veneer_limit)
    │                                   │
    │  ── or ──                         │
    │                                   │
    ├── FirmwareHashQuery ─────────────>│
    │<── FirmwareHashResponse(hash) ────┤
    │                                   │
    │  hash == expected_hash?           │
    │    ├─ Yes → STOP (up-to-date)     │
    │    └─ No  → ⬇                     │
    │                                   │


═══ PHASE 1: CHUNK TRANSFER ── per-chunk CRC ────────────────────────
During transfer central calculates CRC of every chunk it sends, peripheral does as well and 
sends the CRC back in a FirmwareChunkAck. If they match, central sends the next chunk. 
If they don't match, retry the chunk. 
During sending Central calculates the overall CRC over all chunks sent. (central_crc.update(chunk))

  (outer attempt loop: 1..3)

  Central                           Peripheral
    │                                   │
    │ LED ON                            │
    │ central_crc = Crc32::new()        │
    │                                   │
    │   for chunk in firmware[256]:     │
    │     chunk_crc = CRC32(chunk)      │
    │     central_crc.update(chunk)     │
    │                                   │
    │   retry = 0                       │
    │   ┌─ retry < 3 ─────────────────┐ │
    │   │                             │ │
    ├── FirmwareChunk{offset,data} ──>│ │
    │   │                             │ │
    │   │               handler.write_chunk() — full erase on 1st call
    │   │               chunk_crc = CRC32(chunk) │
    │   │                             │ │
    │   │<─ FirmwareChunkAck{offset,──┤ │
    │   │         crc: chunk_crc}     │ │
    │   │                             │ │
    │   CRC match:                    │ │
    │   ack_crc == chunk_crc?         │ │
    │     ├─ Yes → next chunk         │ │
    │     └─ No  → retry++ ────── ────┘ │
    │                                   │
    │   All chunks acked?               │
    │     ├─ Yes → ⬇                    │
    │     └─ No  → attempt++ → Phase 1  │


═══ PHASE 2: END-TO-END CRC ── flash readback ──────────────────────
After all chunks are done, central compares CRC from chunk updates with the firmare CRC from phase 0, and if ok, central sends FirmwareUpdateComplete to peripheral.
Peripheral calculates the CRC over the firmware in DFU partition, sends FirmwareCrcReport(dfu_crc)
to central. Central compares that dfu_crc to the firmware CRC from phase 0, if match sends  FirmwareCrcOk to peripheral, peripheral answers FirmwareUpdateConfirm, marks updated to embassy-boot (SWAP_MAGIC) and resets. 

  Central                           Peripheral
    │                                   │
    │ central_crc.finalize()            │
    │ == expected_hash?                 │
    │   ├─ No  → ABORT (binary bug!)    │
    │   └─ Yes → ⬇                      │
    │                                   │
    ├── FirmwareUpdateComplete ────────>│
    │                                   │
    │                     handler.compute_dfu_crc()
    │                     = CRC32(whole DFU partition)
    │                     via flash readback (256B blocks)
    │                                   │
    │<── FirmwareCrcReport(dfu_crc) ────┤
    │                                   │
    │  dfu_crc == expected_hash?        │
    │    ├─ Yes → ⬇                     │
    │    └─ No  → send CrcFail ─────┐   │
    │                               │   │
    │                               │   attempt++ → Phase 1
    │                               │   │
    │    ├── FirmwareCrcOk ────────>│   │
    │    │                          │   │
    │    │                handler.mark_updated_and_reset()
    │    │               (only after CrcOk — never into corrupt FW)
    │    │                          │   │
    │    │<─ FirmwareUpdateConfirm ─┤   │
    │    │                          │   │
    │    │  DONE ── LED OFF         │   │
    │    │                          │   │
    │    └── CrCok in 5s? ──────────│   │
    │         No → timeout → abort  │   │
    │                               │   │
    │    └── CrcFail in 5s? ───── ──┤   │
    │         No → timeout → abort ─┘   │


═══ RETRY SUMMARY ════════════════════════════════════════════════════

  Layer           Max     Trigger                    Consequence
  ─────           ───     ───────                    ──────────
  Per-chunk       3×      Ack CRC mismatch           Re-send same chunk
                          or 2s timeout

  Outer attempt   3×      Chunk never acked          Full restart of
                          or E2E CRC mismatch        Phase 1 + 2
                          or CRC timeout

  No retry        —       Central CRC != expected    Abort — binary
                          (central_crc.finalize())   mismatch, not TX


═══ SAFETY GATES ════════════════════════════════════════════════════

  Where               What                          Why
  ─────               ────                          ───
  Peripheral boot     mark_updated_and_reset()       Never boot into
                      only on FirmwareCrcOk          corrupt firmware

  After all chunks    central_crc == expected_hash   Catch binary bugs
                                                    before asking peripheral

  End-to-end          compute_dfu_crc()             Catch silent flash
                      flash readback                write errors

  Per-chunk           CRC32(data) in Ack            Catch bitflips,
                                                    packet loss,
                                                    wrong offset
```


HaoboGu:
I think it's worth to support external nor flash as the storage used by dfu split. The remaining flash space is quite tight even for nRF52840(which has 1M flash)

Schievel1:
> I think it's worth to support external nor flash as the storage used by dfu split. The remaining flash space is quite tight even for nRF52840(which has 1M flash)

For the embedded peripheral firmware in central or the flashing peripheral via central relay?
For the latter we can not do that, because the bootloader would need access to the external flash. This creates tons of problems with different chips with different flash types so you would have a bouquet of bootloaders supporting different boards. 
For the first, iirc nrf52840 isn't supported for this anyway right now, since dfu_split only supports uart split and RMK does not support nrf uart splits. This works for RP2040 only at the moment, and RP2040 has 2M or more, which is plenty for  firmware that isn't bloated by the BLE stack.

rmk/src/split/driver.rs
```diff
@@ -102,8 +100,16 @@ impl<const ROW: usize, const COL: usize, const ROW_OFFSET: usize, const COL_OFFS
             return;
         }
 
+        #[cfg(feature = "dfu_split")]
+        self.check_firmware_update().await;
```
HaoboGu, line 127: Will it block forever?

examples/use_config/rp2040_dfu_split/rmk-rp2040-dfu-split-peripheral.bin
```diff
@@ -0,0 +1 @@
+AAAABBBBCCCCDDDD
```
HaoboGu, line 1: The test bytes should be removed

Schievel1: The thing is, in the example we need that file to give something to 'include_bytes!()' otherwise it won't compile. 
On the other hand I don't want to check in the actual binary. 

A completely empty file errors out as well. 

This file here will actually cause a boot loop, because it doesn't boot (obviously) and therefore it doesn't call mark_booted(). So the central sends it to the peripheral, then the peripherals bootloader copies it to ACTIVE, tries to boot it, doesn't work, rolls back old ACTIVE, which is not in DFU partition, boots, centrals sees firmware crc doesn't match, sends this again to the peripheral, etc. 
should I mention this somewhere?

rmk/src/split/driver.rs
Schievel1: No it waits for the splits hash with a two second timeout

Schievel1: Ouch I can see a problem here. Let's convert to draft until I fixed it

examples/use_config/rp2040_dfu_split/rmk-rp2040-dfu-split-peripheral.bin
HaoboGu: Ah I thought this bricks the chip. It's fine if the boot loop is expected

rmk/src/split/driver.rs
Schievel1: ok, before we restarted the timeout every time a message came in. It was rather unlikely, but if the central kept getting other messages that it wasn't waiting for it that loop, it could be stuck in that loop until the peripheral stops sending. Now we have a deadline and if we don't get the message we are looking for before the deadline we time out break the loop. 

It was that what you meant here or was it something different?

HaoboGu: yes, that's what I meant. thanks for fixing it!

HaoboGu:
Thanks a lot, this is really cool

examples/use_config/rp2040_dfu_split/rmk-rp2040-dfu-split-peripheral.bin
Schievel1: It can't really brick the chip. If would brick the chip it there wouldn't have been a working firmware running on the chip before flashing this. But flashing this with dfu_split only works with a working firmware. 

Generally speaking (and that is why I chose embassy-boot and its DFU<>ACTIVE swap mechanism) once you have a working firmware running that can use DFU (whether it's USB-DFU directly or via the dfu_split) embassy-boot can always roll back to this working firmware if the newly flashed firmware isn't running. 
This makes sure for dfu_split that even though it can take while for embassy-boot to do the DFU<>ACTIVE swap, you always end up with a working peripheral again that you can try to flash again without taking the keyboard apart.

Schievel1:
> Thanks a lot, this is really cool

Thanks for merging this in. 
I think this really is a unique feature of RMK and no one else has this. I've been dreaming about being able to do this since the very start of my split keyboard career. 😄 
I have a keyboard with a display on the peripheral and no USB connector on the outside. Every time I wanted to try how something looks on the display I had to take the keyboard apart.

PR #889, opened 2026-06-20
feat: add auto mouse layer behavior

kot149:
Added a basic functionality of auto mouse layer behavior: automatically activates a layer when X/Y cursor motion from a pointing device (e.g., PMW3610, trackball) is detected. The layer is deactivated after a `timeout` of inactivity.
Options such as deactivating the layer when a non-mouse key is pressed are planned for a separate PR.

resolves #781

## Example configuration

```toml
[behavior.auto_mouse_layer]
device_id = 1
target_layer = 3
timeout = "600ms"
threshold = 2
```


docs/docs/main/docs/configuration/behavior.md
```diff
@@ -512,3 +512,62 @@ forks = [
...
+));
+
+// Run the auto mouse layer future alongside other tasks
+embassy_futures::join::join(
+    run_all!(
+        matrix,
+        keyboard,
+        // ...other runnables
+    ),
+    rmk::run_auto_mouse_layer_if_enabled(&keymap)
```
Schievel1: Why can't this impl runnable and be part of the run_all!()?

kot149: Fixed in 5b62a0c0efdcddca35f5248988dde651ec9bce4e, thanks for noting that

docs/docs/main/docs/configuration/behavior.md
```diff
@@ -512,3 +512,61 @@ forks = [
...
+```toml
+[behavior.auto_mouse_layer]
+layer = 3
+timeout = "600ms"
+threshold = 2
+```
+
+| Field        | Type    | Default | Description |
+|--------------|---------|---------|-------------|
+| `layer`      | integer | —       | Layer index to activate (must be `>= 1` and `< [layout.layers]`). |
```
Schievel1: why 1 though? l thought layers index starts at 0.
I mean, yes, its kind of wild to use the default layer as the mouse layer but isn't that up to the user?

docs/docs/main/docs/configuration/behavior.md
```diff
@@ -512,3 +512,61 @@ forks = [
...
+::: warning
+
+Use a dedicated layer that is not bound to any manual keys (like `MO` or `TG`). Because layer states are boolean, manually activating the same layer while the auto-mouse timeout is running may cause the layer to deactivate prematurely when the timeout fires.
+
+:::
+
+::: note Pointer Event Configuration
+
+- **Subscriber Slots**: You must manually increment `[event.pointing].subs` (e.g., increase it by `1`) in your `keyboard.toml` to reserve a subscriber slot for the auto mouse layer's background listener task.
+- **Buffer Size**: If pointing events are dropped under high-frequency input, increase `[event.pointing].channel_size` (default is `8`).
```
Schievel1: a link to how to do that would be good I think

docs/docs/main/docs/configuration/behavior.md
```diff
@@ -512,3 +512,61 @@ forks = [
...
+
+::: note Rust API
+If not using `keyboard.toml`, configure the layer via `BehaviorConfig` and run the helper future alongside your other keyboard tasks:
+
+```rust
+use embassy_time::Duration;
+use rmk::config::{AutoMouseLayerConfig, BehaviorConfig};
+
+// Configure the auto mouse layer
+let mut behavior_config = BehaviorConfig::default();
```
Schievel1: I think we can remove this mut and go with Default::default()

docs/docs/main/docs/configuration/behavior.md
```diff
@@ -512,3 +512,61 @@ forks = [
...
+    2,                          // threshold
+));
+
+// include it in `run_all!` alongside the rest:
+let mut auto_mouse_layer = rmk::run_auto_mouse_layer_if_enabled(&keymap);
+run_all!(
+    matrix,
+    keyboard,
+    // ...other runnables
+    auto_mouse_layer,
```
Schievel1, line 599: This works now thanks to my PR. But generally there shouldn't be a comma after the last.

rmk/src/keyboard/auto_mouse_layer.rs
```diff
@@ -0,0 +1,214 @@
...
+/// pointer motion. Pass the returned value to `run_all!`.
+pub fn run_auto_mouse_layer_if_enabled<'a, 'k>(keymap: &'a KeyMap<'k>) -> AutoMouseLayerRunner<'a, 'k> {
+    AutoMouseLayerRunner { keymap }
+}
+
+impl Runnable for AutoMouseLayerRunner<'_, '_> {
+    async fn run(&mut self) -> ! {
+        let keymap = self.keymap;
+        let Some(config) = keymap.auto_mouse_layer_config() else {
+            core::future::pending().await
```
Schievel1: maybe warn here as well, that user is trying to run the task without config and  that its stuck

docs/docs/main/docs/configuration/behavior.md
```diff
@@ -512,3 +512,61 @@ forks = [
...
+
+| Field        | Type    | Default | Description |
+|--------------|---------|---------|-------------|
+| `layer`      | integer | —       | Layer index to activate (must be `>= 1` and `< [layout.layers]`). |
+| `timeout`    | string  | `"500ms"`| Inactivity duration before deactivation (e.g., `"600ms"`, `"2s"`). |
+| `threshold`  | integer | `1`     | Minimum absolute X/Y delta to trigger motion (`>= 1`). Increase to filter sensor noise. |
+
+::: warning
+
+Use a dedicated layer that is not bound to any manual keys (like `MO` or `TG`). Because layer states are boolean, manually activating the same layer while the auto-mouse timeout is running may cause the layer to deactivate prematurely when the timeout fires.
```
Schievel1: can we not also listen to the keyboard driven layer change and when it occurs set the self_activated to false? not sure if getting this from the keyboard is easily possible though

rmk/src/keyboard/auto_mouse_layer.rs
```diff
@@ -0,0 +1,214 @@
...
+        // Tracks whether this task is the one that turned the layer on.
+        let mut self_activated = false;
+        // Whether we have already warned about the auto mouse layer overlapping a
+        // manually-activated layer.
+        let mut overlap_warned = false;
+
+        loop {
+            // Wait for the next pointing event and filter for cursor motion
+            let event = subscriber.next_event().await;
+            if !is_cursor_motion(&event, threshold) {
```
Schievel1: maybe distinguish pointing devices by id as well? people with several pointing devices maybe want to have different layers or different thresholds for them

docs/docs/main/docs/configuration/behavior.md
kot149: I think it depends on personal preference. I am in favor of using trailing commas (as long as they are grammatically allowed), because it helps me avoid forgetting to add a comma when adding new items.

rmk/src/keyboard/auto_mouse_layer.rs
kot149: Since the method is named `run_auto_mouse_layer_if_enabled()` and claims to do nothing if it is disabled, I dont think it is necessary to issue a warning about it.

docs/docs/main/docs/configuration/behavior.md
```diff
@@ -512,3 +512,91 @@ forks = [
...
+
+::: note Event Configuration
+
+- **Subscriber Slots**: Increment both `[event.pointing].subs` and `[event.layer_change].subs` by `1` in your `keyboard.toml` to reserve slots for this task. See [Event Configuration](./event.md).
+- **Buffer Size**: If pointing events are dropped under high-frequency input, increase `[event.pointing].channel_size` (default `8`). `[event.layer_change].channel_size` defaults to `1` and only needs raising if you burst many layer changes faster than subscribers consume them.
+
+:::
+
+::: note Rust API
+If not using `keyboard.toml`, configure the layer via `BehaviorConfig` and run the helper future alongside your other keyboard tasks. Increase `POINTING_EVENT_SUB_SIZE` and `LAYER_CHANGE_EVENT_SUB_SIZE` by `1` for this task's subscribers.
```
Schievel1: How would you do that without changing the rmk/rmk crate? I thought changing the channel sizes and pub/sub limits is done via keyboard.toml regardless of whether the user is using TOML or Rust API.

docs/docs/main/docs/configuration/behavior.md
```diff
@@ -512,3 +512,91 @@ forks = [
...
+
+```rust
+use embassy_time::Duration;
+use rmk::config::{AutoMouseLayerConfig, BehaviorConfig};
+use rmk::heapless::Vec;
+
+// Configure the auto mouse layer. Each entry targets either a specific
+// `device_id` or, when `device_id` is `None`, acts as a fallback for any
```
Schievel1, line 572: I like that approach, but it breaks with the other features of pointing device, where 255 means all devices. I program too much C at work, that's why I come up with those C-ish solutions of giving certain numbers special meanings instead of using Rusts type system and Option. 
Shall we keep it here as you did it and later align where I used 255 to mean all devices?

docs/docs/main/docs/configuration/behavior.md
Schievel1: fair enough, I have the same approach. I usually add the commas when it is multiline. Because `function(foo, baar,)` looks just wrong. 
Just wanted to point out macros are special with commas, because they usually do not grammatically allow trailing commas.

rmk/src/keyboard/auto_mouse_layer.rs
```diff
@@ -0,0 +1,342 @@
...
+}
+
+/// Per-entry runtime state.
+#[derive(Clone, Copy)]
+struct EntryState {
+    config: AutoMouseLayerConfig,
+    /// `true` while this entry is the one holding the layer on.
+    self_activated: bool,
+    /// Set when the entry is self-activated; the next tick fires at this point
+    /// to release ownership if no motion has refreshed it.
```
Schievel1: `the next tick fires at this point to release ownership if no motion has refreshed it.`

I don't understand this. When deadline is reached the layer is deactivated, no?

rmk/src/keyboard/auto_mouse_layer.rs
```diff
@@ -0,0 +1,342 @@
...
+                        );
+                        entry.overlap_warned = true;
+                    }
+                    if entry.self_activated {
+                        entry.deadline = Some(Instant::now() + entry.config.timeout);
+                    }
+                }
+                Tick::Layer(LayerChangeEvent(top)) => {
+                    for entry in entries.iter_mut() {
+                        // Drop ownership if a keyboard-driven change turned our layer off,
```
Schievel1: I think ownership is a concept in Rust and it's borrow checker that does not apply here. And mentioning it here is kind of misleading. How about 
`// Layer was turned off externally — clear our tracking state so the next pointer motion can re-activate it without stale deadline data.`?

Schievel1 (commented):
A few minor things. 
I haven't tested this myself yet, so maybe wait for that before merging.

rmk/src/keyboard/auto_mouse_layer.rs
kot149: You're right, the original wording was confusing. Reworded in 2cd9237.

rmk/src/keyboard/auto_mouse_layer.rs
kot149: Thanks, adopted your suggested wording.

docs/docs/main/docs/configuration/behavior.md
kot149: Good catch, fixed in b41bc59

docs/docs/main/docs/configuration/behavior.md
kot149: Agree.

docs/docs/main/docs/configuration/behavior.md
```diff
@@ -512,3 +512,91 @@ forks = [
...
+| `timeout`      | string  | `"500ms"`| Inactivity duration before deactivation (e.g., `"600ms"`, `"2s"`). |
+| `threshold`    | integer | `1`     | Minimum absolute X/Y delta to trigger motion (`>= 1`). Increase to filter sensor noise. |
+
+Up to `AUTO_MOUSE_LAYER_MAX_NUM` (4) entries are supported. With `keyboard.toml` a build error is raised if more are configured; via the Rust API any entries beyond the limit are silently dropped.
+
+::: warning
+
+Prefer a dedicated layer that is not bound to any manual keys (like `MO` or `TG`). The auto-mouse task releases its ownership when keyboard-driven changes deactivate the layer, so transient overlap is handled cleanly. Layer state is still a single boolean, however, so pressing `TG(target_layer)` while auto-mouse is active toggles the layer off instead of pinning it on.
+
+Also give each entry its own `target_layer`. Layer state is a single boolean shared by all entries, so if two entries point at the same `target_layer`, the first entry's `timeout` can deactivate it while the other device is still moving.
```
HaoboGu: The layer should be deactivated after the last device stops moving?

kot149: Good point, fixed in 95fc117c47cba1105da7fdc0e90e1bcbc11ce839 .

rmk/src/keyboard/auto_mouse_layer.rs
```diff
@@ -0,0 +1,600 @@
...
+}
+
+#[derive(Debug, PartialEq, Eq)]
+enum PointingOutcome {
+    Holding,
+    OverlapFirstSeen,
+    Idle,
+}
+
+impl Runnable for AutoMouseLayerRunner<'_, '_> {
```
HaoboGu, line 136: I think `AutoMouseLayerRunner` should be a `Processor`:


```
#[processor(subscribe = [LayerChangeEvent, PointingEvent])]
pub struct AutoMouseLayerRunner<'a, 'k> {
    keymap: &'a KeyMap<'k>,
}
```

Then move the initialization code to `new()` and implement `on_xxx` event.

But I'm not sure about the timeout event -- maybe it should be emitted somehow, or add an extra function like `deadline() -> Option<Instant>` in `Processor` trait?

kot149: Adopted both in aae61872 and 8e2049a0.

HaoboGu (approved):
Overall it looks good to me. The DeadlineProcessor trait can be refined in a separate PR.

PR #891, opened 2026-06-22
feat(vial): add host.vial_insecure option to start Vial unlocked

hdwong:
Add a `[host] vial_insecure` boolean (default false) that initializes the Vial lock in the unlocked state, so secured operations such as the Matrix Tester are available immediately without pressing the `unlock_keys` combo. Intended for development; production builds should leave it disabled.

The flag flows from `keyboard.toml [host]` through `HostConfig` and the resolved `Host` into the generated `VialConfig`, then into `VialLock::new`, which now initializes `unlocked` from it. Re-locking and unlock-key polling are unchanged, so a host can still lock/unlock a session that started unlocked. The flag only has effect with the `vial_lock` feature.

`VialConfig::new` keeps its three-argument signature (defaulting the new field to false) so existing pure-Rust configs continue to compile.


HaoboGu:
Extra documents should be added to explain this option

PR #892, opened 2026-06-28
feat(ble): add option to disable battery service

### LINKED ISSUE #567: Feature Request: Add Option to Disable BLE Battery Service (for Battery-Free Devices)
### Description

I'm currently playing with my custom ESP32C6 keyboard with rmk firmware. It is powered by the USB and there is no battery supported.
Currently, rmk registers the BLE Battery Service by default, which is not suitable for my use case. If there is an option to disable it, it would be great for me.

### Device Information

**Platform**: ESP32C6_BLE
**Board**: My custom board without battery support
**RMK Version**: 0.7.8

### Current Problem

There is an unwanted "Keyboard is out of battery“ Indicator.

Image

### Expected Behavior

After disabling the Battery Service, the dialog should disappear.

Thanks for the hard work! This firmware's actually fabulous, by the way.

pcasotti:
Closes #567

### DIFF
--- a/rmk-macro/src/codegen/chip/ble.rs
+++ b/rmk-macro/src/codegen/chip/ble.rs
@@ -81,7 +81,7 @@ pub(crate) fn expand_ble_config(hardware: &Hardware) -> (TokenStream2, TokenStre
             } else {
                 (
                     quote! {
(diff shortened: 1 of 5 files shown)


HaoboGu (commented):
Great, thanks!

PR #911, opened 2026-07-06
fix(usb): don't publish Suspended for a never-enumerated USB device

### LINKED ISSUE #910: BLE output still stops on charge-only USB power: never-enumerated device's bus-idle suspend counts as usb_ready() (follow-up to #681 / #829)
## Summary

Follow-up to #681 (closed as "Fixed by refactoring in #829"): the #829 refactor fixed the `enabled`-level part of the bug, but the charge-only-USB scenario still reproduces on nRF52840 at current `main` (6465fedf) through the **suspend** path.

When the keyboard is plugged into USB power **without a data connection** (charge-only cable, wall charger, power bank), all HID output stops: reports are routed to a USB transport whose endpoints were never configured, and silently dropped, while a connected BLE host receives nothing. Unplugging the cable recovers.

| Situation | Expected transport | Actual |
|---|---|---|
| Nothing plugged in | BLE | BLE ✅ |
| USB cable to a host (PC) | USB | USB ✅ |
| USB power only (charger / charge-only cable) | **BLE** | ❌ routed to a dead USB transport |

## Root cause

1. VBUS appears → embassy-usb enables the device → `UsbDeviceHandler::enabled(true)` sets `UsbState::Enabled`. Correct so far — `Enabled` doesn't count as ready (this is what #829 fixed).
2. With no host on D+/D−, the nRF52840 USBD peripheral raises SUSPEND after **3 ms of bus inactivity** ([embassy-nrf `Event::Suspend`](https://github.com/embassy-rs/embassy/blob/embassy-nrf-v0.11.0/embassy-nrf/src/usb/mod.rs#L264-L268)). embassy-usb calls `Handler::suspended(true)` regardless of whether the device was ever enumerated, and RMK publishes `UsbState::Suspended` ([rmk/src/usb/mod.rs#L477-L503](https://github.com/HaoboGu/rmk/blob/6465fedfda53552a1f51e4d695405dfeee77c028/rmk/src/usb/mod.rs#L477-L503)).
3. Since 6c2328b1 ("fix: respect preferred even when USB is suspended"), `usb_ready()` matches `Configured | Suspended` ([rmk-types/src/connection.rs#L66-L69](https://github.com/HaoboGu/rmk/blob/6465fedfda53552a1f51e4d695405dfeee77c028/rmk-types/src/connection.rs#L66-L69)). The rationale — a suspended host must stay routable for remote wakeup — only holds when the device was **configured** before suspending. On a charge-only cable the device went `Enabled → Suspended` without ever being enumerated, yet now counts as a live transport.
4. Routing then always picks USB: while BLE is disconnected `(usb_ready, ble_ready) = (true, false)` → `Usb` unconditionally; once a BLE host connects, `(true, true)` falls back to `preferred`, whose default is `Usb` when nothing is stored ([rmk/src/state.rs#L92-L105](https://github.com/HaoboGu/rmk/blob/6465fedfda53552a1f51e4d695405dfeee77c028/rmk/src/state.rs#L92-L105)).
5. Every report goes to `USB_REPORT_CHANNEL`; the endpoints were never configured, so writes fail with `EndpointError::Disabled` and remote-wakeup attempts are futile. From the outside this looks like "Bluetooth breaks while charging".

Note the regression window: before 6c2328b1, a connected BLE host still won this tie (`suspended_usb_yields_to_connected_ble` test); after it, the default `preferred = Usb` wins even with BLE connected. The tests that replaced it (`suspended_usb_stays_routable_for_remote_wakeup`, `suspended_usb_with_connected_ble_prefers_preference`) only model a previously-configured host — a never-configured suspend wasn't considered.

## Suggested fix

The handler already tracks the missing distinction: `pre_suspend` exists precisely so an Enabled-but-never-Configured device isn't upgraded to `Configured` on resume. The smallest fix is to only publish `Suspended` when the device was `Configured` before the suspend, staying in `Enabled` otherwise:

- charge-only / charger: state stays `Enabled` → `usb_ready() == false` → BLE keeps working;
- genuinely suspended (previously enumerated) host: still published as `Suspended` → remote wakeup (#825) unaffected.

I have this implemented with unit tests and will open a PR; hardware verification on an nRF52840 BLE split board (charger → BLE keeps typing; PC → enumerates and types over USB; sleeping PC → remote wakeup still works) is underway.

## Environment

- nRF52840 (Cornix split, BLE + USB), RMK `main` @ 6465fedf
- embassy-nrf 0.11 / embassy-usb 0.6

numachang:
Fixes #910 (follow-up to #681 / #829).

## Problem

On charge-only USB power (charge-only cable, wall charger, power bank) the device is enabled by VBUS but never enumerated, and the idle D+/D− bus then triggers a suspend — on nRF52840 the USBD peripheral raises SUSPEND after 3 ms of bus inactivity. `UsbDeviceHandler::suspended(true)` published `UsbState::Suspended` unconditionally, and since 6c2328b1 `usb_ready()` treats `Suspended` as routable (so a sleeping host stays reachable for remote wakeup). A never-enumerated device therefore counted as a live USB transport: reports were routed to endpoints that were never configured and silently dropped, while a connected BLE host received nothing. To the user this looks like "Bluetooth breaks while charging".

## Fix

Only publish `Suspended` when the device was `Configured` before the suspend; a never-enumerated device stays in `Enabled`, which doesn't count as `usb_ready()`. The handler already tracked this distinction in `pre_suspend` — this uses it at publish time instead of only on resume.

- Charge-only / charger: state stays `Enabled` → BLE keeps working, and a host that shows up later still enumerates normally.
- Genuinely suspended (previously configured) host: still published as `Suspended` → remote wakeup (#825) and the `preferred` tie-break are unaffected.

## Tests

Four new unit tests in `rmk/src/usb/mod.rs` covering: never-enumerated suspend stays `Enabled` (the regression), configured suspend/resume round-trip, duplicate-suspend snapshot preservation, and out-of-order resume as a no-op. `cargo nextest run --no-default-features --features=split,vial,storage,async_matrix,_ble`: 540 passed.

Hardware verification on an nRF52840 BLE split board (Cornix) is underway: charger → BLE keeps typing; PC → enumerates and types over USB; sleeping PC → remote wakeup. I'll report results on #910.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/rmk/CHANGELOG.md
+++ b/rmk/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fix BLE output stopping while the keyboard is on charge-only USB power (charge-only cable, wall charger, power bank). A never-enumerated device's bus-idle suspend was published as `Suspended`, which `usb_ready()` treats as routable, so reports were routed to USB endpoints that were never configured and silently dropped ([#910](https://github.com/HaoboGu/rmk/issues/910))
 - Fix stuck key when a combo key is re-pressed while the combo is still held. Previously the re-press overwrote the combo output's HID slot, and on combo release the output couldn't be unregistered, leaving the re-pressed key stuck on the host.
 - Fix stuck combo output when overlapping triggered combos share a key (e.g. `M+,` and `,+.` both containing Comma). Releasing the shared key now dispatches the release of every fully-unwound combo output, not just the first.
 - Fix `unregister_keycode` choosing the wrong HID slot when a combo output and another pressed key share a position. Slot lookup now prefers a `(pos, keycode)` match and falls back to keycode-only.
--- a/rmk/src/usb/mod.rs
+++ b/rmk/src/usb/mod.rs
@@ -432,8 +432,10 @@ impl RequestHandler for UsbRequestHandler {
 }
 
 pub(crate) struct UsbDeviceHandler {
-    /// State to restore on resume. Captured at suspend so an Enabled-but-not-yet-Configured
-    /// device that suspends/resumes doesn't get incorrectly upgraded to Configured.
+    /// State to restore on resume. Only a Configured device is ever published as
+    /// Suspended (see `suspended()`), so this always holds Configured while the
+    /// device is suspended; kept as a snapshot rather than a hardcoded value so
+    /// resume stays correct if another pre-suspend state becomes publishable.
     pre_suspend: UsbState,
 }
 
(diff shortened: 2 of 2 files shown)


HaoboGu:
Have you tested/verified the PR? Will it introduce any regression?

numachang:
Yes — verified at three levels:

**Unit tests** — the four new tests in `rmk/src/usb/mod.rs` in this PR cover the handler transitions: a never-enumerated suspend stays `Enabled` (the bug), a configured suspend still publishes `Suspended` and resume restores `Configured`, a duplicate suspend keeps the pre-suspend snapshot, and an out-of-order resume is a no-op. Full suite passes 540/540 (`cargo nextest run --no-default-features --features=split,vial,storage,async_matrix,_ble`).

**Hardware** — running this PR's build on an nRF52840 BLE split board (Cornix): with a charge-only power source (wall charger / power bank) attached, BLE typing keeps working — before the fix it stopped as soon as the cable was plugged in. Normal operation is unaffected. (Also reported in #910.)

**Regression surface** — the only behavior change is the never-`Configured` case: during a bus suspend it stays `Enabled` instead of publishing `Suspended`. A previously-configured (host-suspended) device still publishes `Suspended` exactly as before, so remote wakeup (#825) and the `preferred` tie-break are unchanged — the existing tests covering those (`suspended_usb_stays_routable_for_remote_wakeup`, `suspended_usb_with_connected_ble_prefers_preference`) are untouched and still pass. The proactive-wakeup signal in the USB writer (`current_usb_state() == UsbState::Suspended`) only ever matters for a configured device, which is preserved.

HaoboGu:
I don't see normal USB connection verification

numachang:
Normal USB connection is now hardware-verified as well, on the same board:

- **Central (left half) plugged into the host PC**: enumerates and types over USB. Verified with the host's Bluetooth turned off, so input was definitely flowing over USB, not BLE.
- **Peripheral (right half) plugged into a USB host**: keyboard keeps working normally.
- **Both halves plugged into the host at the same time**: also works.

So with this PR: charge-only power → BLE keeps working (the fix), a real USB host → enumerates and types over USB, and no regressions observed in normal operation.

HaoboGu:
Thanks!

PR #928, opened 2026-07-11
fix(vial): widen user keycode mask to 5 bits so User16-31 don't alias to User0-15

### LINKED ISSUE #918: Vial keycode conversion truncates user keycodes to 4 bits — User16–31 silently alias to User0–15
### Summary

The VIA/Vial keycode conversion masks user keycodes to 4 bits in both
directions, so only 16 of the 32 defined user keycodes survive the Vial
round-trip. User16–User31 silently alias onto User0–User15.

### Environment / versions

- rmk 0.8.2 (crates.io) on rp2040; also confirmed unchanged on `main`
  (checked 2026-07-11) by source inspection — the bug is platform-independent
  (pure keycode conversion).
- Checked the issue tracker for existing reports: none found.

### Steps to reproduce

1. Put `User16` (or any of User16–User31) on a key in the keymap.
2. Open the keyboard in Vial and look at that key: it renders as `User0`
   (0x7E00) — same as a real User0 key. Editing/saving through Vial writes
   User0 back, silently rewriting the keymap.
3. Equivalently, in code: `to_via_keycode` maps 0x850 (`User16`) to 0x7E00,
   identical to 0x840 (`User0`); `from_via_keycode` only accepts
   `0x7E00..=0x7E0F`, so User16–31 can never be produced from the Vial side.

### Affected versions

- rmk 0.8.2 (released): `src/host/via/keycode_convert.rs`
  - to VIA: `k as u16 & 0xF | 0x7E00` (User16 = 0x850 → 0x7E00, same as User0)
  - from VIA: only `0x7E00..=0x7E0F` is accepted
- current `main` (checked 2026-07-11): the representation changed to
  `Action::User(u8)`, but the conversion still masks with `& 0xF` and the
  reverse arm is still `0x7E00..=0x7E0F`, so ids ≥ 16 alias to 0–15.

### Impact

1. Only User0–User15 can be assigned from Vial.
2. Worse: if a firmware keymap contains User16–User31, viewing/editing the
   keymap in Vial silently rewrites them as User0–User15 (the forward
   conversion aliases them, and the reverse conversion can only produce 0–15).
   No warning is emitted.

This is inconsistent with the type definitions: rmk-types defines
`User0..=User31` (0x840..0x85F) and `is_user()` covers all 32; on `main`,
`Action::User(u8)` has no stated upper bound at all.

### Expected behavior

All 32 user keycodes round-trip through Vial. QMK's protocol has room for
this: `QK_KB_0..QK_KB_31 = 0x7E00..0x7E1F` (the QK_KB range extends to
0x7E3F), and rmk's own comment at the reverse arm says "QK_KB_N, aka UserN".

### Suggested fix

Widen the mask and the accepted range symmetrically:

```rust
// to VIA
k as u16 & 0x1F | 0x7E00
// from VIA
0x7E00..=0x7E1F => { /* ... & 0xFF | 0x840, unchanged */ }
```

(and the equivalent `& 0x1F` for `Action::User(id)` on `main`).

For what it's worth: the 32-entry `User0..=User31` range has been present since
the first published rmk-types release (0.1.0), so this looks like a day-one
mask-width typo (`0xF` vs `0x1F`) rather than a leftover from an older
16-entry range.

I'm happy to open a PR for either branch if this direction is acceptable.

### Context

We (Techmechkeys) are moving our keyboard firmware from QMK to RMK and plan
to use user keycodes for on-keyboard settings switching, which is how we hit
this. Found by source inspection on 0.8.2 and confirmed still present on
`main`; happy to provide any further detail.

techmech-keeb:
Fixes #918

The VIA/Vial conversion masked user keycodes with `0xF` in both directions, so only 16 of the 32 defined user keycodes survived the Vial round-trip: User16–31 aliased onto User0–15 when displayed, and editing such a key in Vial silently wrote the aliased value back.

### Changes

- Widen the forward mask to `0x1F` and the accepted reverse range to `0x7E00..=0x7E1F`, matching QMK's `QK_KB_0..QK_KB_31`.
- Add `User0` / `User16` / `User31` round-trip cases to both conversion tests (`User16` must not alias to `User0`'s 0x7E00).
- CHANGELOG entry under Unreleased → Fixed.

### Testing

Ran the full `rmk` test suite with the CI feature set (`--no-default-features --features split,vial,storage`, via cargo-nextest): **530/530 passed**, including the new round-trip cases.

### DIFF
--- a/rmk/CHANGELOG.md
+++ b/rmk/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fix Vial keycode conversion truncating user keycodes to 4 bits: `User16`–`User31` silently aliased to `User0`–`User15` on the Vial side (assign, view, and save-back). Widen the mask and accepted range to 5 bits (`0x7E00..=0x7E1F`, matching QMK's `QK_KB_0..QK_KB_31`) ([#918](https://github.com/HaoboGu/rmk/issues/918))
 - Fix BLE output stopping while the keyboard is on charge-only USB power (charge-only cable, wall charger, power bank). A never-enumerated device's bus-idle suspend was published as `Suspended`, which `usb_ready()` treats as routable, so reports were routed to USB endpoints that were never configured and silently dropped ([#910](https://github.com/HaoboGu/rmk/issues/910))
 - Fix stuck key when a combo key is re-pressed while the combo is still held. Previously the re-press overwrote the combo output's HID slot, and on combo release the output couldn't be unregistered, leaving the re-pressed key stuck on the host.
 - Fix stuck combo output when overlapping triggered combos share a key (e.g. `M+,` and `,+.` both containing Comma). Releasing the shared key now dispatches the release of every fully-unwound combo output, not just the first.
--- a/rmk/src/host/via/keycode_convert.rs
+++ b/rmk/src/host/via/keycode_convert.rs
@@ -95,7 +95,7 @@ pub(crate) fn to_via_keycode(key_action: KeyAction) -> u16 {
                     0
                 }
             },
-            Action::User(id) => (id as u16 & 0xF) | 0x7E00,
+            Action::User(id) => (id as u16 & 0x1F) | 0x7E00,
             _ => {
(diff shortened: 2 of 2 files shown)


HaoboGu (approved):
Thanks!

PR #933, opened 2026-07-12
feat: deactivate_on_key option for auto-mouse-layer

kot149:
Added the following options to auto mouse layer behavior:

- `deactivate_on_key`: When `true`, pressing any non-mouse key immediately deactivates `target_layer` (ignoring `timeout`).
- `extra_mouse_keys`: Extra keycodes (e.g. `"LCtrl"`, `"Space"`) treated like mouse keys for the purpose of `deactivate_on_key`.
- `reset_timeout_on_key`: When `true`, key presses that do NOT deactivate `target_layer` push the `timeout` deadline forward (reset it to *now + `timeout`*).

## Example configuration

```toml
[[behavior.auto_mouse_layer]]
target_layer = 5
timeout = "1000ms"
threshold = 1
deactivate_on_key = true
extra_mouse_keys = ["LCtrl", "LShift", "LAlt", "LGui"]
reset_timeout_on_key = true
```


HaoboGu (commented):
Is it worth to add an extra `ResolvedAction`? A plain `match` works with no difference

HaoboGu:
And the RAM consumption increases almost 864bytes, which is a little bit more than expected.

kot149:
> Is it worth to add an extra `ResolvedAction`? A plain `match` works with no difference

Agreed, it wasn't worth the extra enum. Removed ResolvedAction and resolve_action; keypress_step now matches on Action directly.

> RAM consumption increases almost 864bytes, which is a little bit more than expected

The capacity of `AUTO_MOUSE_LAYER_EXTRA_MOUSE_KEYS_MAX_NUM = 32` was excessive, so I reduced it to 8 and the RAM consumption increase is now +480 bytes. Most of the remainder is the `ActionEvent` channel I think.

HaoboGu:
The RAM consumption still worries me. 480 bytes is fine for MCUs like nRF52840, but it is too much for MCUs like stm32f103. I believe that it's very rare to add pointing devices and auto-mouse-layer to a stm32f103. Is it possible to make this cost optional? Like only devices that adds pointing device/mouse pay extra RAM for it.

kot149:
I've made the extra RAM opt-in so boards that don't use these options don't pay for them. With no auto-mouse-layer / no `deactivate_on_key` / `reset_timeout_on_key`, the ActionEvent channel and the extra-key buffer are not allocated.

HaoboGu (approved):
lgtm, thanks!

PR #939, opened 2026-07-13
feat: add LatchTap keyboard action

braindefender:
`LatchTap(modifier, key)` is a keymap action that **latches** a modifier for the lifetime of the current layer and **taps a key under it on each press**.

> Unlike `LM(layer, modifier)`, the modifier is **not** bound to the layer-switch key. It is bound to this dedicated key, so you can place several independent `LatchTap` keys (with different modifiers/keys) on the **same** layer.


HaoboGu:
This behavior is quite similar with what #859 does, but narrower. What do you think the way proposed by #859?

braindefender:
> This behavior is quite similar with what #859 does, but narrower. What do you think the way proposed by #859?

Yeah, I will take a closer look at #859. Maybe it will cover all the cases.

I see potential problem in combining OSM/SK into one behavior:
- The behavior is configured once, but some users may want to combine different settings for different keys. How will this be handled?
- For example, in my LatchTap implementation, the layer change is releasing the latched mod (similar to `release_on_layer_change` behavior setting). But what if I want that behavior **only** in LatchTap keys and not in OSM?

PR #950, opened 2026-07-15
docs(rynk): generate the protocol reference from the command tables

HaoboGu:
Generate a human-readable Rynk protocol reference from the command tables, so the docs can't drift from the wire contract.

## What it does

- New `ENDPOINT_META` / `TOPIC_META` tables, emitted by the existing `endpoints!` / `topics!` macros — each row's id, name, request/response (or payload) types, `@bulk`/`cfg` feature gate, and doc comment, as data.
- A rendering test in `rmk-types` turns those into `docs/development/rynk_protocol.md` (frame layout, endpoint table, topic table, compatibility rules). It runs `UPDATE_SNAPSHOTS`-style: a diff fails CI, and regenerating is one command — the same idiom as the existing `wire_*.snap` golden files.
- The tables are feature-independent, so every feature set renders identical output; the reference documents the full protocol and shows each command's gating feature in a column.

## Scope

Additive to `rmk-types` only — no protocol, wire-format, firmware, or host changes. Just `command.rs` (the meta tables), `tests.rs` (the generator + drift gate), and the generated `.md`.

## Testing

`rmk-types` default / `host` / `steno` all green (wire snapshots unchanged); the `protocol_reference` drift test passes under both `rynk` and `rynk,bulk,_ble,split`; `rmk` and the `rynk` workspace still build against the additive change; `cargo +nightly fmt --check` clean.

### DIFF
--- a/docs/docs/main/docs/development/rynk_protocol.md
+++ b/docs/docs/main/docs/development/rynk_protocol.md
@@ -1,0 +1,90 @@
+<!-- GENERATED — do not edit. Rendered from the `endpoints!`/`topics!` tables in
+     rmk-types/src/protocol/rynk/command.rs. Regenerate with:
+     UPDATE_SNAPSHOTS=1 cargo test -p rmk-types --features rynk protocol_reference -->
+
+# Rynk Protocol Reference
+
+Current protocol version: **0.1**.
+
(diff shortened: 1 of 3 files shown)


HaoboGu:
@codex review

PR #962, opened 2026-07-16
Rynk Protocol

HaoboGu:
successor of #848, cleaned the commit history


colonelpanic8:
I tested the Rynk USB path from this PR at `aa3398ea` on a Glove80 (nRF52840, split keyboard).

CDC OUT worked: the firmware received and decoded requests. CDC IN did not complete on the host, even though the firmware's write future returned successfully. I tried asserting DTR, starting the host read before sending, direct libusb access, 64-byte packets, and changing interface/endpoint order and numbers; none changed the result.

Using Rynk's existing 32-byte HID framing on the same build worked reliably. I could read all eight 6×14 keymap layers, and the split central and peripheral both remained connected.

This may be specific to the nRF52840/Embassy CDC path rather than Rynk itself, so I don't think the PR should simply replace CDC everywhere. Would a selectable USB HID transport be useful here? I have the firmware and native Linux host changes working and can extract a small patch if that direction makes sense.

HaoboGu:
> I tested the Rynk USB path from this PR at `aa3398ea` on a Glove80 (nRF52840, split keyboard).
> 
> CDC OUT worked: the firmware received and decoded requests. CDC IN did not complete on the host, even though the firmware's write future returned successfully. I tried asserting DTR, starting the host read before sending, direct libusb access, 64-byte packets, and changing interface/endpoint order and numbers; none changed the result.
> 
> Using Rynk's existing 32-byte HID framing on the same build worked reliably. I could read all eight 6×14 keymap layers, and the split central and peripheral both remained connected.
> 
> This may be specific to the nRF52840/Embassy CDC path rather than Rynk itself, so I don't think the PR should simply replace CDC everywhere. Would a selectable USB HID transport be useful here? I have the firmware and native Linux host changes working and can extract a small patch if that direction makes sense.

I think it might be a linux specific descriptor bug.

PR #964, opened 2026-07-16
fix(via): reject oversized macro set-buffer to avoid out-of-bounds panic

### LINKED ISSUE #963: Out-of-bounds panic in VIA DynamicKeymapMacroSetBuffer from host-controlled size byte
### Summary

The `DynamicKeymapMacroSetBuffer` handler in `rmk/src/host/via/mod.rs` reads a `size` byte straight from the host packet (`report.output_data[3]`, a `u8`) and then slices `report.output_data[4..4 + size]`. `output_data` is `[u8; 32]`, so any `size >= 29` slices past the end and panics:

```
range end index 33 out of range for slice of length 32
```

That panic takes down the host service task.

### Why it slips through

The handler even writes down the bound it never checks. The comment right above the read says `// Current sequence size, <= 28`, but nothing enforces it. The sibling `DynamicKeymapMacroGetBuffer` handler does guard the same limit (`if size <= 28 { ... } else { report.input_data[0] = 0xFF }`). SetBuffer just missed it.

### Reachability and severity

VIA/Vial commands come in as HID feature reports over USB (and BLE), so a malformed or malicious host packet with `output_data[3] >= 29` panics the firmware. It's a robustness/DoS issue, not remote and not memory corruption (Rust panics safely), but a single crafted host report is enough to trigger it.

I have a fix that mirrors the `GetBuffer` guard (reject oversized writes with `0xFF`) plus a regression test that drives `process_via_packet`. PR incoming.

LordAizen1:
Fixes #963.

`DynamicKeymapMacroSetBuffer` reads a `size` byte straight from the host packet (`report.output_data[3]`, a `u8`) and slices `report.output_data[4..4 + size]`. `output_data` is `[u8; 32]`, so any `size >= 29` slices past the end and panics ("range end index 33 out of range for slice of length 32"), which takes down the host service. A malformed or malicious VIA/Vial feature report over USB or BLE can trigger it.

The handler even writes down the bound it never checks. The comment right above the read says:

```rust
// Current sequence size, <= 28
let size = report.output_data[3];
```

but nothing enforces it. The sibling `DynamicKeymapMacroGetBuffer` handler does guard the same limit (`if size <= 28 { ... } else { report.input_data[0] = 0xFF }`). SetBuffer just missed it.

This mirrors the GetBuffer guard: oversized writes are rejected with `0xFF` instead of panicking, and the `offset == 0` reset semantics stay the same.

Adds a test that drives `process_via_packet`: `size = 28` writes the max payload, `size = 29` is rejected (it panicked before this change).

Note: `offset` is unchecked on both Get and Set, but `read/write_macro_buffer` clamp internally, so that one isn't a live panic. Left it for a separate change.

### DIFF
--- a/rmk/src/host/via/mod.rs
+++ b/rmk/src/host/via/mod.rs
@@ -160,17 +160,24 @@ impl<'a> VialService<'a> {
                 let offset = BigEndian::read_u16(&report.output_data[1..3]);
                 // Current sequence size, <= 28
                 let size = report.output_data[3];
-                // End of current sequence in the macro cache
-                // The first sequence, reset the macro cache
-                if offset == 0 {
-                    self.ctx.reset_macro_buffer();
-                }
+                // `output_data` is 32 bytes, so the payload slice output_data[4..4 + size]
+                // is only valid for size <= 28. Reject oversized writes instead of
+                // panicking, mirroring the DynamicKeymapMacroGetBuffer handler above.
+                if size <= 28 {
+                    // End of current sequence in the macro cache
+                    // The first sequence, reset the macro cache
+                    if offset == 0 {
+                        self.ctx.reset_macro_buffer();
+                    }
 
-                // Update macro cache + flush full buffer to storage
-                info!("Setting macro buffer, offset: {}, size: {}", offset, size);
-                self.ctx
-                    .write_macro_buffer(offset as usize, &report.output_data[4..4 + size as usize])
-                    .await;
+                    // Update macro cache + flush full buffer to storage
+                    info!("Setting macro buffer, offset: {}, size: {}", offset, size);
+                    self.ctx
+                        .write_macro_buffer(offset as usize, &report.output_data[4..4 + size as usize])
+                        .await;
+                } else {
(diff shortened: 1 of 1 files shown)


HaoboGu:
I don't think it will happen in practice, unless the host software sends a malformed packet

LordAizen1:
Yeah, a normal VIA/Vial host will always keep size <= 28. My worry is the malformed/malicious case, since output_data comes straight off the USB/BLE HID endpoint and any host process can write whatever it wants into a feature report.
Two reasons I still think it’s worth guarding:
	1.	Right now it panics instead of just ignoring the bad packet, and a panic in the host service is a worse outcome than dropping one malformed request.
	2.	The GetBuffer handler already rejects size > 28 with input_data[0] = 0xFF, so this just makes SetBuffer match what the code does one branch over. Not a new concept, just closing the path that missed it.
Happy to make it a debug_assert or trim the test if you’d rather a lighter touch.

HaoboGu:
yeah, adding it wouldn't do any harm. Thanks for the contribution!

PR #978, opened 2026-07-20
fix(split): defer BLE DFU hash announcement until link is ready

### LINKED ISSUE #977: BLE split peripheral can drop proactive DFU firmware-hash announcement
## Reproducer

Tested on a MoErgo Glove80 nRF52840 split keyboard with RMK commit
`1156f82b` and Rust 1.97.0. Both halves were built and flashed with BLE split
DFU enabled:

```toml
rmk = { features = ["nrf52840_ble", "split", "dfu_nrf", "dfu_split"] }
```

The keyboard configuration used BLE for the split connection and configured
the normal central/peripheral BLE addresses.

1. Flash both halves with that configuration.
2. Boot both halves and allow the split BLE link to connect or reconnect.
3. Capture the central and peripheral trace from connection startup.
4. Observe that the proactive firmware-hash announcement does not arrive at
   the central in the tested session.

The trace shape is: the peripheral attempts its startup
`FirmwareHashResponse` before the central subscribes to the notify
characteristic; the central then completes the CCCD subscription and sends its
initial `ConnectionStatus`, but no proactive `FirmwareHashResponse` appears in
the central's receive loop.

## Root cause

`SplitPeripheral::run` writes `FirmwareHashResponse` immediately at session
startup. On the tested nRF52840 BLE connection, that write occurs before the
central has subscribed to the peripheral's notify characteristic by writing
its CCCD.

RMK pins TrouBLE at `c21b1239`. In that revision,
[`Characteristic::notify_raw`](https://github.com/embassy-rs/trouble/blob/c21b1239fb86de712fc99da803def88378d81177/host/src/attribute.rs#L1019-L1035)
returns `Ok(())` when `should_notify` is false, without transmitting a
notification. RMK therefore observes a successful write even though this
pre-subscription announcement was dropped in the captured session.

## Proposed fix

For BLE only, defer the one-shot proactive hash response until the first
successfully decoded message arrives from the central. The central sends its
initial connection status after subscribing, so that message provides the
readiness barrier. Preserve the immediate announcement on serial split links,
and mark the BLE announcement complete only after a successful driver write so
transient errors remain retryable.

A minimal fix with fake-driver regression coverage is ready. I will link the
pull request here once it is opened.

colonelpanic8:
Fixes #977.

On a BLE split, the peripheral can try to send its firmware hash before the central enables notifications. TrouBLE returns success in that case even though nothing is transmitted, so RMK treats the announcement as complete and the central never sees it.

For BLE, this waits until the first decoded message from the central before making the proactive announcement. Serial split links keep the existing immediate behavior. Failed announcements remain retryable.

Validated with the focused BLE and serial nextest cases, plus the repository format, check, Clippy, full test, and doctest scripts.

### DIFF
--- a/rmk/src/split/peripheral.rs
+++ b/rmk/src/split/peripheral.rs
@@ -24,6 +24,66 @@ use crate::event::{ModifierEvent, SleepStateEvent, WpmUpdateEvent};
 use crate::split::serial::SerialSplitDriver;
 use crate::state::update_status;
 
+#[cfg(any(feature = "dfu_split", test))]
+trait FirmwareHashWriter {
+    async fn write_firmware_hash(&mut self, hash: u32) -> bool;
+}
+
+#[cfg(feature = "dfu_split")]
+impl<S: SplitWriter> FirmwareHashWriter for S {
+    async fn write_firmware_hash(&mut self, hash: u32) -> bool {
+        self.write(&SplitMessage::FirmwareHashResponse(hash)).await.is_ok()
+    }
+}
+
+#[cfg(any(feature = "dfu_split", test))]
+struct FirmwareHashAnnouncement {
+    ready: bool,
+    complete: bool,
+}
+
+#[cfg(any(feature = "dfu_split", test))]
+impl FirmwareHashAnnouncement {
+    const fn new(wait_for_message: bool) -> Self {
+        Self {
+            ready: !wait_for_message,
+            complete: false,
+        }
+    }
+
+    async fn try_announce<W: FirmwareHashWriter>(&mut self, writer: &mut W, hash: u32) {
+        if self.ready && !self.complete && writer.write_firmware_hash(hash).await {
+            self.complete = true;
+        }
+    }
+
+    fn message_received(&mut self) {
+        self.ready = true;
+    }
+}
+
+async fn prepare_inbound_message<W: SplitWriter>(
+    writer: &mut W,
+    #[cfg(feature = "dfu_split")] announcement: &mut FirmwareHashAnnouncement,
+    message: SplitMessage,
+) -> SplitMessage {
+    #[cfg(all(feature = "dfu_split", feature = "_ble"))]
+    {
+        // A decoded BLE message proves that the central has completed its GATT
+        // subscription and can receive notifications.
+        announcement.message_received();
+        announcement
+            .try_announce(writer, crate::dfu::read_embedded_firmware_hash())
+            .await;
+    }
+    #[cfg(not(all(feature = "dfu_split", feature = "_ble")))]
+    let _ = writer;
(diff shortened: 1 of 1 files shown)


HaoboGu:
This is not a good design, I think to fix it in upstream TrouBLE stack is the better way. An alternative way is to send the message after the central completes the CCCD subscription.

colonelpanic8:
Agreed — after looking at this more closely, the current PR's `FirmwareHashAnnouncement` state machine and “first decoded message” proxy put BLE transport readiness in the generic split/DFU flow, which is the wrong layer.

I checked TrouBLE before changing anything further:

- RMK's pinned `c21b1239` is currently identical to TrouBLE `main`, so a revision bump alone cannot change this behavior.
- `Characteristic::notify_raw` deliberately returns `Ok(())` when the peer has not enabled the CCCD, even though no notification is sent.

I've opened embassy-rs/trouble#649 to make the single-connection `notify`/`indicate` APIs return a specific `Error::NotSubscribed` in that case: https://github.com/embassy-rs/trouble/pull/649

That fixes the misleading success result upstream, but it cannot make this one-shot announcement reliable by itself: RMK currently discards the startup write result, and TrouBLE cannot decide which application message should be retried or when.

For the RMK side, I agree your CCCD suggestion is cleaner. I'll rework this so BLE readiness is handled in the BLE peripheral driver: detect and accept the `message_to_central` CCCD subscription, then trigger the firmware-hash announcement. That removes the generic first-message proxy/state machine and leaves serial behavior unchanged.

HaoboGu:
https://github.com/embassy-rs/trouble/pull/649#discussion_r3612532057

Same for this PR

colonelpanic8:
> Same for this PR

okay but even with the upstream fix, we still need to handle things here.

as I mentioned above:

> That fixes the misleading success result upstream, but it cannot make this one-shot announcement reliable by itself: RMK currently discards the startup write result, and TrouBLE cannot decide which application message should be retried or when.
> 
> For the RMK side, I agree your CCCD suggestion is cleaner. I'll rework this so BLE readiness is handled in the BLE peripheral driver: detect and accept the `message_to_central` CCCD subscription, then trigger the firmware-hash announcement. That removes the generic first-message proxy/state machine and leaves serial behavior unchanged.

Schievel1:
I think this PR is trying to solve a problem that doesn't exist. dfu_split is not available for BLE and I think it never will be. Because once we get OTA flashing via BLE from the host to the peripheral (as I have it in my dfu_ble PR) dfu_split for BLE split keyboards is unnecessary. 

If anything that PR should disable the proactive firmware hash if the transport is BLE so we could save a message and some cycles of CPU time. On the other hand, if you have split ble you could just not activate dfu_split as it doesn't give you any benefits anyway.

PR #981, opened 2026-07-20
Use portable Bash shebangs

colonelpanic8:
Changes the executable Bash scripts in `.github/ci` and `scripts/fetch_all.sh` to use `#!/usr/bin/env bash`.

NixOS does not provide `/bin/bash`, so invoking these scripts directly currently fails even when Bash is available on `PATH`.

Validation:
- `bash -n` on all changed scripts
- `git diff --check`
- Direct `.github/ci/discover.sh` run (48 stable, 2 ESP, 16 bloat examples)


HaoboGu:
thanks

PR #984, opened 2026-07-20
Add bounded split application messages

colonelpanic8:
## What

- add a 26-byte opaque application payload to the split wire protocol
- expose bounded central-to-peripheral, peripheral-to-central, and receive queues
- expose split-link state so applications can resynchronize after reconnects
- keep key events ahead of application traffic and use drop-on-full semantics to preserve split responsiveness

## Why

RMK-based firmware sometimes needs to coordinate application-owned state across halves without adding hardware-specific meaning to the core split protocol. Per-key lighting is one consumer, but this API stays opaque and allocation-free.

## Impact

Existing split behavior is unchanged unless the new queues are used. The implementation currently targets a single split peripheral and documents that limitation.

## Checks

- repository formatter
- Rynk host tests, doctests, WASM checks/package typecheck, and clippy

### DIFF
--- a/rmk/src/lib.rs
+++ b/rmk/src/lib.rs
@@ -110,6 +110,9 @@ pub mod matrix;
 pub mod processor;
 #[cfg(feature = "split")]
 pub mod split;
+// Bounded application-message hook for the split protocol.
+#[cfg(feature = "split")]
+pub mod split_app;
 pub mod state;
 #[cfg(feature = "storage")]
 pub mod storage;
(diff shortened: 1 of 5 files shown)


Schievel1:
Iirc this is also useful to transport state of the keyboard to the peripheral that is not available in the usual split protocol right? 

Like for example I have a special function in my firmware that moves the mouse pointer a bit automatically when it's active. And I have a display on the peripheral that should display whether it's active.
Could I use this PR for that?

HaoboGu:
what is this message used for?

Schievel1:
Would be good to have some docs on how to use this.

PR #985, opened 2026-07-20
Expose custom host and runnable macro hooks

colonelpanic8:
## What

- allow `#[Overwritten(HostService)]` to construct an application-defined host service
- add `#[register_processor(runnable)]` for `Runnable` components
- initialize registered processors after keymap and storage setup so constructors can use authoritative keymap state

## Why

Applications that compose RMK services need a supported macro-level extension point rather than copying the generated orchestrator. This keeps custom services and long-running processors inside the normal `#[rmk_keyboard]` lifecycle.

## Impact

Existing generated code is unchanged unless the new annotations are used. Event and polling processor modes retain their current behavior.

## Checks

- repository formatter
- Rynk host tests, doctests, WASM checks/package typecheck, and clippy

### DIFF
--- a/rmk-macro/src/codegen/orchestrator.rs
+++ b/rmk-macro/src/codegen/orchestrator.rs
@@ -470,11 +470,12 @@ fn expand_main(
             // Set all keyboard config
             #rmk_config
 
-            // Initialize the registered processors
-            #registered_processor_initializers
-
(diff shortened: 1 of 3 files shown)


HaoboGu:
I don't see a reason to overwrite host

PR #986, opened 2026-07-20
Use HID for the Rynk USB transport

colonelpanic8:
## What

- replace the USB CDC-ACM Rynk interface with a vendor-HID interface
- fragment Rynk frames into the same fixed 32-byte reports used by the BLE WebHID transport
- trim report padding from the frame length before bytes reach `RynkService`
- select WebHID in the WASM driver while retaining Web Serial support for compatible callers

## Why

CDC framing depends on short packets or ZLP behavior and was not reliable on the qualified Glove80 USB path. Fixed HID reports give Rynk a transport boundary that browsers and firmware handle consistently.

## Validation

This transport has been exercised on physical Glove80 hardware in addition to:

- repository formatter
- Rynk native tests and doctests
- WASM target checks, generated package TypeScript validation, and clippy

### DIFF
--- a/rmk/src/host/mod.rs
+++ b/rmk/src/host/mod.rs
@@ -7,11 +7,12 @@ pub(crate) mod lock;
 pub(crate) mod rynk;
 #[cfg(feature = "storage")]
 pub(crate) mod storage;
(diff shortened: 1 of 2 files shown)


HaoboGu:
won't fix

PR #987, opened 2026-07-20
Add topology-aware lighting with Rynk control

colonelpanic8:
## What

- add physical key topology and generated lighting configuration
- add a composable lighting engine with selectors, sources, effects, layers, output mapping, and RMK state integration
- add native Rynk lighting discovery, query, mutation, and topic support
- add Vial RGB Matrix compatibility backed by the same authoritative lighting state
- add protocol snapshots, loopback coverage, and focused unit tests

## Why

RMK currently has no general model for per-key lighting that can span irregular and split layouts while keeping firmware, Rynk, and Vial state consistent. This implementation makes one lighting service authoritative, so changes made by either host protocol are reflected by later state queries and rendered output.

## Dependency

This branch is stacked on #984 because split keyboards use the bounded application-message channel to propagate lighting state. Once #984 lands, this PR can be rebased to contain only the lighting commits.

This is the complete implementation that supersedes the smaller, closed abstraction draft in #982.

## Checks

- repository formatter
- full RMK feature-matrix tests, including Rynk lighting loopback and Vial RGB Matrix tests
- Rynk native tests and doctests
- WASM target checks and generated package TypeScript validation
- workspace and WASM clippy with warnings denied
- physical Glove80 lighting and host-state round trips exercised on the composed integration branch


colonelpanic8:
Superseded by #1031, which is the same lighting work rebased onto `main`. This PR targeted `feat/rynk`, which upstream merged into `main` in 9448f7a1, so it could no longer be diffed against a live base and showed spurious conflicts. #1031 merges cleanly.

PR #988, opened 2026-07-20
Add application-defined Rynk build information

colonelpanic8:
Adds an additive `GetBuildInfo` endpoint whose bounded display label is deliberately separate from the structured Rynk protocol version. RMK advertises `RMK v<version>` by default; downstream firmware can override it with `HostService::with_build_label`, and the RMK version string is exported for composing richer labels. Includes native and Wasm client APIs, wire snapshots, generated protocol documentation, and loopback coverage.


HaoboGu:
Use GetDeviceInfo

PR #993, opened 2026-07-23
Docs: Suggest SHIFTED and WM after modifiers keycodes list

peterjc:
This addresses my query on Discord, having failed to find this in the docs last night.


HaoboGu (approved):
THanks!

PR #999, opened 2026-07-23
Expose held modifier combinations in keyboard.toml

colonelpanic8:
## What changed

- Add `MOD(modifier)` to the `keyboard.toml` grammar and action parser.
- Expand it to RMK’s existing `Action::Modifier`, without changing `WM(key, modifier)` semantics.
- Document the syntax and test modifier combinations, case-insensitive parsing, and use inside a nestable action slot.

## Motivation

A dedicated modifier-combination key (for example, Hyper) needs to hold several modifiers while arbitrary other keys are pressed. `WM` describes a self-contained key-with-modifiers action, while `MOD` directly expresses the intended held-modifier action.

```toml
MOD(LCtrl | LAlt | LGui)
```

## Validation

- `cargo fmt` / formatting checks
- `rmk-config`: 92 unit tests and 6 integration tests passed
- focused `rmk-macro` parsing and expansion tests passed
- Glove80 `thumbv7em-none-eabihf` firmware check passed using a real `MOD(LCtrl | LAlt | LGui)` configuration

### DIFF
--- a/docs/docs/main/docs/configuration/layout.md
+++ b/docs/docs/main/docs/configuration/layout.md
@@ -104,6 +104,8 @@ The `layer.keys` string should follow several rules:
 
    For simple keycodes with modifiers active, you can use `WM(key, modifier)` to create a keypress with modifier action. Modifiers can be chained together like `LShift | RGui` to have multiple modifiers active.
 
+   To hold one or more modifiers without sending a non-modifier key, use `MOD(modifier)`. For example, `MOD(LCtrl | LAlt | LGui)` holds all three modifiers until the bound key is released.
+
    You may use aliases, prefixed with `@`, like `@my_copy` in the above example. The alias names are case sensitive. The definition of aliases is described below.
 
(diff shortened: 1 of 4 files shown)


HaoboGu:
In what situation you bind this `WM(LCtrl, LAlt | LGui)`?

colonelpanic8:
The concrete use case is a dedicated Hyper key: hold Ctrl+Alt+GUI, then press arbitrary other keys that Hammerspoon interprets as shortcuts.

Looking at it again, encoding that as `WM(LCtrl, LAlt | LGui)` was the wrong abstraction. RMK already has `Action::Modifier` for exactly this behavior, but `keyboard.toml` had no way to express it. I revised the PR to add `MOD(LCtrl | LAlt | LGui)` as first-class syntax and removed the `WM` runtime special case entirely. `WM` semantics are unchanged.

HaoboGu (approved):
thanks!

PR #1000, opened 2026-07-24
Add support for USB logging on peripheral

tompre:
Adds usb init function and logging task for the peripheral depending on the `usb_log` feature flag.
Should this be a separate feature?

Currently this works for nrf52 only, should I add support for rp2040?

Currently peripheral usb logging is mutually exclusive with dfu, should I create a common usb_task for the two? Would that even work with the available hardware resources?

It changes the default log level to trace, so one can change the log level without touching rmk code. I added docs for this.

Let me know if I should change anything.

### DIFF
--- a/docs/docs/main/docs/features/usb_logging.md
+++ b/docs/docs/main/docs/features/usb_logging.md
@@ -14,6 +14,13 @@ rmk = { version = "0.9", default-features = false, features = [
 ] }
 ```
 
+The default log level is debug. You can change the log level by enabling the `max_level_*` feature for the log crate in `Cargo.toml`:
+```toml
+log = {version = "0.4", features = [
+    "max_level_info",
+]}
+```
+
 ::: tip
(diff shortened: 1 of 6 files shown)


HaoboGu:
Thanks for the work! Comments from me:

1. The default log level should be set to info/debug? trace is quite verbose sometimes
2. Yes, it would be great to have it on rp2040
3. The USB instance can be shared, USB-DFU and logging are different endpoints, so they can co-exist.

tompre:
New version:
> 1. The default log level should be set to info/debug? trace is quite verbose sometimes

The log level filter in embassy_usb_logger is set to trace, so all levels are supported without changing the code.
The default log level is set to debug in rmk/Cargo.toml via feature and can be overridden by the user as explained in the docs.
> 2. Yes, it would be great to have it on rp2040

I think I should work for rp2040 too, I can not test it though.
> 3. The USB instance can be shared, USB-DFU and logging are different endpoints, so they can co-exist.

I combined the logic of `run_peripheral_dfu` and `run_peripheral_usb_logger` into `run_peripheral_usb`.

I'm unsure though, if both features run together, as I could not test it at the moment.

HaoboGu:
Thanks!

PR #1002, opened 2026-07-25
fix: sdc memory pool size calculation for nrf52840

Schievel1:
In `a5da70f2` we updated nrf-sdc from `3f3a5070` to `abe49d22`. Apparently the new commit of nrf-sdc needs bigger memory pools probably due to `6527dec chore: update to nrfxlib 3.3.0`

This fixes only the use_rust nRF52840 and nRF52832 examples and the calculation for all nRF528xx chips. 
I can not runtime test nRF54 due to the lack of hardware.


HaoboGu:
Thanks!

PR #1003, opened 2026-07-26
feat(morse): add quick_tap_timeout profile option

kot149:
Added `quick_tap_timeout` morse profile option: re-pressing a morse/tap-hold key within the window after its last release fires the tap action immediately and keeps it held, so the OS auto-repeats the tap instead of triggering the hold action.
`MorseProfile` is now packed into a `u64` instead of a `u32` to make room for `quick_tap_timeout`.

closes #751


rmk/src/keyboard/morse.rs
```diff
@@ -131,7 +149,32 @@ impl<'a> Keyboard<'a> {
...
-                        if let Some(action) = final_action {
+                        let quick_tap_window = Self::quick_tap_window(self.keymap, &k.action);
+                        let stash_for_quick_tap = !hold
+                            && pattern.is_all_taps()
+                            && !pattern.is_empty()
+                            && quick_tap_window.is_some()
+                            && matches!(final_action, Some(a) if a != Action::No);
+
+                        if stash_for_quick_tap {
+                            let action = final_action.unwrap();
```
HaoboGu: can the unwraps be removed?

kot149: Yes, like this. However, in the original code, `final_action` is verified to be `Some` in the `if let Some(action) = final_action` branch, so it can be safely unwrapped. Do you prefer the version without unwrap?

```rs
if !hold
    && pattern.is_all_taps()
    && !pattern.is_empty()
    && let Some(action) = final_action
    && action != Action::No
    && let Some(window) = quick_tap_window
{
    let stashed_action = k.action;
    // ...
    let keep_alive = gap.max(window);
    // ...
} else if let Some(action) = final_action {
    // ...
}
```

HaoboGu: It's fine, and `!pattern.is_empty()` can be ignored with `pattern.is_all_taps()`

kot149: fixed: https://github.com/HaoboGu/rmk/compare/8f7a5181438ef4c35608aaaaf70abf0a41fd1c8c..2f97d14e8d5ce73b9a6c42b798f70775991d918b

HaoboGu (approved):
thanks!

Issue #569, opened 2025-10-07
Support non-english characters in keymap setting


HaoboGu:
In german layout, there's characters like `Ä`, `ö`, etc. They are unicode, but the keycode can be mapped to existing keycode(For example, `ö` -> 0x33 = Keycode for `:;`). 

Support those characters in keymap could be great, for example:

```toml
[[layer]]
keys = """
Ä Ü Ö
"""
```

The simplest way is to add `KEYCODE_ALIAS` of those unicode chars.

tib888:
Just do not predefine these aliases in RMK, because Swedish, Danish, German, Hungarian, etc. layouts may all have different position of the "same" letter! And also the symbols which are "covered" by these letters are also remapped to different locations.

In keyboard.toml I use these aliases to support both the US and the Hungarian layout:

```toml

[aliases]
# US layout aliases for symbols:
"~"="WM(Grave, LShift)"
"`"="Grave"
"!"="WM(Kc1, LShift)"
"@"="WM(Kc2, LShift)"
"#"="WM(Kc3, LShift)"
"$"="WM(Kc4, LShift)"
"%"="WM(Kc5, LShift)"
"^"="WM(Kc6, LShift)"
"&"="WM(Kc7, LShift)"
"*"="WM(Kc8, LShift)"
"("="WM(Kc9, LShift)"
")"="WM(Kc0, LShift)"
"-"="Minus"
"_"="WM(Minus, LShift)"
"="="Equal"
"+"="WM(Equal, LShift)"
"["="LeftBracket"
"{"="WM(LeftBracket, LShift)"
"]"="RightBracket"
"}"="WM(RightBracket, LShift)"
"\\"="Backslash"
"|"="WM(Backslash, LShift)"
";"="Semicolon"
":"="WM(Semicolon, LShift)"
"'"="Quote"
"\""="WM(Quote, LShift)"
","="Comma"
"<"="WM(Comma, LShift)"
"."="Dot"
">"="WM(Dot, LShift)"
"/"="Slash"
"?"="WM(Slash, LShift)"

# HUN layout aliases for symbols:
"h~"="WM(Kc1, RAlt)"
"h`"="WM(Kc7, RAlt)"
"h!"="WM(Kc4, LShift)"
"h@"="WM(V, RAlt)"
"h#"="WM(X, RAlt)"
"h$"="WM(Semicolon, RAlt)"
"h%"="WM(Kc5, LShift)"
[...]

PR #1004, opened 2026-07-26
fix(config): support Unicode keymap aliases

tianrking:
Related to #569.

## Summary

- Parse alias identifiers on UTF-8 character boundaries instead of byte offsets, preventing non-ASCII names from producing invalid string slices.
- Use the same Unicode whitespace definition for alias scanning and keymap parsing, so separators such as NBSP and ideographic space remain valid after expansion.
- Preserve the existing alias delimiters, case sensitivity, recursion limit, and `[aliases]` configuration format.
- Add regression coverage for direct Unicode aliases, recursive expansion into an action, undefined Unicode aliases, and Unicode whitespace separators.
- Document the supported behavior and add a changelog entry.

## Validation

- `cargo test --manifest-path rmk-config/Cargo.toml` (92 unit tests and 6 integration tests passed; doc tests passed)
- `cargo clippy --manifest-path rmk-config/Cargo.toml --all-targets --all-features -- -D warnings`
- `cargo +nightly fmt --manifest-path rmk-config/Cargo.toml --check`
- `git diff upstream/main...HEAD --check`

## Scope

This PR fixes Unicode handling in the existing `keyboard.toml` `[aliases]` path. It does not add the Rust-code alias API requested in #569, so that issue should remain open.


HaoboGu:
This cannot close #569

HaoboGu:
@tianrking Hey could you please update the PR and fix the issue/comment? Thanks!

PR #1007, opened 2026-07-30
doc: rename bootymcbootface -> rmk-boot, change links

Schievel1:
Renames bootymcbootface to rmk-boot and changes the links to our organizations own repository. 

Please merge this after merging https://github.com/rmk-rs/rmk-boot/pulls.

### DIFF
--- a/docs/docs/main/docs/configuration/bootloader.mdx
+++ b/docs/docs/main/docs/configuration/bootloader.mdx
@@ -4,7 +4,7 @@
 RMK supports DFU firmware updates via embassy-boot for **RP2040** and **nRF52840**. An embassy-boot based bootloader splits flash into ACTIVE and DFU slots, providing safe updates with automatic rollback on failure.
 This is an optional feature of RMK, the default bootloaders of the devices can still be used as usual without runtime updates via USB DFU.
 
-A pre-built embassy-boot based bootloader called **bootymcbootface** is available for both platforms. The partition formula is identical:
+A pre-built embassy-boot based bootloader called **rmk-boot** is available for both platforms. The partition formula is identical:
 
 ```text
 bootloader+state = 28K (fixed)
@@ -54,12 +54,12 @@ dfu_size     = 528384
 <Tab label={<Rust />}>
 
 ```rust title="main.rs"
-// Flash layout using the bootymcbootface formula:
+// Flash layout using the rmk-boot formula:
 //   state at 0x6000 (4K), active from 0x7000 (size: (flash_size - 28K (= BOOT2 size + embassy-boot + embassy-boot state) - STORAGE_SIZE (= 128K) - page_size (= 4K)) / 2),
 //   dfu follows active (active_size + page_size (= 4K))
 //
 // All offsets (DFU_OFFSET, DFU_SIZE, STORAGE_OFFSET, etc.) are derived
-// automatically from FLASH_SIZE below — change only that constant when using bootymcbootface.
+// automatically from FLASH_SIZE below — change only that constant when using rmk-boot.
 //
 // ⚠  You can define your own FLASH_SIZE and addresses, but then you must build and
 //    flash a custom embassy-boot bootloader with a matching memory.x!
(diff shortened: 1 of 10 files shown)


tompre:
Hi, thanks for all your work on the bootloader/dfu features!

Do we really need to maintain the code for RMK? Are there any RMK specific changes?
Otherwise I would suggest to pull the codeberg repo directly if a new build is required?

HaoboGu:
My 2c:

1. The bootloader is not rmk-specific
2. RMK will be built with the bootloader(as the second stage bootloader) by default in the future, so the OTA/DFU feature is generally available
3. RMK should provide library to make it easy to create rmk-specific CLI/GUI flashing tool: it sends the `jump to bootloader` command, connects to the bootloader via USB/BLE and flashes the firmware using DFU protocol and reboot the keyboard

For the question of where to host the bootloader: if the bootloader is a generic bootloader and requires no core code change in RMK, maintaining it in codeberg/anywhere else is fine. But if the bootloader requires core code changes in RMK, it's better to maintain it in RMK(or rmk organization?). Otherwise, updating either side might break the other.

Schievel1:
> RMK(or rmk organization?) 

I would prefer organization honestly.

> But if the bootloader requires core code changes in RMK

It does, you need the flash manager and the flash partition offsets/ sizes for the bootloader and RMK must match. 
If you don't specifically set them, their default in RMK fits what bootymcbootface does. 
Updating one side _would_ break the other.

HaoboGu:
> It does, you need the flash manager and the flash partition offsets/ sizes for the bootloader and RMK must match.

The dependency relationship can be somehow reverted, for example, bootloader defines the spaces(fixed addresses, or special partition names in linker script), then RMK follows in user application. In this case, no RMK core code is touched.

HaoboGu:
Anyway, I've created https://github.com/rmk-rs, I will transfer the repo soon

Schievel1:
> bootloader defines the spaces(fixed addresses

I am not following. Isn't this how it is now? The bootloader defines an 'algorithm' that tells us the addresses/ sizes. RMK follows along and creates it's ACTIVE, DFU and STATE partition accordingly. The 'contract' that could break is the algorithm how to calculate the sizes. 
You mean you want to couple this more tightly, like the bootloader tells RMK the addresses and RMK reacts automatically? Doesn't this just shift the contract from the algorithm to the way the bootloader tells RMK?

HaoboGu:
> I am not following. Isn't this how it is now? The bootloader defines an 'algorithm' that tells us the addresses/ sizes. RMK follows along and creates it's ACTIVE, DFU and STATE partition accordingly. The 'contract' that could break is the algorithm how to calculate the sizes.

What I mean is that the bootloader does the full update process, and defines addresses in the linker script -- then the bootloader is firmware agnostic and RMK knows nothing except the flash partition defined in bootloader's linker script.

HaoboGu:
To confirm: both `bootymcbootface` and RMK application support USB-DFU, so that if RMK firmware breaks itself(for example, panic after connecting to BLE/USB), `bootymcbootface` can always do the reflash, am I right? Is there a common way to stop in `bootymcbootface` even the RMK application has been marked booted?

HaoboGu:
Thinking about it a little bit more, I think the current overall shape is fine, but there are many details need to be improved. For example:

1. How to share the flash layout between bootloader and RMK app
2. Hardware agnostic bootloader & RMK app
3. Firmware updating procedure, how to prevent the bad firmware being flashed
4. How to co-exist with the existing adafruit nrf52 bootloader system


This PR can be merged as is, but lots of work to do in the future

HaoboGu:
I've migrated the project to the new rmk-rs organization! And I've created https://github.com/rmk-rs/rmk-boot. I think this PR can go there? And if you like `bootymcbootface`, I can change the repo's name.

Schievel1:
> To confirm: both `bootymcbootface` and RMK application support USB-DFU, so that if RMK firmware breaks itself(for example, panic after connecting to BLE/USB), `bootymcbootface` can always do the reflash, am I right? Is there a common way to stop in `bootymcbootface` even the RMK application has been marked booted?
> 
> 
> 
> 
> 
> 

this is only true for the nrf52840 version of bootymcbootface. When plugging the board in or after connecting reset pin to ground and disconnecting it again bootymcbootface has a short time window where you can double tap reset the bring it into dfu mode. Then you can flash using dfu-util.

The RP2040 doesn't need that as it has its own bootloader in ROM that you can always fall back to if something in the application goes terribly wrong.

Schievel1:
Leave this open, I will use it for the doc update

Schievel1:
@HaoboGu I changed this to only change the docs & comments

HaoboGu:
@Schievel1 Thanks a lot! Do you want to keep the name `bootymcbootface`?

Schievel1:
> @Schievel1 Thanks a lot! Do you want to keep the name `bootymcbootface`?

nah, I got a hang for silly stuff but I realize RMK's vibe is more serious. :D

PR #1008, opened 2026-07-30
feat(ble): Enable Bluetooth LE Connection Subrating for split communication

tompre:
As promised a few months back, here is a first draft for enabling connection subrating support on the peripheral connection.

#### Initial Results
Using a subrating factor of **30** (equivalent to a connection interval of **225ms**):
* **Central:** 50–80 µA
* **Peripheral:** ~30 µA

The latency to return to 7.5ms intervals is **225ms + 2×7.5ms** (compared to 5–7 seconds when using standard connection updates). 

This allows us to use sleep mode much more aggressively, as input lag only impacts the very first keystroke.

#### Open Questions / Next Steps 

- [X] **Fast Polling Loop (Resolved)**  
   [#886](https://github.com/HaoboGu/rmk/issues/886) introduced a fast polling loop (5ms) leading to 500–700 µA current draw on Central. *(Tracked/resolved via [#1010](https://github.com/HaoboGu/rmk/issues/1010)).*

- [X] **Central Max Latency**  
   I adjusted Central max latency to match the subrating duration. Unsure what effect this has on OS-level support.
   We will use max_latency = 30 for now and leave timing optimization for a later commit.

- [X] **GATT Buffer Sizing (Resolved)**  
   `SplitMessage` GATT messages were sending the entire buffer rather than just the encoded payload. Truncating to the encoded data allows faster resets of the subrate factor. *(Merged via [#1011](https://github.com/HaoboGu/rmk/pull/1011)).*  
   * Reference: [`rmk/src/split/ble/peripheral.rs#L113-L116`](https://github.com/HaoboGu/rmk/blob/936a2a828fddc195d20e42bd952119cd3afa8172/rmk/src/split/ble/peripheral.rs#L113-L116) and [`rmk/src/split/ble/central.rs#L449-L450`](https://github.com/HaoboGu/rmk/blob/936a2a828fddc195d20e42bd952119cd3afa8172/rmk/src/split/ble/central.rs#L449-L450).

### DIFF
--- a/docs/docs/main/docs/features/low_power.md
+++ b/docs/docs/main/docs/features/low_power.md
@@ -67,6 +67,19 @@ Two related behaviors are always on:
 - When BLE advertising times out without a connection (after 5 minutes), the keyboard sleeps immediately and waits for a key or pointing event before it advertises again.
 - `NrfAdc` takes a `light_sleep` interval as its last argument. When the analog inputs have been idle for more than 1.2 seconds, the ADC polls at that interval instead of `polling_interval`. A joystick configured in `keyboard.toml` uses 350ms.
 
+### Split central sleep using BLE Connection Subrating (nrf52840 only)
+
+To improve central power usage and decrease peripheral latency problems during wake-up you can activate the `subrating` feature in your `Cargo.toml`.
+This enables BLE Connection Subrating which lets the central sleep for longer intervals, while allowing for a quick switch back to the fast connection intervals.
+
+The mean perihperal latency is ~450ms for the first keypress only. The keyboard instantly switches to the active connection settings after that. Depending on your layout and habits this allows to use the sleep feature more agressively. Try to reduce the timeout to get lower battery usage, e.g.:
+```toml
+[rmk]
+split_central_sleep_timeout_seconds = 60
+```
+
+If the central is not connected to a host, the latency is further increased to ~1867ms which reduces the power consumption of the central to ~20µA, barely more then the peripheral.
+
 ## External VCC
 
 Some boards, such as the nice!nano have an external 3.3V regulator that can be used to power the LEDs. If not used, the regulator can be disabled by pulling `P0_13` low to safe power.
--- a/examples/use_config/nrf52840_ble_split/Cargo.toml
+++ b/examples/use_config/nrf52840_ble_split/Cargo.toml
@@ -10,7 +10,7 @@ edition = "2024"
 license = "MIT OR Apache-2.0"
 
 [dependencies]
-rmk = { path = "../../../rmk", default-features = false, features = ["defmt", "storage", "vial", "watchdog", "nrf52840_ble", "split", "async_matrix", "adafruit_bl"] }
+rmk = { path = "../../../rmk", default-features = false, features = ["defmt", "storage", "vial", "watchdog", "nrf52840_ble", "split", "async_matrix", "adafruit_bl", "subrating"] }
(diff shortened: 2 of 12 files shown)


HaoboGu:
> #886 introduced a fast polling loop (5ms) which leads to 500-700uA of current draw in central. We should fix that anyway I think.

Yes, the fix looks fine

> I changed Central max latency to be similarly long as the subrating. I'm unsure what this does with OS support.

I don't see any issue max_latency = 30

> SplitMessage gatt messages always use the full buffer rather than only sending the encoded data, sending only the data might allow for faster resets of the subrate factor, if this was not done on purpose?

Yeah, only encoded buffer should be sent, that needs a fix. `to_slice` returns the encoded buf so the fix is easy.

The result is quite impressive! Thanks a lot for your effort!

tompre:
I updated the description. With the work on #1027 and #1028 ongoing, I think I wait for them to merge before fixing conflicts.

HaoboGu:
@tompre #1028 has landed, could you please rebase this branch onto the latest main? Code structure changed a lot(and I think it's better!)

tompre:
The pull request should be ready:
- I added docs
- I added different settings for connected sleep vs. not-connected sleep. => Central will now have the same current draw as the peripheral when not connected.
- I enabled the feature for the nrf52840_ble_split examples so it gets checked.

HaoboGu (approved):
Fantastic! It looks great.

Could you also add a table of the default connection parameters/subrating parameters somewhere?

tompre:
> Fantastic! It looks great.

Thanks.
> Could you also add a table of the default connection parameters/subrating parameters somewhere?

Of course, where should I put it? In the docs? or in split/ble/central.rs?

I also did some measurements:
``` org
* Current draw on central with subrating
During sleep:
| max l. (HC) | in [ms] | SubF (SC) | in [ms] | avg. [µA] | KP latency [ms] |
|-------------+---------+-----------+---------+-----------+-----------------|
|          30 |    225. |        30 |    225. |        75 |           232.5 |
|          60 |    450. |        30 |    225. |        59 |           232.5 |
|         180 |   1350. |        30 |    225. |        48 |           232.5 |
|         300 |   2250. |        30 |    225. |        48 |           232.5 |
|          30 |    225. |        60 |    450. |        63 |           457.5 |
|          60 |    450. |        60 |    450. |        48 |           457.5 |
|         180 |   1350. |        60 |    450. |        39 |           457.5 |
|         300 |   2250. |        60 |    450. |        39 |           457.5 |
|         300 |   2250. |       120 |    900. |        32 |           907.5 |
#+TBLFM: $2=$-1*7.5::$4=$-1*7.5::$6=($3+1)*7.5
HC ... Host Connection
SC ... Split Connection
KP ... Key Press

In active mode without pressing any key, the average current is: ~600µA

* Current draw on peripheral with subrating
During Sleep with subrate factor: 60 and max_latency: 7, witch leads to a max
sleep time of 3.6s, the average current is: 21µA

In active mode without pressing any key, the average current depends on the max
latency of the split connection:
| max_latency |  [ms] | avg. [µA] |
|-------------+-------+-----------|
|          10 |   75. |        72 |
|          30 |  225. |        38 |
|          60 |  450. |        30 |
|         300 | 2250. |        21 |
#+TBLFM: $2=$-1*7.5
```

HaoboGu:
I prefer to put it in split/ble/central.rs because it's a developer oriented info

HaoboGu:
And the result is quite impressive! Could you also test the avg current when SC=10？

HaoboGu:
@tompre I'm merging it now, thanks very much for bringing this feature to RMK!

PR #1011, opened 2026-07-31
fix(ble): send exact payload size for `SplitMessage`

tompre:
As mentioned in #1008, split messages are not sent efficiently at the moment. Instead the whole buffer is sent even if only part of it was used for the encoding.

This pull request changes the Characteristics type to variable length and enables efficient bandwidth usage.

It introduces `GattSplitMessage`, a wrapper around a `[u8; SplitMessage::POSTCARD_MAX_SIZE]` buffer.
It uses `SplitMessage::POSTCARD_MAX_SIZE` as the length, as `SPLIT_MESSAGE_MAX_SIZE` includes bytes needed for the COBS encoding for the serial link, which is not necessary for BLE.

### DIFF
--- a/rmk/src/split/ble/central.rs
+++ b/rmk/src/split/ble/central.rs
@@ -9,6 +9,7 @@ use embassy_time::{Duration, Timer, with_timeout};
 use heapless::VecView;
 use trouble_host::prelude::*;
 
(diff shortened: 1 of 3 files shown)


HaoboGu (approved):
Looks good, thanks very much!

PR #1017, opened 2026-08-01
fix(dfu_split): wait for passthrough signal instead of polling

### LINKED ISSUE #1010: Regression on central power usage
As already mentioned in #1008 there is a fast polling loop (5ms) that prevents the central from long sleep intervals and low power consumption:
https://github.com/HaoboGu/rmk/blob/936a2a828fddc195d20e42bd952119cd3afa8172/rmk/src/split/driver.rs#L158

The change was introduced in #886 to poll for new dfu commands to pass through:
https://github.com/HaoboGu/rmk/blob/936a2a828fddc195d20e42bd952119cd3afa8172/rmk/src/split/driver.rs#L130-L134

My first idea was to increase the polling time but that would decrease data throughput to a unusable level.

I think we might need to change the design of the passthrough command queue:
https://github.com/HaoboGu/rmk/blob/936a2a828fddc195d20e42bd952119cd3afa8172/rmk/src/dfu/split/central.rs#L210-L218  

@Schievel1: Would it be possible to use the event macro to create a mpsc channel and wait for new commans?

Schievel1:
Quick and easy fix for #1010 by using embassy_sync's signal instead of polling every 5ms. 

Maybe I will revisit this in the future and do some refinement, but this should do the trick for now. 

fix: #1010

### DIFF
--- a/rmk/src/dfu/mod.rs
+++ b/rmk/src/dfu/mod.rs
@@ -58,7 +58,8 @@ mod split;
 use self::split::PassthroughDfuHandler;
 #[cfg(feature = "dfu_split")]
 pub(crate) use self::split::{
-    PASSTHROUGH_TARGET, PassthroughCommand, passthrough_done_if_empty, passthrough_pending, passthrough_take_command,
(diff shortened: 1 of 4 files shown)


HaoboGu (approved):
thanks

PR #1018, opened 2026-08-02
feat: use rmk-boot.x linker script for DFU partition layout

Schievel1:
This PR should simplify the process of setting the right flash layout when using rmk-boot. 
The calculation formula is still the same (flash size - what the bootloader takes up - storage = rest. ACTIVE = rest/2 - 1 page, DFU = rest/2 + 1 page) and rmk-boot's `build.rs` creates a `memory.x` accordingly. Additionally it should create a `rmk-memory.x` with the matching flash layout of the (RMK-)application + a few linker symbols that the application can use to initialize the flash and its partitions. 
e.g. a `rmk-memory.x` looks like this: 
```
MEMORY {
  FLASH : ORIGIN = 0x10007000, LENGTH = 966656   /* absolute XIP, where your firmware lives */
  RAM   : ORIGIN = 0x20000000, LENGTH = 256K
}

__rmk_boot_state_offset   = 0x6000;   /* flash-relative, consumed at runtime */
__rmk_boot_state_size     = 0x1000;
__rmk_boot_dfu_offset     = 0xF3000;
__rmk_boot_dfu_size       = 0xED000;
__rmk_boot_storage_offset = 0x1E0000;
```

rmk-boot not only creates that `rmk-memory.x` at build, it also provides matching `rmk-memory.x`-files for its pre-built variants. (rmk-rp2040-16mb-memory.x for example)

The users can copy those files into their project directory and rename them to `memory.x`.
The users code can then use `rmk::dfu::init_flash_from_linkerscript()` instead of `rmk::dfu::init_flash()` to partition the flash without giving any offsets and sizes. For use_config this happens automatically. 

User who build their own bootloader with embassy-boots swap feature can change the symbols in memory.x accordingly. 

Related PRs:
https://github.com/rmk-rs/rmk-boot/pull/2


HaoboGu:
The symbol existence is checked in rmk when bootloader feature like `rmk-boot` is enabled. This means that if `rmk-boot` is defined, the user should either define their own `__rmk_boot*` symbols in their linkerscript or use `rmk-boot.x`.

And I think it should not be overwritten by `keyboard.toml`. Let user define their layout in their linker script is better imo, because I think 99% users won't do it. That makes the configuration cleaner.

The last point from me is the storage size, default 128KB is a little bit more. For example, nRF52833 has only 512KB flash(I just realized that it's impossible to use rmk-boot on it). I want it to be tunable, or default to 32KB.

Schievel1:
> The symbol existence is checked in rmk when bootloader feature like `rmk-boot` is enabled. This means that if `rmk-boot` is defined, the user should either define their own `__rmk_boot*` symbols in their linkerscript or use `rmk-boot.x`.
> 
> And I think it should not be overwritten by `keyboard.toml`. Let user define their layout in their linker script is better imo, because I think 99% users won't do it. That makes the configuration cleaner.
> 
> The last point from me is the storage size, default 128KB is a little bit more. For example, nRF52833 has only 512KB flash(I just realized that it's impossible to use rmk-boot on it). I want it to be tunable, or default to 32KB.

for now I prefer to bind everything bootloader related to the `dfu` feature, since this is the only benefit we get at the moment with `rmk-boot`. There is not other reason to use it. As we spoke on discord, I will try and make bootloader versions that don't have DFU partitions and the failsafe swap, then we can split a new `rmk-boot` feature off of `dfu`. 

The offsets and sizes are now bound to symbols defined in `rmk-boot.x`. This file is created by rmk-boot at built. The pre-built binaries come with pre-created `rmx-boot.x`, that the user copies in their project dir instead of a `memory.x`.
`keyboard.toml` can't change that layout anymore, instead when users want to roll their own bootloader, their have to define the offsets in their linker script and link with it. 

Storage size is not easily changeable in RMK when using embassy-boot, because when changing the storage size (offset) this also changes the size of the ACTIVE and DFU partitions. So you end up with new sizes and offsets for them as well. I tuned down the default to 32K and in rmk-boot I added the possibility to easily change the storage size. Then users just need to do change that, build rmk-boot, flash it and copy the `rmk-boot.x` into their project to change the storage size.

examples/use_rust/rp2040_embassy_boot_split/build.rs
```diff
@@ -11,12 +11,18 @@ fn main() {
...
 
     let out = &PathBuf::from(env::var_os("OUT_DIR").unwrap());
-    File::create(out.join("memory.x"))
-        .unwrap()
-        .write_all(include_bytes!("memory.x"))
-        .unwrap();
-    println!("cargo:rustc-link-search={}", out.display());
-    println!("cargo:rerun-if-changed=memory.x");
+
+    if Path::new("rmk-boot.x").exists() {
```
HaoboGu: I prefer to have `rmk-boot.x` in rmk-boot project only and keep `memory.x`. When users want to use rmk-boot, just add 1 line like:

```rust
println!("cargo:rustc-link-arg=-Trmk-boot.x");
```

This is the pattern of many Rust embedded projects, like cortex-m-rt, defmt. I don't know if it's possible for rmk-boot. I prefer this way if it's possible.

Schievel1: That would require manual changes in memory.x because when using RMK-boot the flash starts at 0x7000 for the app. 

Anyway, maybe I was thinking way too complicated here. How about rmk-boot gives memory.x files for every option. Same thing as now, no change in build.rs needed and intuitive. that memory.x would then be the same content like what rmk-boot.x is now, it already contains the right flash layout and the few symbols defined.

HaoboGu: Yeah, maybe this is the simplest solution. If we do this, can users add their own memory.x if they want?

Schievel1: I was thinking, RMK-boot generates a rmk-boot-memory.x (for example) for the RMK application during build, user copies it into their project directory and renames it to memory.x.

Users can edit it to their liking then. 
 They can still write their own memory.x manually, it's just a helper.

// maybe I should mention: I already planned the noswap feature where there is no DFU partition and a bigger ACTIVE partition for the firmware instead. This requires a change in the memory.x for the user. I wanted to add a feature in rmk-boot for uf2 flashes which could require more space than the current 24kB, what would require a change in the memory.x again. 
I don't want to require the user to find out what settings must be done in their flash layout in memory.x given the features of rmk-boot they use. 
That's why I came up with that copy thing in the first place.

Schievel1: I updated the PR to use `memory.x` instead of a new `rmk-boot.x`. Please have a look if that fits better.

rmk/src/dfu/rp.rs
```diff
@@ -74,3 +74,33 @@ pub fn mark_booted() {
...
+/// Initialize flash using partition offsets from the linker (rmk-boot.x).
+///
+/// Requires `rmk-boot.x` to be linked into the firmware binary
+/// (e.g. `-Trmk-boot.x` in `.cargo/config.toml`).
+///
+/// # Safety
+///
+/// Reads linker-defined absolute symbols. The symbols must be present in the
+/// linked binary or the firmware will not link.
+pub fn init_flash_from_linkerscript(flash_peri: Peri<'static, FLASH>) -> PartitionType {
```
HaoboGu, line 90: This function is duplicate, this should be a generic should over `(Async)NorFlash` in `embedded-storage` trait.

HaoboGu (approved):
Yeah I agree with you! The current way looks great and the only comment from me is same as before -- we should build the DFU based on `embedded-storage` trait, it isn't a blocker of this PR, but should be done in the near future

Schievel1:
Will be done when I finally get my external flash to test things out. 

I prefer smaller PRs that are about one thing only. It's easier to review and test this way.

HaoboGu:
Should I merge this PR as is and leave the `impl NorFlash` in #1030?

Schievel1:
Yes that's I wanted to do. 
I'm still not sure about the Norflash. 

But I wanted to have another go in this PR.
It seems like someone at embassy-boot had the same thought as you did and they already have the functionality to read the partition offsets inside the bootloader. If that's the case it would be nice to use the same symbol names in RMK. 

Also let me do the small fixes in the corresponding PR in rmk-boot first. They should be merged at the same time.

Schievel1:
https://github.com/embassy-rs/embassy/blob/89f0ce53bb64d827a00fec5222df4e831b6d541b/embassy-boot/src/firmware_updater/blocking.rs#L59

Using those symbols now. I think it's better to have the same symbols instead of doing our own thing.

I will test this once more with rp and nrf this weekend and then mark it ready to merge.

Schievel1:
Tested with RP2040 and nRF52840. I think this is good now.

HaoboGu:
great, thanks!

PR #1027, opened 2026-08-04
feat(rynk): dongle groundwork — advertising constants, header peek, pairing window

HaoboGu:


HaoboGu:
Superseded by #1028 — closing so the stack collapses to a single PR.

Nothing is lost: all 8 files this branch touches are byte-identical between its head and `feat/dongle-core`. The company-id half additionally landed in main via #1039.

The split was along the wrong seam. Everything here was inert on its own — `dongle_pairing_window_secs` gated on `CARGO_FEATURE_DONGLE` and the subscriber block on `features = ["dongle"]`, but the `dongle` feature only arrives in #1028, so neither ever fired; `DONGLE_SEEKING_ADV_KIND`, the `Cmd` 0x09xx reservation and `RynkHeader::peek` had no callers. And it was not only inert: it moved the split peripheral onto `RMK_ADV_COMPANY_ID` and changed the value while the matching central-side change stayed in #1028, so at this tip the peripheral advertises `53 52` while the central still matches `18 e1` and split BLE does not pair at all.

The contents now sit in the commits that use them, so no commit lands inert and none leaves split broken.

PR #1030, opened 2026-08-04
feat(dfu_ext): Add ext dfu flash

Schievel1:
Add possibility to use external flash as DFU partition. 
Also a partial rewrite of DFU architecture. I split the struct dfustate into two separate structs. The first struct, `UsbProxyDfuHandler` only forwards DFU commands into a channel. This struct is the USB handler and has static lifetime. The second struct, `RmkDfuInterface` holds the partitions and doesn't need static lifetime. It blocks on messages from that channel and does the writing into partitions, mark updated and such. The result is that partitions do not need to be in static cells.
Only downside it that someone has to run that handler with the partitions now but I think that's fair enough.

Tests I need to carry out on hardware:
| Done | Platform | Test | Bootloader build (cargo features) |
|---|---|---|---|
| [ ] | nRF52833 | DFU flash & swap | `nrf52833` |
| [ ] | nRF52833 | DFU flash & rollback | `nrf52833` |
| [ ] | nRF52833 | Bootloader direct DFU flash & swap | `nrf52833` |
| [ ] | nRF52833 | Bootloader direct DFU flash & rollback | `nrf52833` |
| [ ] | nRF52833 | noswap bootloader direct DFU flash | `nrf52833,noswap` |
| [ ] | nRF52833 | dfu_ext flash & swap | `nrf52833,dfu_ext` |
| [ ] | nRF52833 | dfu_ext flash & rollback | `nrf52833,dfu_ext` |
| [ ] | nRF52833 | dfu_ext bootloader direct DFU flash & swap | `nrf52833,dfu_ext` |
| [ ] | nRF52833 | dfu_ext bootloader direct DFU flash & rollback | `nrf52833,dfu_ext` |
| [x] | nRF52840 | DFU flash & swap | `nrf52840` |
| [x] | nRF52840 | DFU flash & rollback | `nrf52840` |
| [x] | nRF52840 | Bootloader direct DFU flash & swap | `nrf52840` |
| [x] | nRF52840 | Bootloader direct DFU flash & rollback | `nrf52840` |
| [x] | nRF52840 | noswap bootloader direct DFU flash | `nrf52840,noswap` |
| [x] | nRF52840 | dfu_ext flash & swap | `nrf52840,dfu_ext` |
| [x] | nRF52840 | dfu_ext flash & rollback | `nrf52840,dfu_ext` |
| [x] | nRF52840 | dfu_ext bootloader direct DFU flash & swap | `nrf52840,dfu_ext` |
| [x] | nRF52840 | dfu_ext bootloader direct DFU flash & rollback | `nrf52840,dfu_ext` |
| [x] | RP2040-2MB | DFU flash & swap | `rp2040,rp2040-2mb` |
| [x] | RP2040-2MB | DFU flash & rollback | `rp2040,rp2040-2mb` |
| [x] | RP2040-2MB | dfu_split flash & swap | `rp2040,rp2040-2mb` |
| [x] | RP2040-2MB | dfu_split flash & rollback | `rp2040,rp2040-2mb` |
| [x] | RP2040-2MB | dfu_split, dfu_ext flash & swap | `rp2040,rp2040-2mb,dfu_ext` |
| [x] | RP2040-2MB | dfu_split, dfu_ext flash & rollback | `rp2040,rp2040-2mb,dfu_ext` |
| [x] | RP2040-2MB | dfu_ext flash & swap | `rp2040,rp2040-2mb,dfu_ext` |
| [x] | RP2040-2MB | dfu_ext flash & rollback | `rp2040,rp2040-2mb,dfu_ext` |


rmk/src/dfu/mod.rs
```diff
@@ -40,13 +249,128 @@ pub use self::rp::{DFU_WRITE_SIZE, get_manager, init_flash, mark_booted};
...
+/// globally, using partition offsets from the linker (rmk-boot.x).
+///
+/// `flash` is the raw flash peripheral — `p.FLASH` on RP2040, `p.NVMC` on
+/// nRF. The flash is parked in a `'static` mutex, the [`DfuFlashManager`]
+/// is stored globally and the storage partition is returned.
+///
+/// Requires `rmk-boot.x` to be linked into the firmware binary
+/// (e.g. `-Trmk-boot.x` in `.cargo/config.toml`).
+#[cfg(any(feature = "dfu_rp", feature = "dfu_nrf"))]
+pub fn init_flash_from_linkerscript(flash: FlashType) -> PartitionType {
```
Schievel1: honestly, best I can do is hiding the type behind a FlashType and importing the concrete type depending on feature set. 

The problem is, that the flash must be stored in a static for the usb registration, but static can't be generics. 
So in order to have this as `impl NorFlash` the user would have to make this static, which overcomplicates the user facing API imo. 

Maybe you have a better idea.

rmk/src/dfu/mod.rs
```diff
@@ -40,13 +249,128 @@ pub use self::rp::{DFU_WRITE_SIZE, get_manager, init_flash, mark_booted};
...
+/// build the DFU manager with the external (SPI) flash as the DFU download
+/// partition.
+///
+/// Combines [`init_flash_from_linkerscript`] with external-flash DFU setup:
+/// `dfu_flash` (wrapped in a `'static` mutex) becomes the manager's DFU
+/// partition — `mgr.dfu_partition()` returns it — while the state and storage
+/// partitions stay on the internal flash. Returns the internal storage
+/// partition.
+#[cfg(any(feature = "dfu_rp", feature = "dfu_nrf"))]
+pub fn init_flash_from_linkerscript_with_external_dfu<M: RawMutex + Sync, DFU: NorFlash + Send>(
```
Schievel1: Maybe handing in the partition instead of the whole flash is better here. Flash is cheap and abundant, most users get an 8mb flash while firmware is only ~400kb. It's a waste to use the whole 8mb flash for only that. 
Handing in a (smaller) partition instead would give the user the possibility to create another one in the unused area. For images for displays or animations idk

rmk/src/dfu/mod.rs
HaoboGu: The current design can be simplified a lot by using `NorFlash`, and 'static is not needed actually.

We can just have:

```rust
pub struct RmkDfuInterface<'d, DFU: NorFlash, STATE: NorFlash> {
    central: DfuState<RmkDfuHandler<FirmwareHandler<'d, DFU, STATE, ResetImmediate, BLOCK_SIZE_DFU>>>,
    #[cfg(feature = "dfu_split")]
    passthrough: [Option<DfuState<RmkDfuHandler<PassthroughDfuHandler>>>; MAX_PASSTHROUGH_ALTS],
    #[cfg(feature = "dfu_split")]
    num_passthrough: usize,
    current_alt: u8,
}
impl<'d, DFU: NorFlash, STATE: NorFlash> RmkDfuInterface<'d, DFU, STATE> {
    pub fn new(
        dfu: DFU,
        state: STATE,
        aligned: &'d mut AlignedBuffer<DFU_ALIGN>,
        #[cfg(feature = "dfu_split")] num_peripherals: usize,
    ) -> Self {
        const { assert!(STATE::WRITE_SIZE <= DFU_ALIGN) };
        let updater = BlockingFirmwareUpdater::new(
            FirmwareUpdaterConfig { dfu, state },
            &mut aligned.0[..STATE::WRITE_SIZE],
        );
        // ...
    }
}

pub fn mark_booted(state: impl NorFlash, aligned: &mut AlignedBuffer<DFU_ALIGN>);
```

And in the user code it becomes:

```rust
let flash = Mutex::<CriticalSectionRawMutex, _>::new(RefCell::new(Nvmc::new(p.NVMC)));
let mut aligned = AlignedBuffer([0; 32]);

rmk::dfu::mark_booted(BlockingPartition::new(&flash, STATE_OFFSET, STATE_SIZE), &mut aligned);

let mut dfu = RmkDfuInterface::new(
    BlockingPartition::new(&flash, DFU_OFFSET, DFU_SIZE),
    BlockingPartition::new(&flash, STATE_OFFSET, STATE_SIZE),
    &mut aligned,
    SPLIT_PERIPHERALS_NUM,
);

let storage = async_flash_wrapper(BlockingPartition::new(&flash, STORAGE_OFFSET, STORAGE_SIZE));
```

Schievel1: The UsbTransport requires things to have static lifetime. 
src/usb/mod.rs:
```
pub struct UsbTransport<'a, D: Driver<'static>> {
```
and 
```
impl<'a, D: Driver<'static>> UsbTransport<'a, D> {
    pub fn new(driver: D, device_config: DeviceConfig<'static>) -> Self {
```

Only way to do static I think is when the type is known, and it's only known in user code. So we have to shift what we do here into user code. (the `let flash_mutex: &'static MutexType = FLASH_CELL.init(Mutex::new(RefCell::new(flash)));`)

Like this: 
```
// types only added to make things a bit easier to overlook
type InternalFlashMutex =
    Mutex<CriticalSectionRawMutex, RefCell<Flash<'static, FLASH, Blocking, { rmk::dfu::FLASH_SIZE }>>>;
type InternalFlashPartition =
    BlockingPartition<'static, CriticalSectionRawMutex, Flash<'static, FLASH, Blocking, { rmk::dfu::FLASH_SIZE }>>;


    static FLASH_MUTEX: StaticCell<InternalFlashMutex> = StaticCell::new();
    let flash_mutex = FLASH_MUTEX.init(InternalFlashMutex::new(RefCell::new(embassy_rp::flash::Flash::<
        _,
        embassy_rp::flash::Blocking,
        { rmk::dfu::FLASH_SIZE },
    >::new_blocking(p.FLASH))));

    let state_partition =
        InternalFlashPartition::new(flash_mutex, dfu_flash_layout.state_offset, dfu_flash_layout.state_size);
    let flash = async_flash_wrapper(InternalFlashPartition::new(
        flash_mutex,
        dfu_flash_layout.storage_offset,
        dfu_flash_layout.storage_size,
    ));


```

I will try to make UsbTransport & co a shorter lifetime, but that is a bigger change.

rmk/src/dfu/mod.rs
Schievel1: This works anyway because we accept trait norflash

rmk/src/dfu/mod.rs
Schievel1: > I will try to make UsbTransport & co a shorter lifetime, but that is a bigger change.

I tinkered around with this, but I think it only gets worse from here. 
So I think we should be going what you proposed, instead of hiding the staticcell complexity behind a function that needs concrete types and therefore must be implemented separately for each. 
``` Rust
type ExternalFlash = W25qNorFlash<spi::Spi<'static, peripherals::SPI0, spi::Blocking>, Output<'static>>;
type ExternalPartition = BlockingPartition<'static, CriticalSectionRawMutex, ExternalFlash>;
type InternalFlashMutex =
    Mutex<CriticalSectionRawMutex, RefCell<Flash<'static, FLASH, Blocking, { rmk::dfu::FLASH_SIZE }>>>;
type InternalFlashPartition =
    BlockingPartition<'static, CriticalSectionRawMutex, Flash<'static, FLASH, Blocking, { rmk::dfu::FLASH_SIZE }>>;


#[embassy_executor::main]
async fn main(_spawner: Spawner) {
    static DFU_MUTEX: StaticCell<Mutex<CriticalSectionRawMutex, RefCell<ExternalFlash>>> = StaticCell::new();
    let dfu_mutex = DFU_MUTEX.init(Mutex::new(RefCell::new(ext_flash)));
    let dfu_partition = ExternalPartition::new(dfu_mutex, 0, dfu_mutex.lock(|c| c.borrow().capacity() as u32));

    let dfu_flash_layout = dfu_flash_layout();
    static FLASH_MUTEX: StaticCell<InternalFlashMutex> = StaticCell::new();
    let flash_mutex = FLASH_MUTEX.init(InternalFlashMutex::new(RefCell::new(embassy_rp::flash::Flash::<
        _,
        embassy_rp::flash::Blocking,
        { rmk::dfu::FLASH_SIZE },
    >::new_blocking(p.FLASH))));
    let mut state_partition =
        InternalFlashPartition::new(flash_mutex, dfu_flash_layout.state_offset, dfu_flash_layout.state_size);
    let flash = async_flash_wrapper(InternalFlashPartition::new(
        flash_mutex,
        dfu_flash_layout.storage_offset,
        dfu_flash_layout.storage_size,
    ));


    rmk::dfu::mark_booted(&mut state_partition);


    static DFU_IFACE: StaticCell<rmk::dfu::RmkDfuInterface<ExternalPartition, InternalFlashPartition>> =
        StaticCell::new();
    let dfu_iface = DFU_IFACE.init(rmk::dfu::RmkDfuInterface::new(dfu_partition, state_partition));
    let mut usb_transport =
        UsbTransport::new_with_dfu(driver, rmk_config.device_config, dfu_iface).with_host_service(&host_service);

// ...
}

```

The handler is then generic over norflash:
```Rust

pub struct RmkDfuInterface<'d, DFU: NorFlash, STATE: NorFlash> {
    central: DfuState<RmkDfuHandler<FirmwareHandler<'d, DFU, STATE, ResetImmediate, BLOCK_SIZE_DFU>>>,
    #[cfg(feature = "dfu_split")]
    passthrough: [Option<DfuState<RmkDfuHandler<PassthroughDfuHandler>>>; MAX_PASSTHROUGH_ALTS],
    #[cfg(feature = "dfu_split")]
    num_passthrough: usize,
    current_alt: u8,
}

impl<'d, DFU: NorFlash, STATE: NorFlash> RmkDfuInterface<'d, DFU, STATE> {
    /// Build the DFU interface from a DFU download partition and a boot state
    /// partition.
    pub fn new(dfu: DFU, state: STATE, #[cfg(feature = "dfu_split")] num_peripherals: usize) -> Self {
...
```

HaoboGu: What I don't understand is, why `UsbTransport` needs dfu_iface? The transport should do the transport thing only, forward the data to the DFU part. So it should not own dfu info.

Schievel1: Is all comes down to embassy-usb-dfu and embassy-boot do not split partition ownership and usb handling apart. 

Ultimately the partitions are owned by the `BlockingFirmwareUpdater` in embassy-boot.
The `BlockingFirmwareUpdater` is owned by the `FirmwareHandler` in embassy-usb-dfu, which in turn is owned (& wrapped) by `RmkDfuHandler` which is owned by `DfuState` which does the usb stuff. 

Maybe I can split the usb part and the firmware handler apart. DfuState gets a handler that only does forward messages between the usb and the actual firmware handler. The actual one then could live on the stack in main and hold the partitions and do the DFU stuff with them if it's receiving messages.

Schievel1:
Alright taking inspiration from the peripheraldfuhandler and its passthroughhandler I split the dfustate into two separate structs. The first only forwards dfu commands into a channel it is the usb handler and has static lifetime, the other holds the partitions and doesn't need static lifetime. It blocks on messages from that channel and does the writing into partitions, mark updated and such. 

Only downside it that some has to run that handler with the partitions now but I think that's fair enough.

rmk/src/dfu/mod.rs
HaoboGu: The solution is to create separated `Partition`s for DFU & Storage? The `Partition` already implements `NorFlash` if I remember it correctly

For example:

```rust
pub type DfuFlash<F> = embassy_sync::mutex::Mutex<CriticalSectionRawMutex, F>;
pub type DfuPartition<'a, F> = embassy_embedded_hal::flash::partition::Partition<'a, CriticalSectionRawMutex, F>;

pub fn partitions_from_linkerscript<F: NorFlash>(flash: &DfuFlash<F>) -> (DfuPartition<'_, F>, DfuSession<'_, F>) {
    use embassy_embedded_hal::flash::partition::Partition;

    unsafe extern "C" {
        static __bootloader_state_start: u8;
        static __bootloader_state_end: u8;
        static __bootloader_dfu_start: u8;
        static __bootloader_dfu_end: u8;
        static __bootloader_storage_start: u8;
        static __bootloader_storage_end: u8;
    }
    let part = |start: &u8, end: &u8| {
        let addr = |s: &u8| core::ptr::from_ref(s) as usize as u32;
        let (start, end) = (addr(start), addr(end));
        Partition::new(flash, start, end - start)
    };
    // SAFETY: linker-defined absolute symbols — reading their addresses is safe.
    unsafe {
        (
            part(&__bootloader_storage_start, &__bootloader_storage_end),
            DfuSession::new(
                part(&__bootloader_dfu_start, &__bootloader_dfu_end),
                part(&__bootloader_state_start, &__bootloader_state_end),
            ),
        )
    }
}

// user code
let dfu_flash = rmk::dfu::DfuFlash::new(rmk::storage::async_flash_wrapper(Nvmc::new(p.NVMC)));
let (flash, mut dfu_session) = rmk::dfu::partitions_from_linkerscript(&dfu_flash);


```

Schievel1: DFU and Storage partitions have always been separate. 
The solution is to separate USB handling for DFU and writing to the partitions.

Before: one handler does everything, usb requires that handler to be static, therefore everything that handler holds must be static. The handler holds the partition, so they had to be static.

Now: one handler does the USB handling for DFU, but all it does is piping the DFU commands into a channel. Another handler reads from the channel and does the flash operations for DFU and therefore holds the partitions.
Via the channel in the middle the lifetimes of both structs and therefore of the usb handler and the partitions are separated. 

Have a look at one of the examples, they are pretty close now to what you proposed in the first place. 
I think having a function that creates all the partitions directly from the linkerscript is also possible. Apart from the DFU partition in external flash, because that needs its own flash mutex.

Schievel1:
I just found out I am lacking nrf52833 hardware to test this out. Because the BBC Microbit I bought is nrf52833, but it does not have a USB connector. The one USB socket is has is for the on board debugging chip apparently. 
So either someone with nrf52833 hardware is willing to test this out, at least the noswap path, or we go with code review only.
The nrf52833 path is very similar to the  nrf52840, in fact the only difference is the flash partitioning in the provided memory.x from rmk-boot: the 33 has 512kB flash and the 40 has 1MB.

Schievel1:
I think this is ready now. The one failing CI is because of the version mismatch of crate and what the test expects because you just released. 

The noswap and nrf52833 path are just changes in the docs for RMK, that's why I included them in here. All users need to do is use the fitting memory.x

If someone happens to know a fitting w25q driver I would prefer  the user to include it not to have it in this repo. I haven't found one.

docs/docs/main/docs/configuration/bootloader.mdx
```diff
@@ -173,16 +194,206 @@ See the [DFU lock section](../user_guide/flash_firmware/use_embassy_boot.mdx#unl
...
+sck = "PIN_18"
+mosi = "PIN_19"
+miso = "PIN_16"
+cs = "PIN_17"
+```
+
+The function must have the signature:
+
+```rust
+fn init(spi: impl SpiBus, cs: impl OutputPin, flash_size: u32) -> impl NorFlash
```
HaoboGu, line 344: I'm not sure, is this enough for most types of external flashes? I'm thinking about the similar thing for a long time -- to use the external/3rd-party drivers in `keyboard.toml`.

Maybe we should define a universal API of using 3rd-parth drivers in `keyboard.toml`?

rmk-macro/Cargo.toml
```diff
@@ -30,10 +30,12 @@ serde_json = { version = "1", optional = true }
...
 ## Expose the `run_tests!` simulator-scenario macro (rmk test suite only)
 _simulator = ["dep:toml", "dep:serde", "dep:serde_json"]
-dfu_rp = []
-## Enable nRF DFU firmware update code generation
-dfu_nrf = []
+## Base DFU code generation, pulled by `dfu_rp`/`dfu_nrf`
+dfu = []
+dfu_rp = ["dfu"]
+dfu_nrf = ["dfu"]
 dfu_split = []
```
HaoboGu, line 37: Why `dfu_split` and `dfu_ext` don't require `dfu`?

And, is `dfu` an internal feature? If so, use `_dfu`?

HaoboGu (commented):
I did a roughly review on the files except code in the main crate, the design looks great, I have a few comments.

Another point I don't understand is, why using `RMK_DFU_BIN` in examples? Does cargo workspace work better in this case?

I'll review the remaining files soon, thanks for the work! I feel that it improves a looooot than the first version

Schievel1:
I haven't looked a cargo workspaces. All I wanted was a quick way to tell the build.rs to use a different memory.x for each binary.
Also if people use cargo make bin it should work automatically.

This is also only for the very edge case if you want two different flash layouts for central and peripheral. E.g. central has an external flash for DFU partition and peripheral uses internal flash. I thought that's useful because peripherals tend to have smaller space requirements. But each side needs a different bootloader for that as well. 
If you have the same layout on both sides there's no environment variables needed and you would just build with a normal build.rs like it's in the other examples.

HaoboGu:
I just feel the `RMK_DFU_BIN` is a little bit verbose, maybe something built-in in cargo like `cargo::rustc-link-arg-bin=BIN_NAME=X` is better?



I'm imaging something like:
```rust
// build.rs

println!(
    "cargo::rustc-link-arg-bin=central=-Tmemory-central.x"
);
```

when compiling using `cargo build --release --bin central`, the `memory-central.x` will be automatically used.

I didn't test whether it works, just checked the doc here: https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-link-arg-bin

HaoboGu:
2 more comments:

1. Now the update is blocking, I prefer an async way: use `FirmwareUpdater` and remove `FirmwareHandler` 
2. For the SPI, use `embedded_hal_async`(and `embassy_embedded_hal` if needed)

rmk-macro/Cargo.toml
Schievel1: Yes the idea is that dfu is internal. I will rename that. 

dfu_split requires dfu_rp OR dfu_nrf, as does dfu_ext and dfu_lock. dfu_rp and dfu_nrf pull in dfu. So the idea here is that users active dfu_rp/nrf for their chip, then activate dfu_split/ext/lock on top of that if they need it. dfu_split/dfu_ext alone pulling in dfu (but not dfu_rp/nrf) is not valid. 
But you can not mirror this relation in rusts cargo.toml. You can only pull in dependencies. 
So I did checks in rmk/src/lib.rs instead.

docs/docs/main/docs/configuration/bootloader.mdx
Schievel1: Not quite sure how this should work but it's definitely beyond the scope of this PR. Because this one is 14000 lines already lol.
We can talk about this in discord though, I'm interested in the idea how you would design that. 

I was thinking about making this more generic and accept different things than SPI. But then again, external norflash chips not having  SPI are so obscure if users really have one of those they can use the rust configuration.

Schievel1:
Right we should either use Firmwareupdater or blockingfirmwareupdater. Now that we decoupled usb and dfu there's no reason to use embassy-usb-dfu for that anymore. 

But regarding the async... is it really worth the hassle? In case of a firmwareupdate (and a following reset) do we need to not block and keep being async? 
Because when done completely we would have to make the internal flash async as well. (With dma channels and whatnot. For nrf I think async flash not provided.

Or should we rather only make the external flash async and wrap the internal in blockingasync similar to what you do with storage?

HaoboGu:
For internal flash we can wrap it using `blockingasync`, which is actually blocking method. A single async firmware update interface is preferred.

examples/use_config/rp2040_dfu_split_dfu_ext/build.rs
```diff
@@ -0,0 +1,70 @@
...
+
+    // An empty memory.x so link.x's `INCLUDE memory.x` is a no-op;
+    // the real MEMORY section comes from -Tmemory-<bin>.x below.
+    fs::write(out.join("memory.x"), "/* empty — provided per binary via -T flag */\n").unwrap();
+
+    println!("cargo:rustc-link-search={}", out.display());
+
+    // Per-binary linker scripts: memory layout + standard link.x + defmt.x
+    println!("cargo:rustc-link-arg-bin=central=-Tmemory-central.x");
+    println!("cargo:rustc-link-arg-bin=central=-Tlink.x");
```
HaoboGu: `link.x`, `defmt.x` and `-nmagic` can be shared, no `central/peripheral` is needed I think

```rust
    println!("cargo:rustc-link-arg=--nmagic");
    println!("cargo:rustc-link-arg=-Tlink.x");
    println!("cargo:rustc-link-arg=-Tdefmt.x");
```

rmk-macro/src/codegen/feature.rs
```diff
@@ -51,3 +51,12 @@ pub(crate) fn is_feature_enabled(feature_list: &Option<Vec<String>>, feature: &s
     }
     false
 }
+
+/// Whether the flash-init expansion defines the `state_partition` and
+/// `dfu_partition` locals. Sound because rmk's lib guards enforce
+/// `dfu ⇒ dfu_rp | dfu_nrf`, so `dfu` on means a chip backend — and with it
+/// the partitions — is compiled in. All code referencing those partitions
+/// gates on this instead of re-deriving `dfu_rp || dfu_nrf`.
+pub(crate) fn defines_dfu_partitions() -> bool {
```
HaoboGu: Is this kind of thin wrapper needed?

rmk/src/dfu/mod.rs
```diff
@@ -67,73 +173,69 @@ pub use self::split::{
...
-    }
-
-    pub fn storage_partition(&self) -> PartitionType {
-        BlockingPartition::new(self.flash_mutex, self.storage_offset, self.storage_size)
-    }
+/// Fire-and-forget command queue. The USB DFU proxy (ISR context) pushes; the
+/// [`RmkDfuInterface`] updater task pops. Protected by a critical-section
+/// mutex so it is safe from both contexts.
+#[cfg(feature = "_dfu")]
+pub(crate) static DFU_CMD: BlockingMutex<CriticalSectionRawMutex, RefCell<heapless::Vec<DfuCmd, DFU_CMD_QUEUE_SIZE>>> =
```
HaoboGu: DFU_CMD/DFU_WAKE/DFU_BUSY: Is there a reason not to use `embassy-sync::channel`?

rmk/src/dfu/mod.rs
```diff
@@ -153,263 +255,143 @@ pub fn is_dfu_unlocked() -> bool {
...
-                // — adaptive back-pressure without a fixed timeout.
-                // Uses a volatile store because `buf` is still borrowed by
-                // `resp` at this point.
-                if resp.is_some() && PASSTHROUGH_TARGET.load(Ordering::Acquire) != usize::MAX {
-                    unsafe {
-                        core::ptr::write_volatile(buf_ptr.add(4), 4u8);
+#[cfg(feature = "_dfu")]
+impl<DFU: NorFlash, STATE: NorFlash> Runnable for RmkDfuInterface<DFU, STATE> {
+    async fn run(&mut self) -> ! {
+        loop {
```
HaoboGu, line 467: Should `mark_booted` be called here? 

For example, add `self.updater.mark_booted()` before the `loop`

Then the user-space's mark booted can be removed, and `mark_booted` public API can be deleted

rmk/src/usb/mod.rs
```diff
@@ -240,6 +261,236 @@ pub(crate) fn new_usb_builder<'d, D: Driver<'d>>(
...
+                // ── Flow control: dfuDNBUSY override ────────────────────────
+                // Byte 4 of the 6-byte GETSTATUS response is the DFU state.
+                // While the updater queue is non-empty we override it with
+                // 4 (= dfuDNBUSY) so the host polls again after 50 ms instead
+                // of queueing more 512B blocks. The store is volatile because
+                // `buf` is still borrowed by `resp`.
+                let buf_ptr = buf.as_mut_ptr();
+                let resp = self.central.control_in(req, buf);
+                if resp.is_some() && DFU_BUSY.load(Ordering::Acquire) {
+                    unsafe { core::ptr::write_volatile(buf_ptr.add(4), 4u8) };
```
HaoboGu: Idea: check `DFU_BUSY` before calling `self.central.control_in` instead of patching its reply afterwards: while it is set, answer GETSTATUS with dfuDNBUSY and a short bwPollTimeout and skip `DfuState`
This way `DfuState` only moves forward after the flash write really finished, so the `write_volatile` hack goes away, the host no longer waits the hard-coded 50 ms per block, and `Start`/`SystemReset` don't need to go through the queue anymore.

examples/use_rust/nrf52840_dfu_ext/src/main.rs
```diff
@@ -0,0 +1,136 @@
...
+
+    let ext_flash = ExternalFlash::new(dfu_spi, dfu_cs, 8 * 1024 * 1024);
+
+    let dfu_mutex =
+        embassy_sync::mutex::Mutex::<embassy_sync::blocking_mutex::raw::CriticalSectionRawMutex, _>::new(ext_flash);
+    let dfu_partition = Partition::new(&dfu_mutex, 0, 8 * 1024 * 1024);
+
+    // Internal flash: only the boot state and storage partitions are used, so
+    // the internal DFU partition from the linkerscript layout is discarded.
+    let flash_mutex = FlashMutex::new(async_flash_wrapper(Nvmc::new(p.NVMC)));
```
HaoboGu, line 74: Use `nrf_mpsl::Flash::take(mpsl, p.NVMC)`, not `async_flash_wrapper`?

rmk/src/dfu/split/peripheral.rs
```diff
@@ -1,71 +1,99 @@
...
 
-impl SplitDfuHandler {
-    /// Create a new handler from the global [`DfuFlashManager`].
-    /// Returns `None` if `init_flash` has not been called yet.
-    pub fn new() -> Option<Self> {
-        let mgr = get_manager()?;
-        Some(Self {
-            dfu_partition: mgr.dfu_partition(),
-            state_partition: mgr.state_partition(),
+impl<DFU: NorFlash + Clone, STATE: NorFlash + Clone> SplitDfuHandler<DFU, STATE> {
```
HaoboGu: I feel that `SplitDfuHandler` is quite similar with `RmkDfuInterface`, can they be merged? 

Just an idea: process DFU logic & write using a separate `DfuSession`, and both usb & split message just do the forwarding? Actually the idea is quite similar as `RynkSession`, and I also want to include the firmware update feature to the rynk protocol. Maybe we can make the DFU processing a part of rynk? and then, USB/BLE transport is automatically included and the overall architecture is simplified a lot.

rmk-macro/src/codegen/feature.rs
Schievel1: Not really needed, but i wanted to have this instead of playing around with the cfg features everywhere.  The conditions was more complicated before I started refactoring the features.
Should I use cfg!(feature = "_dfu") directly instead?

// just wanted to point out, we can not use is_feature_enabled() because it is reading the users Cargo.toml which has no _dfu feature directly. I keep forgetting that.

rmk/src/dfu/mod.rs
Schievel1: hm with the updated bootloader people can use the bootloader's dfu mode to flash firmware. If it's not the noswap bootloader, it writes the firmware onto the DFU-Partiton and applies the same swap logic like it does when we flash at RMK's runtime. 
They still need a way to mark the firmware booted in order to prevent the bootloader from rolling back. But they don't need to use RMK's DFU mode (and no RmkDfuInterface).

Also, when using dfu_split and flashing the peripheral it does not necessarily run a RmkDfuInterface. 

We could add it in RmkDfuInterface, but imo this makes the usage more complicated instead of simplifying things. Because then you have to explain to users 'in this condition you need to call mark_booted yourself, but in this other condition its not needed' instead of 'when using dfu call mark_booted()' and be done with it.

rmk/src/dfu/mod.rs
Schievel1: I was under the impression that you can't use channel in ISR context, but apparently you can. I will have a go at exchanging this with channel. The command channel for the peripheral as well. Although for peripheral this could be more complicated because we have multiple consumers that must leave a message alone of its not for them.

rmk/src/dfu/split/peripheral.rs
Schievel1: I also had the idea to merge them into one, but refrained from the idea because I tought the Splithandler is too different:
- RmkDfuInterface has FirmwareUpdater, SplitDfuHandler does that directly. 
- SplitDfuHandler has the CRC readbacks and whatnot, that FirmwareHandler does not provide that.
- SplitDfuHandlers get directly called in the peripherals message handling, theres not channel. 

If anything in the long run we could get rid of FirmwareUpdater and roll our own that works for both. 

> Maybe we can make the DFU processing a part of rynk?

Yes I was thinking about the same thing. After this PR i wanted to start using Rynk to implement the OTA Updates via bluetooth. I havent looked that much into Rynk yet but if it also works for USB DFU, I think it would be good to offer that as well.
 But i think we should keep the updates via USB using dfu-util in. dfu-util is kind of an industry standard by now.

rmk/src/dfu/mod.rs
HaoboGu: yeah, if the userspace's `mark_booted()` cannot be fully removed, I prefer to keep it as is

rmk-macro/src/codegen/feature.rs
HaoboGu: > Should I use cfg!(feature = "_dfu") directly instead?

yeah I think so

rmk/src/usb/mod.rs
Schievel1: good idea, I will try that

examples/use_rust/nrf52840_dfu_ext/src/main.rs
Schievel1: mpsl what? This example doesn't even use BLE.

rmk/src/dfu/mod.rs
HaoboGu: If it has multiple consumers, use the event system/pubsub channel?

examples/use_rust/nrf52840_dfu_ext/src/main.rs
HaoboGu: oh I didn't realize that. For BLE examples mpsl should be used, and for non-BLE example, the current code is correct

rmk/src/dfu/split/peripheral.rs
HaoboGu: Yes, I agree, just an idea, both can coexist I think. We can leave it as a future plan.

rmk/src/dfu/mod.rs
Schievel1: nah what the multiple consumers need to do is peeking whether the message is for them and only taking it out if it is. 
Anyway, I will get this done in one way or the other

rmk/src/dfu/split/peripheral.rs
Schievel1: Hm but maybe I'm gonna tackle the merge of SplitDfuHandler and RmkDfuInterface in this PR. Also ProxyUsbDfuHandler and PassthroughDfuHandler basically do the same thing, getting DFU commands from USB and puttng them in channels. 

I convert back to draft for the time being and report back

Schievel1:
Ok, little design change, we got rid of the separate Handlers for central and peripheral. There is one single `ProxyUsbDfuHandler` which does the USB stuff and sends the DFU commands down a channel. 
The central has a `FlashDfuHandler` which receives from the channel and writes to flash. 
If the DFU commands in the channel are for a peripheral, it leaves them alone and instead `run_peripheral_manager` takes them out and sends them via split to the peripheral.
There, peripheral's own `FlashDfuHandler` gets called from `SplitPeripheral::run()` with the messages and writes the message into flash. 

The design is really clean now, I like it. Hope you do, too. :)

```
//! ┌──────────────────────────────────────────────────────────────────┐
//! │  Host (dfu-util / WebUSB)                                        │
//! │    USB Control Transfer (GET_DESCRIPTOR / DFU_DNLOAD)            │
//! └──────────────┬──────────────────────────────────┬────────────-───┘
//!                │                                  │
//!                ▼                                  ▼
//! ┌─-─────────────────────────┐        ┌──────────────────────────────┐
//! │  UsbDfuIface              │        │  ProxyUsbDfuHandler          │
//! │  (USB control handler)    │        │  (ISR → DFU_CHANNEL)         │
//! │                           │        │                              │
//! │  alt 0 → Central          │        │  target = DfuTarget::Central │
//! │  alt 1 → Peripheral(0)    │        │  writes: DfuCmd::Write(tgt,  │
//! │  alt 2 → Peripheral(1)    │        │          offset, data[512])  │
//! │  ...                      │        └──────────┬───────────────────┘
//! └──-────────────────────────┘                   │
//!                                                 │
//!                           DFU_CHANNEL (cap 4)   ▼
//! ┌───────────────────────────────────────────────────────────────┐
//! │                                                               │
//! │  ┌─── PeripheralManager (central event loop) ──────────────┐  │
//! │  │  peek DFU_CHANNEL for DfuTarget::Peripheral(n)          │  │
//! │  │  forward as SplitMessage::FirmwareChunk → split link    │──────────┐
//! │  │  → peripheral FlashDfuHandler                           │  │       │
//! │  └─────────────────────────────────────────────────────────┘  │       │
//! │                                                               │       │
//! │  ┌─── FlashDfuHandler (central) ─────────────────────────-─┐  │       │
//! │  │  peek DFU_CHANNEL for DfuTarget::Central                │  │       │
//! │  │  start → write_chunk(offset, data[512]) → finish        │  │       │
//! │  │  erase on demand, NorFlash::write to DFU partition      │  │       │
//! │  │  finish → sanity check (MSP+reset vector)               │  │       │
//! │  │         → mark_updated_and_reset()                      │  │       │
//! │  └──────────────────────────────────────────────────────-──┘  │       │
//! └───────────────────────────────────────────────────────────────┘       │
//!                                                                         │
//!                                                ┌─────split link (UART)──┘
//!                                                ▼                         
//! ┌─── Peripheral (direct calls, no channel) ─────────────────────┐       
//! │  FlashDfuHandler::write_chunk(offset, data)                   │
//! │  FlashDfuHandler::compute_dfu_crc() → FirmwareCrcReport       │
//! │  mark_updated_and_reset() only on FirmwareCrcOk               │
//! └───────────────────────────────────────────────────────────────┘
```

rmk/src/dfu/mod.rs
```diff
@@ -1,439 +1,555 @@
...
+///
+/// `state` is the boot state partition — typically built with
+/// [`partitions_from_linkerscript`].
+///
+/// Must be called *after* the firmware is confirmed running.
+///
+/// When using [`FlashDfuHandler`], this is called automatically — but users
+/// need this free function if they use a swap-based bootloader outside of RMK
+/// and don't have a `FlashDfuHandler`.
+pub async fn mark_booted<STATE: NorFlash>(state: &mut STATE) {
```
HaoboGu, line 181: Is there a way to ensure that the `mark_booted` is called only once after the firmware starts? It panics when it's called twice.

use_config + usb + dfu examples panic due to the second call of `mark_booted`

rmk/src/dfu/mod.rs
```diff
@@ -1,439 +1,555 @@
...
+        loop {
+            // Drain the shared command channel. DFU_CHANNEL is a single MPSC
+            // shared by central and all peripherals. Each consumer peeks the
+            // head without removing it, checks whether the command is for
+            // them, and either receives (removes) it or leaves it for the
+            // next consumer. This inner loop runs until the channel is empty
+            // (or only contains peripheral commands), then yields once to
+            // the executor before re-checking.
+            loop {
+                match DFU_CHANNEL.try_peek() {
```
HaoboGu: `loop` + `try_peek` is not recommended. I prefer to have a dedicated channel and use `channel.receive().await`. Async design is better than loop + polling.

rmk-macro/src/codegen/registered_processor.rs
```diff
@@ -21,7 +19,7 @@ pub(crate) fn expand_registered_processor_init(
     initializers.extend(i);
     executors.extend(e);
 
-    if is_feature_enabled(rmk_features, "dfu_rp") || is_feature_enabled(rmk_features, "dfu_nrf") {
+    if cfg!(feature = "_dfu") {
         create_dfu_led_processor(hardware, &mut initializers, &mut executors);
```
HaoboGu, line 23: should `split_side_dfu` be used here

rmk/src/dfu/mod.rs
```diff
@@ -1,439 +1,555 @@
...
+                    }
+                }
+            }
+            DfuCmd::SystemReset(DfuTarget::Central) => {
+                #[cfg(all(
+                    target_arch = "arm",
+                    target_os = "none",
+                    any(target_abi = "eabi", target_abi = "eabihf")
+                ))]
+                cortex_m::peripheral::SCB::sys_reset();
```
HaoboGu: It supports only cortex-m, use `crate::boot::reboot_keyboard()` instead?

rmk/src/dfu/mod.rs
Schievel1: No, it just should not be called where use_config does it. I forgot to remove it from the macros.

But the panic because of the second init of the staticcell there is just because this is a relic from embassy-usb-dfus FirmwareUpdater. I don't see a reason why we should need static there, but I could be mistaken.

rmk-macro/src/codegen/registered_processor.rs
Schievel1: you mean `dfu_split`? If so, no. There are other ways than `dfu_split` to flash the peripheral via dfu. E.g. one is the normal dfu flash over USB like in the example `rp2040_usb_dfu_split`.
this doesnt need `dfu_split` so the LED wouldn't work.
Also `dfu_ble` is going to flash directly from host to keyboard using rynk, so that would be the next feature that breaks led support for peripheral.

rmk/src/dfu/mod.rs
HaoboGu: maybe the static can be removed somehow?

rmk-macro/src/codegen/registered_processor.rs
HaoboGu: If understand the code correctly, the `hardware.dfu` defaults to the central's dfu config, when the peripheral also has a LED, the `hardware.dfu` is wrong.

rmk/src/dfu/mod.rs
Schievel1: Yes it is removed. Idk why github doesn't show "outdated" on this one

rmk/src/dfu/mod.rs
Schievel1: True. It is waiting on the channel to have a message now. 
During transfer it is running in a tight loop but that is alright I think as long as it's only during transfer.

HaoboGu: Nice!

rmk/src/dfu/mod.rs
Schievel1: sure thing, I wasn't aware this exists.

rmk-macro/src/codegen/registered_processor.rs
Schievel1: Ah now I get it. You're talking about the parameter, not the feature gate. 
Yes correct, this is wrong. I will overwrite `hardware.dfu` for split if a dfu config for the split exists.

rmk/src/dfu/mod.rs
```diff
@@ -1,439 +1,548 @@
...
+        matches!(self.target(), DfuTarget::Peripheral(id) if (id as usize) < MAX_DFU_ALTS)
     }
 }
 
-// ---------------------------------------------------------------------------
-// DfuStringProvider
-// ---------------------------------------------------------------------------
+/// Command channel: the USB DFU proxy (ISR context) sends via
+/// [`DFU_CHANNEL`]; the [`FlashDfuHandler`] updater task receives.
+pub(crate) static DFU_CHANNEL: Channel<CriticalSectionRawMutex, DfuCmd, DFU_CMD_QUEUE_SIZE> = Channel::new();
```
Schievel1: Maybe it is better to change this to pubsub instead of doing that whack implementation with the `try_peek()` and signals. 
I deliberately chose this in order to safe RAM, but I don't really like the `run()` method of `FlashDfuHandler` with all the `ready_to_receive`, `try_peek`, `try_receive` and whatnot.

Schievel1: The RAM usage isn't that much more, since the payload is by far the biggest weight here, the capacity of the channel is the most important thing when it comes to RAM. And both CAPs of the channels are 4. 

I like the design with pubsub channel way better than before. Let me know what you think.

//  in order to use the channels `is_empty()` in the ISR I added the `empty()` method in the #event macro and made the used mutex configurable.

rmk/src/dfu/mod.rs
```diff
@@ -1,491 +1,588 @@
...
+        storage_offset: core::ptr::addr_of!(__bootloader_storage_start) as usize as u32,
+        storage_size: core::ptr::addr_of!(__bootloader_storage_end) as usize as u32
+            - core::ptr::addr_of!(__bootloader_storage_start) as usize as u32,
+    }
+}
 
-#[cfg(any(feature = "dfu_rp", feature = "dfu_nrf"))]
-use embassy_embedded_hal::flash::partition::BlockingPartition;
+/// Mutex guarding the flash, shared by all partitions.
+pub type FlashMutex<F> = embassy_sync::mutex::Mutex<CriticalSectionRawMutex, F>;
```
HaoboGu: Use crate::RawMutex by default? `CriticalSectionRawMutex` is not required actually

rmk/src/dfu/mod.rs
```diff
@@ -1,491 +1,588 @@
...
-        let deadline = embassy_time::Instant::now() + embassy_time::Duration::from_secs(10);
-        loop {
-            if DFU_STARTED.load(Ordering::Acquire) {
-                info!("dfu_lock: DFU download started, staying unlocked");
-                break;
+            info!("dfu_lock: DFU activity detected, unlock window open for 10 s");
+            info!("dfu_lock: waiting for unlock keys");
+            publish_event(DfuStatusEvent::new(DfuStatus::LockWaiting));
+
+            // 2. Unlock-Key-Polling (10 s)
```
HaoboGu: Should dfu_lock a separate lock besides host_lock? Maybe we need a unified lock machanism. Leave it to the future work maybe.

rmk/src/dfu/mod.rs
HaoboGu: I agree! I think the current shape is great.

rmk/src/driver/w25q.rs
```diff
@@ -0,0 +1,251 @@
...
+const CMD_JEDEC_ID: u8 = 0x9F;
+
+const PAGE_SIZE: u32 = 256;
+const SECTOR_SIZE: u32 = 4096;
+const BLOCK64_SIZE: u32 = 65536;
+
+/// [`NorFlash`] implementation for W25Q and compatible 25-series SPI NOR flash
+/// chips. Uses standard JEDEC commands valid across Winbond, Macronix, ISSI,
+/// and similar families.
+pub struct W25qNorFlash<BUS: SpiBus, CS: OutputPin> {
```
HaoboGu, line 24: Did you check https://crates.io/crates/w25 ?

rmk/src/split/driver.rs
```diff
@@ -350,136 +354,201 @@ impl<T: SplitReader + SplitWriter> PeripheralManager<T> {
...
-    /// `FirmwareChunkAck` before proceeding to the next.
+    /// Called from the event loop when a [`DfuCmdEvent`] is received via PubSub.
+    /// Forwards DFU commands targeted at this peripheral over the split link.
     ///
     /// On `Finish`, triggers end-to-end CRC verification: the peripheral
     /// reads back its DFU partition, sends the CRC-32, the central
     /// compares, and sends `FirmwareCrcOk` / `FirmwareCrcFail`.
     #[cfg(feature = "dfu_split")]
-    async fn handle_passthrough(&mut self) {
+    async fn handle_dfu_event(&mut self, cmd_event: DfuCmdEvent) {
```
HaoboGu: Move it to `src/split/dfu.rs`?

rmk/src/split/peripheral.rs
```diff
@@ -30,27 +32,47 @@ use crate::state::update_status;
...
     #[cfg(not(feature = "_ble"))] S: Write + Read,
+    #[cfg(all(feature = "dfu_split", not(feature = "_ble")))] DFU: NorFlash + Clone,
+    #[cfg(all(feature = "dfu_split", not(feature = "_ble")))] STATE: NorFlash + Clone,
 >(
     #[cfg(feature = "_ble")] id: usize,
     #[cfg(feature = "_ble")] controller: C,
     #[cfg(feature = "_ble")] address: [u8; 6],
     #[cfg(not(feature = "_ble"))] serial: S,
+    #[cfg(all(feature = "dfu_split", not(feature = "_ble")))] dfu_partition: DFU,
+    #[cfg(all(feature = "dfu_split", not(feature = "_ble")))] state_partition: STATE,
```
HaoboGu: Should `DFU+STATE` in a separate Runnable for peripheral? Just like central's  `dfu_iface`?

rmk/src/usb/mod.rs
```diff
@@ -240,6 +263,249 @@ pub(crate) fn new_usb_builder<'d, D: Driver<'d>>(
...
+/// alt 1..N = split peripherals).
+///
+/// Runs inside the USB interrupt. It never touches flash: every download
+/// `start`/`write`/`finish`/`system_reset` is forwarded to the async
+/// [`FlashDfuHandler`](crate::dfu::FlashDfuHandler) updater task through
+/// the event system ([`DfuCmdEvent`](crate::event::DfuCmdEvent)). The DFU
+/// lock gate (if enabled) is checked here so every DFU start path shares
+/// one place.
+#[cfg(feature = "_dfu")]
+struct ProxyUsbDfuHandler {
```
HaoboGu: Add a `src/usb/dfu.rs`?

HaoboGu (approved):
Some minor comments from me, overall I think the current shape is pretty good!

docs/docs/main/docs/configuration/bootloader.mdx
```diff
@@ -173,16 +192,204 @@ See the [DFU lock section](../user_guide/flash_firmware/use_embassy_boot.mdx#unl
...
+
+```toml title="keyboard.toml"
+[dfu.external_flash]
+# Flash chip driver: "w25q" (built-in) or "custom"
+driver = "w25q"
+# Total flash size in bytes (8 MB = 8388608)
+flash_size = 8388608
+
+[dfu.external_flash.spi]
+instance = "SPI0"
```
HaoboGu, line 253: tx_dma/rx_dma should be added here?

rmk/src/split/peripheral.rs
```diff
@@ -201,47 +219,55 @@ impl<S: SplitWriter + SplitReader> SplitPeripheral<S> {
...
-                                    self.dfu_handler = None;
                                 }
-                            } else {
-                                error!("dfu_split: no active DFU session");
+                            };
+
+                            if ok {
+                                self.split_driver.write(&SplitMessage::FirmwareUpdateConfirm).await.ok();
+                                embassy_time::Timer::after_millis(50).await;
+                                dfu_handler.mark_updated_and_reset().await.ok();
```
HaoboGu: Should the reset be sent by the central?

rmk/src/split/peripheral.rs
```diff
@@ -201,47 +219,55 @@ impl<S: SplitWriter + SplitReader> SplitPeripheral<S> {
...
                             }
```
HaoboGu, line 255:
```rust
else {
  self.dfu_handler = None
}
```

is removed, is it correct?

HaoboGu:
Now for USB peripheral, there are two FlashDfuHandler(one for peripheral's USB iface, and another for central's passthrough). I think the two can be merged to a single one? Because the shared DFU/STATE might cause some issues. I prefer to have only one FlashDfuHandler per device.

rmk/src/split/driver.rs
HaoboGu: And this function is quite similar with `send_firmware_update`

rmk/src/dfu/mod.rs
Schievel1: If I remember correctly I did not use host_lock for dfu because it requires additional software on the host that asks the keyboard to unlock. I only wanted to have plain dfu-util and not require the user to install anything extra.

rmk/src/split/driver.rs
Schievel1: Yeah, but for `send_firmware_update` we have the whole firmware already in flash and we just need to send it over to the split. Here we get it chunk by chunk from the host. 
I tried merging both, but it gets too nitty gritty imo.  This function is already way too big for my taste, when I merge them I introduce a bunch of extra paths that are conditionally taken.

rmk/src/split/peripheral.rs
Schievel1: In it's own runnable it thinks it is central, you can actually do that and run it on a peripheral, if you connect it to a USB you can flash it from the host like it's a central.

rmk/src/dfu/mod.rs
HaoboGu: fine

rmk/src/split/driver.rs
HaoboGu: got it. It's okay!

rmk/src/dfu/mod.rs
Schievel1: True that. I first thought we need CriticalSectionRawMutex, because we were working in ISR context, but not that the flash is separated, we aren't. And the ISR is not even an ISR but the bottom half. I will fix that, I think I have a bunch of more places where the thread mutex is sufficient.

rmk/src/usb/mod.rs
Schievel1: ok

rmk/src/dfu/mod.rs
HaoboGu: thread mode mutex is enough for almost all single core situation

docs/docs/main/docs/configuration/bootloader.mdx
Schievel1: yes they should. They aren't needed, since they have defaults, but the docs should mention them as option so people know what's up

rmk/src/split/peripheral.rs
Schievel1: not really sure what for. We could only mark_updated here, but leave the reset for later. The only thing we get from that is that the time window between mark updated and the system reset is bigger. A bigger window to confuse people if the reset didn't make it through.

rmk/src/split/peripheral.rs
Schievel1: This is correct, as I reset `offset` in the ProxyUsbDfuHandler now. But it's good you mention it, because I found that I forgot to reset written_len and last_erased_page on the peripheral. So I will add that in and thanks.

Schievel1:
> Now for USB peripheral, there are two FlashDfuHandler(one for peripheral's USB iface, and another for central's passthrough). I think the two can be merged to a single one? Because the shared DFU/STATE might cause some issues. I prefer to have only one FlashDfuHandler per device.

I think this is a mere edge case. If you have dfu_split, you don't want to have direct dfu on the peripheral anyway, with all the usb interrupts and whatnot on this side. For most cases it is either USB DFU or dfu_split.
I could imagine the peripheral could just send DfuCmd into the event channel, when it gets SplitMessage::FirmwareChunk etc. and the FlashDfuHandler handles them as any other. 
The calls in the message handler for the SplitMessage::Firmware* were synchronously waiting for a result from the FlashDfuHandler though. So we have to decouple this and send something back through a channel as well.

Schievel1:
I'm too tired now. I will review my last commit tomorrow morning and then convert to ready again.

HaoboGu:
just go at your own pace, no hurry at all :P

Schievel1:
> just go at your own pace, no hurry at all :P

I will squash the commits later into 2 or 3. just wanted to leave them as is for now for to easily see what I've don't since the last review

Schievel1:
Alrighty, I did that with the channel an decoupled run_rmk_peripheral and the FlashDfuHandler. 
Please review this change once more, and I will then squash everything into two or three commits that make sense.

rmk/src/dfu/mod.rs
```diff
@@ -1,491 +1,692 @@
...
+    /// DFU partition on the very first chunk.
+    pub async fn write_chunk(&mut self, offset: u32, data: &[u8]) -> Result<(), ()> {
+        if offset == 0 {
+            // New download session: drop stale state from an aborted previous
+            // session so its tail can't poison this image's CRC and its erase
+            // cache can't skip erasing page 0. The central always streams
+            // monotonically from 0; chunk-0 retries happen before any other
+            // offset is written, so this is safe.
+            self.written_len = 0;
+            self.last_erased_page = None;
```
HaoboGu: should `DFU_WRITE_FAILED` be cleared here as well?

HaoboGu (approved):
Only 1 more comment from me, I think it's ready to merge

rmk/src/dfu/mod.rs
Schievel1: We already reset it in handle_cmd when we geht a DfuStart. 

But that begs the question why we don't do it in one place, right?
I will change that to do everything in handle_cmd.

Schievel1: done

Schievel1:
> Only 1 more comment from me, I think it's ready to merge

Right, I will squash the commits, sign them all and then mark this ready again. 

Thanks

HaoboGu:
Thanks very much!

PR #1036, opened 2026-08-07
fix(keyboard): consume one-shot modifier on next key press, not release

jpds:
Fix the issue whereby a one-shot mod like `osm!(ModifierCombination::LSHIFT)` can cause a rolled set of keys to be capitalized as the modifier isn't deactivated after it is pressed - rather than released.

### DIFF
--- a/rmk/src/keyboard/oneshot.rs
+++ b/rmk/src/keyboard/oneshot.rs
@@ -163,17 +163,15 @@ impl<'a> Keyboard<'a> {
     /// Update OSM state based on the keyboard event.
     /// Returns `true` if the OSM was consumed (transitioned from Single to None).
     pub(crate) fn update_osm(&mut self, event: KeyboardEvent) -> bool {
-        let quick_release = self.keymap.one_shot_modifiers_config().quick_release;
         match self.osm_state {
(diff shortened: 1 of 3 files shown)


rmk/src/keyboard/oneshot.rs
```diff
@@ -163,17 +163,15 @@ impl<'a> Keyboard<'a> {
     /// Update OSM state based on the keyboard event.
     /// Returns `true` if the OSM was consumed (transitioned from Single to None).
     pub(crate) fn update_osm(&mut self, event: KeyboardEvent) -> bool {
-        let quick_release = self.keymap.one_shot_modifiers_config().quick_release;
```
HaoboGu, line 166: I'm not sure, why removing this option?

jpds: This option isn't used here anymore, but it's used in `keyboard.rs` (the extra quick-release report is sent from there). For this function, it now always consumes Single on press instead of gating that on `quick_release`, which is what fixes the rollover bug.

HaoboGu: Thanks!

HaoboGu:
thanks!

PR #1038, opened 2026-08-09
feat(ble): expose multiple battery services for split peripherals, with configurable user descriptions

### LINKED ISSUE #710: Battery reporting for peripheral split half is missing
The current implementation of `BatteryService` only reports the current battery state of the _central_ split half. The information from the _peripheral_ is only logged to console.

RMK should report both battery states. I think it is possible because zmk has done this as well zmkfirmware/zmk#764.

I am willing to implement this, but probably wise to wait for #707 .

kot149:
## Summary

resolves #710

- Expose battery levels from configured split peripherals through standard Battery Service instances on the central's BLE GATT server, including host-readable and notifiable levels for each peripheral.
- Add configurable `battery_user_description` values for the central and split peripheral Battery Level characteristics through `[ble]`, `[split.central]`, and `[[split.peripheral]]`, with descriptive defaults for each battery service.

## Testing

Tested with a split keyboard (one split peripheral), using [Microsoft Bluetooth LE Explorer](https://apps.microsoft.com/detail/9n0ztkf1qd98) and [zmk-battery-center](https://github.com/kot149/zmk-battery-center) and verified:

- the battery level can be read
- the Characteristic User Description can be read
- the bettery level update is notified

> image

> image

Only tested with a single peripheral. Configurations with two or more peripherals should also work but not tested.

## Specification reference

This implementation follows the Bluetooth SIG Battery Service Specification v1.1, especially the requirements for multiple Battery Service instances and unique Characteristic Presentation Format Description values for each Battery Level characteristic:
https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/BAS_v1.1/out/en/index-en.html


rmk-config/src/resolved/build_constants.rs
```diff
@@ -3,6 +3,13 @@ use serde::Deserialize;
 use crate::{DEFAULT_PASSKEY_ENTRY_TIMEOUT_SECS, MIN_PASSKEY_ENTRY_TIMEOUT_SECS};
 
 const SUBSCRIBER_DEFAULT_CONFIG: &str = include_str!("../default_config/subscriber_default.toml");
+// trouble-host stores a 4-byte header and 6 bytes per CCCD. Keep the table size
```
HaoboGu: I think those constants can be removed -- trouble-host generates them automatically

rmk-config/src/resolved/build_constants.rs
```diff
@@ -205,6 +250,25 @@ impl crate::KeyboardTomlConfig {
     }
 }
 
+fn validate_split_battery_gatt_capacity(count: usize, active_features: &[&str]) -> Result<(), String> {
```
HaoboGu: Remove it too?

rmk-config/src/resolved/build_constants.rs
kot149: agreed, removed in 7b85aae6f0c57dd7341f6e62b5927904708e0d9d

rmk-config/src/resolved/build_constants.rs
kot149: removed in 7b85aae6f0c57dd7341f6e62b5927904708e0d9d

HaoboGu:
Do you have any idea why Flash/RAM increases that much for split central?

rmk-config/src/resolved/build_constants.rs
```diff
@@ -278,6 +315,95 @@ mod tests {
...
+        });
+        config.auto_calculate_parameters();
+
+        let constants = config.build_constants(&["split", "_ble"]).unwrap();
+
+        assert_eq!(constants.split_battery_peripheral_ids, [0, 2]);
+    }
+
+    #[test]
+    fn configless_split_assumes_all_peripherals_may_report_battery() {
```
HaoboGu: This is incorrect, for peripherals which don't have battery detection, this locks the peripheral's battery level to 0?

rmk/src/ble/ble_server.rs
```diff
@@ -18,16 +20,37 @@ use rmk_types::protocol::rynk::{
     RYNK_BLE_CHUNK_SIZE, RYNK_HID_REPORT_SIZE, RYNK_INPUT_CHAR_UUID, RYNK_OUTPUT_CHAR_UUID, RYNK_SERVICE_UUID,
 };
 
-#[cfg(feature = "vial")]
+#[cfg(all(feature = "vial", feature = "split"))]
 #[gatt_server]
 pub(crate) struct Server {
     pub(crate) battery_service: BatteryService,
+    pub(crate) peripheral_battery_services: ConfiguredPeripheralBatteryServices,
```
HaoboGu: For both Rust API and keyboard.toml users, is there a way to disable the peripheral battery service if the peripheral doesn't have battery detection?

rmk/Cargo.toml
```diff
@@ -212,6 +212,7 @@ use_1m_phy = []
 ## Enable split keyboard support.
 split = [
     "trouble-host?/connection-event-queue-size-2",
+    "trouble-host?/client-att-table-size-128",
```
HaoboGu: This increases RAM/Flash usage a lot. I think this is not necessary in default setting. And when users have multiple peripherals, setting it in `.cargo/config.toml`'s `[ENV]` section is preferred maybe?

rmk-config/src/resolved/build_constants.rs
```diff
@@ -82,7 +85,35 @@ impl crate::KeyboardTomlConfig {
...
+                    .iter()
+                    .enumerate()
+                    .filter_map(|(id, peripheral)| peripheral.battery_adc_pin.as_ref().map(|_| id))
+                    .collect(),
+                None => (0..split_peripherals_num).collect(),
+            }
+        } else {
+            Vec::new()
+        };
+        let central_battery_user_description = self
```
HaoboGu, line 113: should it also be `split` gated?

rmk/src/ble/mod.rs
```diff
@@ -767,6 +794,8 @@ async fn run_ble_keyboard<
     let mut ble_hid_server = BleHidServer::new(server, conn);
     let mut ble_led_reader = BleLedReader;
     let mut ble_battery_server = config.enabled.then(|| BleBatteryServer::new(server, conn));
+    #[cfg(feature = "split")]
+    let mut ble_peripheral_battery_server = BlePeripheralBatteryServer::new(server, conn);
```
HaoboGu: Can the server be ignored when `SPLIT_BATTERY_PERIPHERALS_NUM == 0`?

rmk/src/ble/battery_service.rs
```diff
@@ -133,3 +235,244 @@ impl<P: PacketPool> BleBatteryServer<'_, '_, '_, P> {
...
+                peripherals.levels[1],
+                trouble_host::prelude::descriptors::CHARACTERISTIC_USER_DESCRIPTION.into()
+            ),
+            b"Right".as_slice()
+        );
+        assert_ne!(peripherals.levels[0].handle, peripherals.levels[1].handle);
+    }
+
+    #[test]
+    fn peripheral_battery_presentation_format_supports_maximum_configured_id() {
```
HaoboGu: What does this test do?

rmk/src/ble/battery_service.rs
```diff
@@ -133,3 +235,244 @@ impl<P: PacketPool> BleBatteryServer<'_, '_, '_, P> {
...
+        crate::RawMutex,
+        PeripheralBatteryEvent,
+        { crate::PERIPHERAL_BATTERY_EVENT_CHANNEL_SIZE },
+        { crate::PERIPHERAL_BATTERY_EVENT_SUB_SIZE },
+        { crate::PERIPHERAL_BATTERY_EVENT_PUB_SIZE },
+    >,
+}
+
+#[cfg(feature = "split")]
+fn peripheral_battery_slot(peripheral_id: usize) -> Option<usize> {
```
HaoboGu: Remove this function?

rmk-config/src/resolved/build_constants.rs
kot149: You're right. When no `[split]` configuration was available, the build constants previously treated every split peripheral as battery-enabled. I changed the fallback so that no peripheral Battery Services are generated unless the peripherals are explicitly configured with `battery_adc_pin`. commit: 39b12cbc.

rmk/Cargo.toml
kot149: Agreed. I removed `client-att-table-size-128` from the default `split` feature, so split builds now use trouble-host's default client ATT table size. Projects that need additional capacity can configure it explicitly in .cargo/config.toml `TROUBLE_HOST_CLIENT_ATT_TABLE_SIZE = "128"`.

commit:  50705474

rmk-config/src/resolved/build_constants.rs
kot149: The peripheral battery ID resolution is already gated by the downstream `split` feature and returns an empty list for non-split builds

rmk/src/ble/battery_service.rs
kot149: This is a boundary test for the Characteristic Presentation Format description value.

Peripheral IDs are zero-based, while the assigned description value is `peripheral_id + 1`. Therefore, the maximum supported peripheral ID, 254, must be encoded as description `0x00ff`.

I renamed the test to make that mapping explicit: `peripheral_battery_presentation_format_maps_id_254_to_assigned_description`

Updated in 39b12cbc.

rmk/src/ble/battery_service.rs
kot149: Agreed, it was a thin wrapper with only one call site, so I removed it and now call `find_peripheral_battery_slot` directly with `SPLIT_BATTERY_PERIPHERAL_IDS`. commit: 39b12cbc.

rmk/src/ble/mod.rs
kot149: Yes. The Peripheral Battery Server constructor is now evaluated only when `SPLIT_BATTERY_PERIPHERALS_NUM > 0`.

When the count is zero, it no longer creates a `PeripheralBatteryEvent` subscriber or runs the server initialization. The event itself remains available because displays and user-defined processors may still consume peripheral battery events independently of the GATT service.

Fixed in 39b12cbc.

rmk/src/ble/ble_server.rs
kot149: Fixed in 39b12cbc: Peripheral Battery Services are now generated only for peripherals configured with `battery_adc_pin`. For Rust API builds without a `keyboard.toml`, they are disabled by default; Rust API users can enable them through `keyboard.toml` specified by `KEYBOARD_TOML_PATH` .

kot149:
> Do you have any idea why Flash/RAM increases that much for split central?

The largest avoidable cause was enabling trouble-host's 128-byte client ATT table for every split build, as you mentioned. I removed the default ATT table expansion and now generate Peripheral Battery Services only for peripherals with battery_adc_pin configured. This reduced the split central overhead as follows:

- use_config: .text +4,920 -> +2,956 bytes, .data +128 -> 0 bytes, .bss +1,864 -> +1,080 bytes
- use_rust: .text +4,136 -> +2,040 bytes, .data +128 -> 0 bytes, .bss +1,864 -> +680 bytes

The remaining increase comes from the new GATT descriptors and, when configured, the additional Battery Service and notification handling. The common +272 B .bss in BLE builds is mainly the cost of the two new descriptors on the central Battery Level characteristic. Split peripheral builds now have no .bss increase.

rmk/src/ble/battery_service.rs
```diff
@@ -10,18 +10,120 @@ use trouble_host::prelude::*;
...
+pub(crate) struct PeripheralBatteryServices<const N: usize> {
+    pub(crate) levels: [Characteristic<u8>; N],
+}
+
+#[cfg(feature = "split")]
+pub(crate) type ConfiguredPeripheralBatteryServices =
+    PeripheralBatteryServices<{ crate::SPLIT_BATTERY_PERIPHERALS_NUM }>;
+
+#[cfg(feature = "split")]
+impl<const N: usize> PeripheralBatteryServices<N> {
```
HaoboGu: Macro from trouble-host should be used here? Actually I think there's no need to add const generics here -- use `SPLIT_BATTERY_PERIPHERALS_NUM` is always fine? There will never be 2 `PeripheralBatteryServices` I think

rmk-config/src/resolved/build_constants.rs
```diff
@@ -83,7 +86,35 @@ impl crate::KeyboardTomlConfig {
...
             rmk.split_peripherals_num
         };
-
+        let split_battery_peripheral_ids = if active_features.contains(&"split") {
+            match &self.split {
+                Some(split) => split
+                    .peripheral
+                    .iter()
+                    .enumerate()
+                    .filter_map(|(id, peripheral)| peripheral.battery_adc_pin.as_ref().map(|_| id))
```
HaoboGu, line 106: So if a peripheral configures a `battery_user_description`, but no `battery_adc_pin` is configured, what will happen? Will this kind of configuration be rejected?

rmk/src/ble/battery_service.rs
```diff
@@ -10,18 +10,120 @@ use trouble_host::prelude::*;
 use super::ble_server::Server;
 use crate::ble::sleep::SLEEPING_STATE;
 use crate::core_traits::Runnable;
+#[cfg(feature = "split")]
+use crate::event::PeripheralBatteryEvent;
 use crate::event::{BatteryStatusEvent, SubscribableEvent};
 use crate::keyboard::LAST_KEY_TIMESTAMP;
 
+const CPF_FORMAT_UINT8: u8 = 0x04;
```
HaoboGu: What does `CPF` mean?

kot149: It means Characteristic Presentation Format. I'll expand the constant names instead of using the CPF abbreviation to make this clearer.

rmk-config/src/resolved/build_constants.rs
kot149: Currently it is silently ignored. I agree that's confusing, I'll reject `battery_user_description` on a peripheral unless `battery_adc_pin` is also configured.

rmk/src/ble/battery_service.rs
kot149: Agreed that the const generic is unnecessary for the actual configured service since there is only one `PeripheralBatteryServices` instance. I'll simplify it to use `SPLIT_BATTERY_PERIPHERALS_NUM` directly.
I think the manual `AttributeTable` construction is still useful here because the presentation format description and user description differ for each peripheral instance.

rmk/src/ble/battery_service.rs
kot149: fixed: c5defe9e009a86ecab32daa43da9ecbf6d578dce

rmk-config/src/resolved/build_constants.rs
kot149: fixed: de175b58392d3c97c623f9e264f13cf7d2001c95 and 5358a87ee87b10665e5ee214f7da3bfed45071a2

rmk/src/ble/battery_service.rs
kot149: fixed in edc139020cf718cee804389ce4527dfcad323940.

HaoboGu (commented):
It looks good! I have only 1 more comment: when the split peripheral disconnects, will the battery level for peripheral be set to 0 or marked disconnected somehow?

kot149:
> when the split peripheral disconnects, will the battery level for peripheral be set to 0 or marked disconnected somehow?

The Battery Service does not have a standard value for "disconnected" or "unknown", and 0 means an actual 0% battery level.
Currently, the last reported battery level is kept until the central is restarted, and when the central starts up with the peripheral disconnected, the GATT Battery Level remains at its initial value of 0%, since the last reported value is only stored in RAM. This might seem inconsistent, but treating 0% as disconnected also feels unnatural to me.

HaoboGu:
That's fine, thank you very much for your effort! I'm merging it

alxzh:
How did you make it show up in the zmk-battery-center? Does it require any special configuration?

kot149:
@alxzh 
> How did you make it show up in the zmk-battery-center? Does it require any special configuration?

No special configuration is required; only `battery_adc_pin` for your split peripheral is needed (`battery_user_description` is optional).
Updating the GATT structure might require re-pairing the BLE host.

PR #1043, opened 2026-08-12
fix(keyboard): deduplicate held keycodes in the keyboard report

jpds:
Whilst testing another change, I rolled the brackets row on my symbol layer (`>]})`) - I discovered the the `}` would disappear with a rapid roll as it has the same HID code as `]`.


HaoboGu:
Inserting `self.send_keyboard_report_with_resolved_modifiers(true).await;` to the register key function is not a good idea. It's weird to "send another keyboard report" while registering a new key. I prefer solutions like caching duplicate held keys or caching held positions.

jpds:
@HaoboGu Amended this with a `build_keyboard_report()` to dedup the held keycodes.

HaoboGu:
neat, thanks!

PR #1048, opened 2026-08-12
perf(matrix): idle-scan when no key is pressed or debouncing

jpds:
This branch adds `matrix_idle_scan_ms` which matrix scans will sleep for between key presses, instead of busy-polling continuously.

A held or bouncing key keeps the scan at its active rate, so releases and chord changes still land immediately.


HaoboGu:
I don't think this is needed?

For wireless keyboards, `async_matrix` should be enabled and when there's no key held, the keyboard just waits for interrupt, no scanning is performed. And for wired keyboards, always scanning in maximum speed is preferred I think

jpds:
My :keyboard: isn't wireless, and I think there is a benefit to not having a small ARM CPU spin cycles continuously (even if it's just for power or wear-and-tear). ZMK also ships with this option+setting: https://github.com/zmkfirmware/zmk/blob/main/docs/docs/config/kscan.md whilst, interestingly, there is a request for QMK to add this: https://github.com/qmk/qmk_firmware/issues/23209 .

Beyond that, I don't think there's a perceptible difference between always-on and 10ms.

HaoboGu:
For power or wear-and-tear, you can also enable `async_matrix` in wired keyboard,  which I think is much better.

jpds:
Just tried re-enabling async matrix on my ZSA Voyager and I remembered why I abandoned this months ago: it only has an MCU on the left hand side so, whilst that half can be made async - the right still has to run in sync mode/full power.

This branch probably makes more sense for keyboards in a hardware configuration like this.

PR #1060, opened 2026-08-16
fix(via): return persisted layout options from GetKeyboardValue

### LINKED ISSUE #1058: Vial layout options are persisted but never read back — `GetKeyboardValue(LayoutOptions)` always returns 0
### Summary

The VIA/Vial layout-options value is correctly **saved** to storage on
`SetKeyboardValue(LayoutOptions)`, but it is never **read back**: the
`GetKeyboardValue(LayoutOptions)` handler still has a
`// TODO: retrieve layout option from storage` and returns a constant `0`,
and the boot-time restore path only reads `default_layer` from the same
`LayoutConfig` record, discarding `layout_option`. As a result, a user's
layout-option selection (e.g. an alternative bottom-row layout) reverts to
option 0 in the GUI on every reconnect.

Since the Vial client keeps no local copy — it re-reads this value from the
keyboard on every reload and renders whatever the device returns — the
device is the only source of truth, so the selection can never survive a
replug today.

### Environment / versions

- rmk 0.8.2 (crates.io) and current `main` (checked 2026-08-16, by source
  inspection); platform-independent (pure VIA protocol handling).
- Reproduced on hardware: an RP2040 keyboard running rmk 0.8.2 — after
  changing the layout option in Vial, disconnecting and reconnecting shows
  the default option again. A vial-qmk keyboard tested with the same
  procedure retains its selection across the replug, as expected.

### Steps to reproduce

1. Build a keyboard whose `vial.json` defines layout options (`labels`).
2. In Vial, select a non-default layout option.
3. Unplug/replug the keyboard — the GUI shows the default option again.

Code path (main):

- SET (works): `SetKeyboardValue` → `ctx.set_layout_options(...)` → stored in
  `LayoutConfig.layout_option` (`rmk/src/host/via/mod.rs`).
- GET (missing): `rmk/src/host/via/mod.rs` — `ViaKeyboardInfo::LayoutOptions`
  arm returns `let layout_option: u32 = 0;` with the TODO comment.
- Boot restore (missing): `rmk/src/host/storage.rs` `read_keymap` restores only
  `config.default_layer` from `LayoutConfig` and ignores `layout_option`.

### Expected behavior

`GetKeyboardValue(LayoutOptions)` returns the stored value, so the user's
layout-option selection survives reconnects (matching the behavior VIA/Vial
users expect from QMK-based keyboards).

### Actual behavior / Impact

The GUI renders option 0 after every reconnect. Keymap data itself is
unaffected (display-only), but every user of a non-default layout option has
to re-select it each session.

### Suggested fix

Restore `layout_option` at startup alongside `default_layer` (the record is
already being read in `read_keymap`), keep it available to the via handler
(e.g. in the same place the SET path writes through), and return it from
`GetKeyboardValue(LayoutOptions)`. Happy to open a PR if this direction is
acceptable.

### Context

We (Techmechkeys) are migrating our OLSK60 keyboard from QMK to RMK; it uses
Vial layout options for its bottom-row variants, which is how we hit this.
Found by source inspection on 0.8.2, confirmed unchanged on `main`
(2026-08-16), and reproduced on our RP2040 hardware (with a vial-qmk board as
a control, which retains the selection). Also verified against the Vial
client's protocol expectations (vial-gui `keyboard_comm.py`: the client
re-reads the value from the device on every reload and keeps no persistent
copy).

techmech-keeb:
Fixes #1058

`SetKeyboardValue(LayoutOptions)` persists the selection, but `GetKeyboardValue(LayoutOptions)` returned a hardcoded `0` (TODO) and the boot path restored only `default_layer` from `LayoutConfig`, so the Vial GUI reverts to option 0 on every reconnect.

As discussed in #1058, this restores the persisted `layout_option` and returns it through `GetKeyboardValue(LayoutOptions)`:

- Keep the layout option as `KeyMap` runtime state (same pattern as `mouse_buttons`; no public config API change)
- `read_keymap` restores `layout_option` from the stored `LayoutConfig` alongside `default_layer`
- `HostContext::set_layout_options` updates the RAM state first and still forwards to flash under the `storage` feature; without `storage` the value now lasts for the session instead of always reading 0
- Two regression tests: a set→get roundtrip through the via handler, and a storage restore test that fails if `read_keymap` stops restoring `layout_option` (or `KeyMap::build` stops copying it)

Testing: `cargo nextest run --no-default-features --features=split,vial,storage,async_matrix,_ble` (496 passed), `bash scripts/test_all.sh` (all rows green), nightly `cargo fmt --check` (no diffs).

### DIFF
--- a/rmk/src/host/context.rs
+++ b/rmk/src/host/context.rs
@@ -280,10 +280,13 @@ impl<'a> KeyboardContext<'a> {
     }
 
     pub async fn set_layout_options(&self, opts: u32) {
+        self.keymap.set_layout_option(opts);
         #[cfg(feature = "storage")]
         FLASH_CHANNEL.send(FlashOperationMessage::LayoutOptions(opts)).await;
-        #[cfg(not(feature = "storage"))]
-        let _ = opts;
+    }
+
+    pub fn layout_options(&self) -> u32 {
+        self.keymap.layout_option()
     }
 
     pub async fn reset_storage(&self) {
--- a/rmk/src/host/storage.rs
+++ b/rmk/src/host/storage.rs
@@ -107,6 +107,8 @@ impl<F: AsyncNorFlash, const ROW: usize, const COL: usize, const NUM_LAYER: usiz
                 (StorageKey::LayoutConfig, StorageData::LayoutConfig(config)) => {
                     // Restore the default (base) layer set via a `PDF` key
                     behavior.default_layer = config.default_layer;
+                    // Restore the VIA/Vial layout options selection
+                    data.layout_option = config.layout_option;
                 }
                 _ => continue,
             }
--- a/rmk/src/host/via/mod.rs
+++ b/rmk/src/host/via/mod.rs
@@ -55,8 +55,7 @@ impl<'a> VialService<'a> {
                             BigEndian::write_u32(&mut report.input_data[2..6], value);
                         }
                         ViaKeyboardInfo::LayoutOptions => {
-                            // TODO: retrieve layout option from storage
-                            let layout_option: u32 = 0;
+                            let layout_option = self.ctx.layout_options();
                             BigEndian::write_u32(&mut report.input_data[2..6], layout_option);
                         }
                         #[cfg(not(feature = "host_lock"))]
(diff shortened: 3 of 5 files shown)


HaoboGu (approved):
Nice, thanks

HaoboGu:
I'll merge it after you test it on the real hardware

techmech-keeb:
Tested on real hardware — the layout option now survives a reconnect.

**Setup**
- Board: OLSK60v2 (RP2040, 2 MB flash), the same unit that reproduced #1058
- Firmware: this PR's branch (`7e963de`) built via `examples/use_config/rp2040`, with a minimal 2×2 matrix on the board's real matrix pins and a `vial.json` carrying two layout options (a 3-choice one and a boolean one)
- Client: Vial

**Result**
1. Selected a non-default layout option in Vial
2. Unplugged and replugged the keyboard
3. Vial showed the selected option again — before this change the same board always fell back to option 0

I also confirmed `bootmagic` works on this board (holding the configured key while plugging in drops it into the RP2040 bootloader), which is how I flashed back afterwards.

Let me know if you'd like the test config or any other combination checked.

HaoboGu:
Nice, thanks for testing!

PR #1078, opened 2026-08-21
fix(keymap): bounds-check key positions before layer indexing

jpds:
Row and column positions come from matrix scanning - an out-of-range position used to panic with an index-out-of-bounds in `get_action_at`/`set_action_at`.

This branch ensures that these are row, column, and layer validated against the configured sizes, warn, and return `KeyAction::No` on reads and do nothing on writes.

### DIFF
--- a/rmk/src/keymap.rs
+++ b/rmk/src/keymap.rs
@@ -254,12 +254,13 @@ impl KeyMapInner<'_> {
     fn pop_layer_from_cache(&mut self, pos: KeyboardEventPos) -> u8 {
         match pos {
             KeyboardEventPos::Key(key_pos) => {
-                let row = key_pos.row as usize;
-                let col = key_pos.col as usize;
-                let ci = self.cache_index(row, col);
-                let layer = self.layer_cache[ci];
-                self.layer_cache[ci] = self.behavior.default_layer;
-                layer
+                let ci = self.cache_index(key_pos.row as usize, key_pos.col as usize);
+                if let Some(cache) = self.layer_cache.get_mut(ci) {
+                    let layer = *cache;
+                    *cache = self.behavior.default_layer;
+                    return layer;
+                }
+                self.behavior.default_layer
             }
             KeyboardEventPos::RotaryEncoder(encoder_pos) => {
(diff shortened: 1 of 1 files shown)


HaoboGu:
Could you provide a case/config which reproduces this? I thought the row/col was already checked for splits and matrix

jpds:
The paths you mentioned are indeed covered. The issue identified here is with the VIA host interface where `DynamicKeymapGetKeyCode` (0x04) / `DynamicKeymapSetKeyCode` (0x05) take layer/row/col straight from
 the report bytes with no validation (`src/host/via/mod.rs:104-124`)::

```text
rmk/src/host/via/mod.rs  (VIA raw HID, report bytes 0x04/0x05)
└─ KeyMapCtx::get_keycode/set_keycode        host/context.rs:37,56
   ├─ get_action_at                          keymap.rs:165
   │  └─ 💥 self.layers[layer_index(l,r,c)]  keymap.rs:175   ← flat idx wraps across layers
   └─ set_action_at                          keymap.rs:195
      └─ 💥 self.layers[idx] = action        keymap.rs:205
```

VIA passes raw `command_data[0..2]` straight into those, which would cause RMK to crash.

Whilst digging through this, the LLM I was using also found another issue with the event path → layer cache, which is now fixed in the new second commit:

```text
matrix scan / split peer / custom input device → KeyboardEvent
└─ Keyboard::process_inner                   keyboard.rs:376,414,440,448,532,575,645
   └─ morse.rs:266 (morse leader too)
      └─ get_action_with_layer_cache         keymap.rs:222
         ├─ [release] pop_layer_from_cache   keymap.rs:254
         │  └─ 💥 layer_cache[cache_index]   keymap.rs:258  ← read + reset
         ├─ [press] get_action_at            keymap.rs:230  ← safe after commit 1,
         │  │    returns No (≠ Transparent)     so fallthrough happens
         │  └─ save_layer_cache              keymap.rs:234→279
         │     └─ 💥 layer_cache[ci] = l     keymap.rs:283  ← write
``

HaoboGu (approved):
Thanks a lot!

PR #1079, opened 2026-08-22
fix(keyboard): re-strike a shared HID usage so a roll types both keys

### LINKED ISSUE #809: Missed keystrokes when alternating MODIFIER(KC) and KC in quick succession
I have encountered the following problem after flashing RMK on an RP2040-Zero:

- On my main layer, I have bound two keys to produce the following two characters: `<` and `>`. In Vial, they correspond to the keycodes `KC_NONUS_BSLASH` and `LSFT(KC_NONUS_BSLASH)`, respectively.
- When I try to quickly input them in succession, it appears that some of the inputs aren't registered at all. For example, inputting the string `<><` almost always returns something like `<<` or `<>`.

Below is a short clip showing the problem.

https://github.com/user-attachments/assets/1f8d63ca-9a67-4ae7-914f-b6dfdb6ef6c1

snowyukitty:
Closes #809.

## The defect

Two matrix positions can map to the same HID usage. On the reporter's board `<`
is `KC_NONUS_BSLASH` and `>` is `LSFT(KC_NONUS_BSLASH)` — different keys, one
usage, differing only by the shift modifier.

State tracking is already right. `register_keycode` gives each position its own
report slot, so `held_keycodes` holds `NonusBackslash` twice with two distinct
owners. The report is where they merge (`keyboard.rs:1886`):

```rust
for k in self.held_keycodes {
    let code = k as u8;
    if code != 0 && !keycodes[..n].contains(&code) {
```

A USB HID host emits a character on the transition of a usage from absent to
present in the keycode array. A usage that is already present and merely gains
a modifier produces nothing. So rolling `<` into `>` sends:

```
["NonusBackslash"]
["LShift","NonusBackslash"]
["LShift","NonusBackslash"]
[]
```

One make for two intended keystrokes. Roll `<` into `>` and type `<` again and
the host receives `<` `<`, which is the `<<` in the report.

## Why the obvious repair is not available

The deduplication is deliberate. `fix(keyboard): deduplicate held keycodes in
the keyboard report` added it along with
`hid_reports::shared_keycode_roll_keeps_second_key`, whose comment states the
intent: the shared usage stays down until the last holder releases it. Before
that, releasing the *first* of two owners cleared a usage the second owner still
held — a dropped key.

These are two halves of one problem. Removing the deduplication brings back the
dropped key; keeping only the deduplication keeps the lost character. Both have
to hold at once.

## The change

A press that gives an already-held usage an **additional owner** now sends one
extra report first: the ordinary report with only that usage suppressed. The
normal report follows, so the host observes a fresh absent→present transition.

Everything else stays. The suppressed report keeps every other held usage and
every held modifier — it is never an all-clear report — and the usage still
remains down until the last owner releases it.

The decision comes from counting how many report slots hold that usage on
behalf of a registered key event, immediately before and immediately after
registration. That is an observation of what registration did rather than a
prediction of what it would do, which matters in three places:

- a real modifier never enters `held_keycodes`, so it can never re-strike a
  usage — holding a key and tapping `Shift` is one activation plus an
  independent modifier;
- two positions bound to the same plain keycode gain an owner just as a
  `SHIFTED()` pair does, so they roll into two characters too;
- a press that finds no free slot registers nothing and so gains no owner. It
  synthesises nothing and follows the existing rollover behaviour, rather than
  emitting a stroke for a key the report cannot track.

`build_keyboard_report` accepts the usage to suppress. The ordinary send path
passes `HidKeyCode::No`, which suppresses nothing, while a re-strike passes the
shared usage. Both paths use `send_keyboard_report_suppressing` and the existing
`send_report` transport. No
allocation, no feature gate, no new configuration, no change to
`KeyboardReport`'s shape or the six-key array.

One extra report per re-strike is the intended cost. No other report count
changes.

## Tests

`rmk/tests/scenarios/shared_usage_restrike.toml`, twelve cases, plus
`shared_keycode_roll_keeps_second_key` updated rather than replaced — its
subject, that the usage stays down while an owner holds it, is exactly the half
this must preserve, so only the new transition report is inserted into its
expectations and its comment now states both halves.

On this branch all five `RMK_TEST_FEATURESETS` rows pass:

| features | tests |
| --- | ---: |
| *(none)* | 453 passed |
| `vial,host_lock,_no_usb,steno,passkey_entry` | 520 passed |
| `rynk,_ble,split,async_matrix,storage` | 569 passed |
| `dongle,_ble,storage` | 492 passed |
| `dongle,vial,_ble,storage` | 520 passed |

On unmodified `main` with the same test files, the `rynk,_ble,split,…` row
gives **561 passed, 8 failed**. The eight are the ones that assert the new
transition:

`shared_keycode_roll_keeps_second_key`, `rolled_shared_usage_types_both_keys`,
`reverse_roll_shifted_first`, `rolled_same_keycode_two_positions_types_twice`,
`shared_usage_survives_first_release`, `transition_preserves_other_held_keys`,
`transition_preserves_held_modifiers`,
`plain_keycode_restrike_preserves_an_unrelated_held_key`.

The rest pass on `main` too, and they are what keeps the change honest rather
than just proving the symptom: each names a case where a stroke must *not* be
synthesised, so a rule that fired too eagerly would break them.

- `held_key_plus_tapped_real_shift_types_once` and
  `modifier_only_change_sends_no_transition` — a modifier arriving or leaving
  while a usage is held changes the modifier byte and nothing else;
- `sequential_typing_is_unchanged` — the already-correct `<` `>` `<` stream from
  the report is byte-identical after the change;
- `held_key_sends_no_extra_reports` — one key held sends exactly one make and
  one break, so host auto-repeat is not chopped;
- `an_unregistered_seventh_press_synthesises_no_stroke` — with the six-key array
  full, a seventh press mapped to a held usage registers nothing, gains no
  owner, and emits no extra report.

They do not cover every possible false positive; they cover those four classes.

`cargo fmt --all -- --check` on nightly and
`cargo clippy … --all-targets -- -D warnings` are both clean.

## A limit of this validation

The transition report and the make that follows are two consecutive reports,
intended to become two interrupt transfers over USB or two notifications over
BLE HID. A link that coalesced or dropped one would hide the make again.

The scenario harness asserts the report **stream**, so it proves the firmware
emits both. It does not observe an actual transfer on either transport, so it
proves nothing about delivery over USB or BLE, and I have not tested that on
hardware.

### DIFF
--- a/rmk/src/keyboard.rs
+++ b/rmk/src/keyboard.rs
@@ -1486,8 +1486,6 @@ impl<'a> Keyboard<'a> {
             return;
         }
 
-        // Count owners before and after, so what follows depends on what the
-        // registration actually did rather than on what it was expected to do.
         let owners_before = self.usage_owners(key);
 
         if event.pressed {
@@ -1496,9 +1494,8 @@ impl<'a> Keyboard<'a> {
             self.unregister_key(key, event);
         }
 
-        // A press that gives an already-held usage an additional owner needs a
-        // report without that usage before the ordinary one, or the host never
-        // sees the absent->present transition that produces a character.
+        // A press that adds another owner to a held usage needs one report without
+        // that usage first, or the host never sees the absent->present transition.
         if event.pressed && owners_before > 0 && self.usage_owners(key) > owners_before {
             self.send_keyboard_report_suppressing(event.pressed, key).await;
         }
@@ -1506,24 +1503,10 @@ impl<'a> Keyboard<'a> {
         self.send_keyboard_report_with_resolved_modifiers(event.pressed).await;
     }
 
-    /// How many report slots are holding `key`.
+    /// How many report slots hold `key`.
     ///
-    /// Two matrix positions can map to the same HID usage - `<` and `>` on a
-    /// board where `>` is the shifted form of `<`, or simply the same keycode
-    /// bound twice. Each gets its own slot, but the report collapses them to
-    /// one usage, so rolling from one to the other would otherwise leave the
-    /// usage continuously down and the second key would produce nothing.
-    ///
-    /// Comparing this count across the registration keys the decision on
-    /// ownership actually changing. Two positions bound to the same plain
-    /// keycode roll into two characters; a real modifier never enters
-    /// `held_keycodes` so it cannot re-strike anything; and a press that finds
-    /// no free slot registers nothing and so synthesises nothing, leaving the
-    /// established rollover behaviour alone.
-    ///
-    /// Counted over `held_keycodes` alone, because that is the array the
-    /// outgoing report is built from: this measures what the host is about to
-    /// see rather than a bookkeeping view of it.
+    /// Two matrix positions can map to one HID usage, and the report collapses
+    /// the duplicates, so only the slot count separates a re-strike from a hold.
     fn usage_owners(&self, key: HidKeyCode) -> usize {
         if key == HidKeyCode::No {
             return 0;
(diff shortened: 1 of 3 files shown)


HaoboGu:
Please simplify and clean the code. The current code is low-quality

snowyukitty:
Simplified, thanks.

- `build_keyboard_report` now takes the suppressed usage as a parameter itself, so the second builder that used to sit behind it is gone, and `send_keyboard_report_with_resolved_modifiers` calls the suppressing send instead of repeating the send-and-yield pair. Two new functions instead of three, and no delegation layer.
- The comment blocks in `keyboard.rs` and in the scenario header are shortened.

Against the base this branch was cut from, `keyboard.rs` now adds 29 lines and removes 5; before it added 57 and removed 1.

Behaviour is unchanged, and the test counts on every `RMK_TEST_FEATURESETS` row are the same as before (453 / 520 / 569 / 492 / 520). `cargo clippy -- -D warnings` and `cargo +nightly fmt --check` are clean.

If you had a different shape in mind — no new function at all, or the ownership count folded into `register_key` — say which and I'll redo it.

PR #1083, opened 2026-08-23
perf(keyboard): stop one-shot expiry from blocking the keyboard task

jpds:
When a one-shot modifier or layer was released while armed (Single state), the keyboard task waited inline on `select!(Timer::after(timeout), next_event)` until either the timeout expired or another key arrived. This would park the whole task, so held-key repeats, mouse wheel repeats, and every other timed behavior froze until one of those two things happened.

This PR instead stores the expiry as deadlines on the `Keyboard` struct (`osm_deadline`, `osl_deadline`). The main loop in `run()` already races its subscriber against the mouse-repeat deadline; it now races against the earliest of all pending deadlines and calls `fire_oneshot_timeout()` when a one-shot expires.

It also fixes two behavioral bugs:

- Any key event during the armed window used to replace the timer wait (the event went into `unprocessed_events`), so the modifier stayed armed indefinitely if no further key arrived before the original timeout. Intervening keys no longer touch the deadline.
- Quick-release rolling into a tap-hold key had a documented known bug in `one_shot.toml`: the OSM deadline loop timed out before the tap resolved. The roll now applies the modifier correctly.

### DIFF
--- a/rmk/src/keyboard.rs
+++ b/rmk/src/keyboard.rs
@@ -142,7 +142,7 @@ impl Runnable for Keyboard<'_> {
     /// The report is sent using `send_report`.
     async fn run(&mut self) -> ! {
         loop {
-            // TODO: Now the unprocessed_events is only used in one-shot keys and clear peer key.
+            // TODO: Now the unprocessed_events is only used in held_for_5s.
             // Maybe it can be removed in the future?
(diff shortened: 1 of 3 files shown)


HaoboGu:
brilliant, thanks a lot!

PR #1085, opened 2026-08-24
perf(keyboard): profile-key hold no longer blocks the keyboard task

jpds:
Also port the User key to use a deadline timeout, instead of blocking the keyboard task.

Follows on from #1083.

### DIFF
--- a/rmk/src/keyboard.rs
+++ b/rmk/src/keyboard.rs
@@ -1,7 +1,5 @@
 use core::fmt::Debug;
 
-#[cfg(feature = "_ble")]
-use embassy_futures::select::{Either, select};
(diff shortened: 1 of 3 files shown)


HaoboGu:
I'm thinking about a unified approach for hold & repeat actions, what do you think?

jpds:
That actually sounds great, as it would simplify the work as I across the four branches I did for this eager deadline stuff.

jpds:
My final branch for the work is at: https://github.com/rmk-rs/rmk/compare/main...jpds:rmk:perf/combo-reset-deadline - if you want to use that as the base for a rework.

jpds:
Closing for #1109.

PR #1107, opened 2026-08-31
usb: honour SET_PROTOCOL on the boot-subclass keyboard interface

techmech-keeb:
### Summary

The keyboard interface is declared with `HidSubclass::Boot` and `HidBootProtocol::Keyboard` — v0.9.0 added this as *"Advertise the boot keyboard protocol on the primary HID interface, so the keyboard works in BIOS/UEFI and other boot-protocol-only hosts"*. But `UsbRequestHandler` does not override `set_protocol`, so embassy-usb's default implementation rejects `SET_PROTOCOL(Boot)` and `GET_PROTOCOL` always answers report mode.

The descriptor therefore advertises boot support that the device refuses to enter. A host that requires the switch before it will use the keyboard — a BIOS or UEFI setup screen, a KVM switch, a BMC — can be left without a working keyboard, while hosts that parse the report descriptor keep working, so the failure only shows up on some setups. This completes the request-handler side of the boot-protocol support advertised in v0.9.0.

### What this changes

`UsbRequestHandler` keeps the selected mode as a field and answers `SET_PROTOCOL` / `GET_PROTOCOL` from it.

Nothing changes about what the interface transmits: `KeyboardReport` is already the boot keyboard layout from HID 1.11 Appendix B — one modifier byte, one reserved byte and six keycodes, with no report ID — so the same report is valid in either protocol mode. The mode only changes what `GET_PROTOCOL` answers.

Two consequences worth stating plainly:

- The same handler type serves every HID interface, so the composite, steno and Vial interfaces now store and report the mode as well, instead of rejecting `SET_PROTOCOL(Boot)` through embassy-usb's default. HID 1.11 §7.2.5 and §7.2.6 define these requests for devices in the boot subclass, so hosts do not send them to those interfaces; carrying a per-interface flag just to keep rejecting them is more state than this fix needs.
- The mode is not restored to the report protocol on a USB bus reset. embassy-usb's `RequestHandler` gets no reset callback — I opened [embassy-rs/embassy#6891](https://github.com/embassy-rs/embassy/issues/6891) about adding one, and the `TODO` in the code points at it. A host is expected to set the protocol again when it reconfigures the keyboard (HID 1.11 §7.2.6 and Appendix F.4), so what remains after a reset is a stale `GET_PROTOCOL` answer until it does.

### Testing

**Unit tests.** `cargo nextest run --no-default-features --features=split,vial,storage,async_matrix,_ble` → 502 passed. `bash scripts/test_all.sh` → all rows pass. `bash scripts/format_all.sh` leaves no diff. Removing the store in `set_protocol` makes the new test fail, so it covers the behaviour it claims to.

**On hardware.** An earlier revision of this change was verified on an RP2040 keyboard against a Linux host, issuing the control transfers directly with pyusb. Both firmwares expose the same HID interfaces `[(0, 1, 1), (1, 0, 0), (2, 0, 0)]` — if0 declares boot/keyboard in both:

| | before | after |
|---|---|---|
| `SET_PROTOCOL(Boot)` | **STALL** | accepted |
| `GET_PROTOCOL` after that | Report | Boot |

That revision answered the requests from the device-level `Handler` and also restored report mode on a bus reset. The current revision answers them from the HID class's request handler and drops the reset restore, so it has **not** been re-run on hardware. The change was cherry-picked onto our production tree for that test — same code, different base — and typing was unaffected afterwards. I have **not** tested against an actual BIOS/UEFI, KVM or BMC: the hardware evidence here is the control-transfer behaviour, not end-to-end use on a boot-only host.

### Known gaps, unchanged by this PR

`GET_REPORT`, `SET_IDLE`/`GET_IDLE` and idle-rate retransmission are still unimplemented, and the control-pipe `SET_REPORT(Output)` is accepted without driving the LEDs (LED state comes from the interrupt OUT endpoint). Happy to follow up on those separately if you would like them covered.

### DIFF
--- a/rmk/src/usb/mod.rs
+++ b/rmk/src/usb/mod.rs
@@ -3,7 +3,7 @@ use embassy_futures::select::{Either, select};
 use embassy_sync::signal::Signal;
 #[cfg(feature = "usb_log")]
 use embassy_usb::class::cdc_acm::CdcAcmClass;
-use embassy_usb::class::hid::{HidReader, HidReaderWriter, HidWriter, ReportId, RequestHandler};
+use embassy_usb::class::hid::{HidProtocolMode, HidReader, HidReaderWriter, HidWriter, ReportId, RequestHandler};
 use embassy_usb::control::OutResponse;
 use embassy_usb::driver::{Driver, EndpointError};
 use embassy_usb::{Builder, Handler, UsbDevice};
@@ -553,7 +553,11 @@ macro_rules! usb_hid_state_and_config {
         }
 
         let state = paste::paste! { [<$descriptor:snake:upper _STATE>].init(::embassy_usb::class::hid::State::new()) };
-        let request_handler = paste::paste! { [<$descriptor:snake:upper _HANDLER>].init($crate::usb::UsbRequestHandler {}) };
+        let request_handler = paste::paste! {
+            [<$descriptor:snake:upper _HANDLER>].init($crate::usb::UsbRequestHandler {
+                protocol: ::embassy_usb::class::hid::HidProtocolMode::Report,
+            })
+        };
 
         let hid_config = ::embassy_usb::class::hid::Config {
             report_descriptor: <$descriptor>::desc(),
@@ -604,11 +608,26 @@ pub(crate) use add_usb_reader_writer;
 pub(crate) use add_usb_writer;
 pub(crate) use usb_hid_state_and_config;
 
-pub(crate) struct UsbRequestHandler {}
+pub(crate) struct UsbRequestHandler {
+    pub(crate) protocol: HidProtocolMode,
+}
 
 impl RequestHandler for UsbRequestHandler {
     fn set_report(&mut self, id: ReportId, data: &[u8]) -> OutResponse {
         info!("Set report for {:?}: {:?}", id, data);
+        OutResponse::Accepted
+    }
+
+    fn get_protocol(&self) -> HidProtocolMode {
+        self.protocol
+    }
+
+    fn set_protocol(&mut self, protocol: HidProtocolMode) -> OutResponse {
+        // KeyboardReport is already the 8-byte boot keyboard layout, so the
+        // mode only changes what GET_PROTOCOL answers.
+        // TODO: Return to Report on a bus reset once embassy-usb tells the
+        // request handler about it (embassy-rs/embassy#6891).
+        self.protocol = protocol;
         OutResponse::Accepted
     }
 }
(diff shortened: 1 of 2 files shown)


rmk/src/usb/mod.rs
```diff
@@ -604,13 +609,55 @@ pub(crate) use add_usb_reader_writer;
...
 pub(crate) use usb_hid_state_and_config;
 
-pub(crate) struct UsbRequestHandler {}
+/// Set while the host has selected the boot protocol on the boot-subclass
+/// keyboard interface. Shared with `UsbDeviceHandler` so a bus reset can
+/// restore the report protocol (HID 1.11 Appendix F.3); only the keyboard
+/// interface is boot-capable, so one flag is enough.
+// TODO: Move this into `UsbRequestHandler` if embassy-rs/embassy#6891 adds
+// a bus-reset callback to `RequestHandler`.
+static BOOT_PROTOCOL_ACTIVE: AtomicBool = AtomicBool::new(false);
```
HaoboGu: Why use a static? A field in UsbRequestHandler is enough I think

techmech-keeb: Good point — the protocol mode does belong to the keyboard interface. The reason it ended up as a global is that embassy-usb delivers bus reset only through the device-level `Handler`; the HID `RequestHandler` has no reset callback, so a field owned solely by `UsbRequestHandler` would not be reachable when the reset arrives, and HID 1.11 Appendix F.3 requires a boot keyboard to return to the report protocol on reset.

I'll replace the bare global with a small shared `ProtocolState` that both the request handler and the device handler reference, so the state stays interface-scoped. Does that match what you had in mind?

HaoboGu: Maybe merge `UsbRequestHandler` and `UsbHandler` to a single handler?

HaoboGu: Don't create shared resources if not necessary

techmech-keeb: Done — pushed. The protocol mode is now a field on the device handler, which handles `SET_PROTOCOL`/`GET_PROTOCOL` in `control_out`/`control_in` and resets it to report mode in `reset()`. There is no separate static or shared protocol state; `UsbRequestHandler` now only provides the remaining `set_report` hook.

Two details worth flagging in the diff:

- The handler is given the boot keyboard's interface number rather than answering every interface-class request with those request codes. Class request codes are not unique across classes: HID `GET_PROTOCOL` and DFU `GETSTATUS` both use `bRequest` 3. A peripheral serving logger/DFU interfaces but no keyboard is configured with `None`, so this handler leaves those requests alone.
- A malformed `SET_PROTOCOL` request aimed at the keyboard is rejected here rather than passed on, because embassy-usb's HID handler does not validate `wLength`.

rmk/src/usb/mod.rs
```diff
@@ -658,6 +701,37 @@ impl Handler for UsbDeviceHandler {
         }
     }
 
+    fn control_out(&mut self, req: Request, _data: &[u8]) -> Option<OutResponse> {
```
HaoboGu: I don't know whether this is correct, why abandon the `RequestHandler` way? which I think is much cleaner.

rmk/src/usb/mod.rs
```diff
@@ -613,20 +627,46 @@ impl RequestHandler for UsbRequestHandler {
...
     /// Suspended (see `suspended()`), so this always holds Configured while the
     /// device is suspended; kept as a snapshot rather than a hardcoded value so
     /// resume stays correct if another pre-suspend state becomes publishable.
     pre_suspend: UsbState,
+    /// Interface number of the boot-subclass keyboard, or `None` on a build
+    /// with no keyboard interface. The number identifies which interface's
+    /// SET_PROTOCOL/GET_PROTOCOL this handler answers; without it the same
+    /// class request codes would shadow another class on the device, such as
+    /// DFU's GETSTATUS.
+    boot_keyboard_interface: Option<u8>,
```
HaoboGu: Why `Option<u8>`?

techmech-keeb: Done — pushed, and updated the description to match. Taking this as: keep the mode as a plain field on `UsbRequestHandler` and handle `SET_PROTOCOL`/`GET_PROTOCOL` there. The device-level interception and the interface-number state are gone; each HID `RequestHandler` is already attached to one interface.

This gives up restoring the report protocol on a USB bus reset, because embassy-usb's `RequestHandler` has no reset callback (embassy-rs/embassy#6891). RMK's keyboard writer sends the same 8-byte boot-compatible report in either mode, so after a reset the remaining difference is a stale `GET_PROTOCOL` answer until the host sends `SET_PROTOCOL` again.

PR #1119, opened 2026-09-06
fix(ble): keep battery values current during sleep and align peripheral notification gating

kot149:
This updates BLE battery reporting so that sleep suppresses notifications without suppressing GATT characteristic updates.

* Keep central and split peripheral battery GATT values up to date while sleeping.
* Use the shared activity state for central battery notification gating instead of key input only.
* Separates GATT value updates from notifications, allowing battery values to remain readable and current without the additional power cost of notifications during sleep.

### DIFF
--- a/rmk/CHANGELOG.md
+++ b/rmk/CHANGELOG.md
@@ -21,6 +21,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix UTF-8 panics when non-ASCII names are used in `keyboard.toml` aliases, and accept Unicode whitespace between keymap actions
 - Honour `SET_PROTOCOL` on the boot-subclass keyboard interface. The keyboard interface now accepts both modes and reports the selected one
 - Reboot instead of re-running trouble's host runner after it stops, and when a split link the host still holds is unknown to the controller.
+- Expand central BLE battery notification gating beyond recent key input to recent activity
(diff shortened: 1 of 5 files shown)


tompre:
> It makes sense to go for consistency. I will try to do some measurements, when I have some free time.
I noticed a bump in power usage on the peripheral a few weeks back. I hoping this fixes it, although I don't understand the reason for the increased power drain.

Sorry, for the late response. It took me a long time to tracked down my peripheral power drain to the controller board and not the software. I suspect that I have a solder bridge somewhere, but I can't find it.

I could not find a huge difference in power savings from this pull request, although my measurements are not that precise any more, so take it with a grain of salt.
However I can say with certainty, that the difference is really low.

In general, I'm also unsure if we even need to stop reporting battery levels during sleep at all, also for the central side.
@HaoboGu is there a reason why we do that in the first place?
From the BLE perspective it should not be necessary. Battery level should not change that fast, so in the end we will most probably send a notification every few hours or even days?

HaoboGu:
Reporting battery level wakes up the host, which is an unexpected behavior? So disabling the battery report is okay

tompre:
> Reporting battery level wakes up the host, which is an unexpected behavior? So disabling the battery report is okay

I'm so sorry! I was so focused on the keyboard side, that I forgot the host.
I just had a look in the BAS 1.1/0 specification and the behavior is required.

Sorry for my comments.

PR #1126, opened 2026-09-09
feat(vial): make macro count configurable

hitsmaxft:
## Summary

Make the dynamic macro count reported by the Vial protocol configurable instead of always returning 32.

- adds `VialConfig::macro_count` with a backwards-compatible default of 32;
- adds `VialConfig::with_macro_count()` for Rust-configured keyboards;
- adds `[host].vial_macro_count` for `keyboard.toml` users;
- wires the resolved value through `rmk-macro` code generation;
- returns the configured value from `DynamicKeymapMacroGetCount` (`0x0c`).

`vial_macro_count` is deliberately independent of `rmk.macro_space_size`: one controls the number of slots exposed to Vial, while the other controls their total byte storage.

## Validation

- `rmk-config`: 102 unit tests and 8 keyboard TOML validation tests passed;
- `rmk-macro`: 65 unit tests and the compile-fail suite passed;
- exact Vial protocol test passed with `macro_count = 1`;
- `cargo check` passed for `vial,host_lock,storage`;
- Clippy passed with `-D warnings` for `rmk`, `rmk-config`, and `rmk-macro`.

The default remains 32, so existing `VialConfig::new(...)`, `VialConfig::default()`, and `keyboard.toml` configurations preserve the previous wire behavior.

### DIFF
--- a/docs/docs/main/docs/configuration/appendix.md
+++ b/docs/docs/main/docs/configuration/appendix.md
@@ -576,6 +576,8 @@ defmt_log = true
 [host]
 # Whether Vial is enabled (default: true in keyboard.toml config)
 vial_enabled = true
+# Number of dynamic macro slots reported to Vial (default: 32).
+vial_macro_count = 32
 # Whether Rynk is enabled (experimental, default: false in keyboard.toml config)
 # Rynk and Vial are mutually exclusive and must match Cargo features.
 rynk_enabled = false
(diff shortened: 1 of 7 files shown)


docs/docs/main/docs/configuration/appendix.md
```diff
@@ -576,6 +576,8 @@ defmt_log = true
 [host]
 # Whether Vial is enabled (default: true in keyboard.toml config)
 vial_enabled = true
+# Number of dynamic macro slots reported to Vial (default: 32).
+vial_macro_count = 32
```
HaoboGu, line 580: I don't know whether putting it here is correct -- vial macro space is not configured here.

hitsmaxft: Good point. I placed it under `[host]` because it controls the Vial protocol capability, but having the count there while `macro_space_size` lives under `[rmk]` makes the two related limits easy to miss. Would you prefer that I move the TOML option to `[rmk]`, next to `macro_space_size`, while keeping the Rust `VialConfig::with_macro_count()` API? I lean toward moving the TOML option to `[rmk]`.

PR #1131, opened 2026-09-10
fix(ble): advertise a paired profile to its bonded host only

### LINKED ISSUE #901: BLE profile switching unreliable with two bonded hosts in range — advertising is unfiltered, so the wrong host can take the connection
**Setup:** Corne split on nice!nano v2 (`nrf52840_ble` + `split` + `storage`), rev b982049, keyboard.toml config. Two computers on the same desk: one bonded to profile 0, the other to profile 1.

**What happens:** With both machines powered on, switching to profile 0 often never connects — the keyboard just keeps advertising (my status display shows the advertising state indefinitely, 60s+). Switching back to profile 1 works in 1-2s. If I turn Bluetooth off on the profile-1 machine, profile 0 connects fine. So the profiles themselves are fine, they just fight each other.

**What I found digging through the code:** `advertise()` in `rmk/src/ble/mod.rs` always uses `Advertisement::ConnectableScannableUndirected` with the default (unfiltered) `AdvFilterPolicy`, and the adv payload is identical for every profile. A profile only decides which bond gets loaded into the stack (`update_stack_bonds()`). So after a profile switch, every bonded host in range races for the same advertisement and the first CONNECT_IND wins. From reading the code (I don't have a debug probe attached, so this part is inferred, not logged): when the "wrong" host wins, encryption can't succeed against the loaded LTK, and while that connection lingers the keyboard isn't advertising — so the host the profile was actually switched to has no way in until the link drops and the race starts over.

The wireless docs say "If you've connected a host to a profile, other devices will not be able to connect to this profile without manually clearing it first" — that holds at the encryption level, but the other device can still *connect* and occupy the slot, which blocks the host the profile was switched to.

**Suggested fix (patched locally, tested today):** do what ZMK does — when the active profile has a bond, advertise only to that peer:

- `peripheral.set_filter_accept_list(&[peer_identity_addr])` before advertising
- `filter_policy: FilterConnAndScan` when the profile has a bond, `Unfiltered` when it's empty — so pairing new hosts still works, and BT_CLR returns the profile to open advertising
- I gated it on `identity.irk.is_some()`: a bond without an IRK can't be resolved from rotating private addresses, so those fall back to open advertising instead of risking locking that host out

trouble already syncs the controller resolving list from bond IRKs, so RPAs from macOS/Windows resolve before the connection is accepted, and nrf-sdc implements all the needed HCI commands (`LeAddDeviceToFilterAcceptList` etc.). The whole change is ~20 lines in `rmk/src/ble/mod.rs` plus two `ControllerCmdSync` bounds on `BleTransport`.

Tested it today over repeated switches between the two computers: the wrong machine no longer takes the connection slot, and the intended one reconnects each time (with normal host-side latency, a second or two). Before the patch, switching toward the second machine frequently hung in advertising forever.

Happy to open a PR if you'd take this.

aTomi412:
Closes #901

## Problem

Every host profile sends the same unfiltered HID advertisement. With two bonded hosts awake and in range, whichever answers first takes the profile. When it's the wrong one (it holds the key of another profile), encryption fails: the right host is locked out while that connection lingers, and at least Windows 11 (since KB5124008) stops auto-reconnecting after such a failure, so the keyboard only comes back after a manual connect from the host.

## Fix

- On a paired host profile, load the bonded host's identity address into the controller's filter accept list and advertise with `AdvFilterPolicy::FilterConnAndScan` (`Adv::BondedHost`). An unpaired profile still advertises openly, so pairing and `Clear BT` work as before. The existing early-disconnect check from #1024 stays as the second line of defense.
- Hosts that distribute an IRK (Windows, macOS, iOS, Android) connect from rotating private addresses. The controller can only match those against the accept list if it resolves them itself, so the nRF SoftDevice Controller is now built with `support_le_privacy()` (macro codegen and the Rust API examples). trouble-host already syncs the controller's resolving list from the loaded bond; on an SDC without LL privacy those commands were failing quietly.
- Falls back to open advertising when filtering can't work: a controller without LL privacy (checked once via `LeReadLocalSupportedFeatures`) for a bond that has an IRK, a split central with more than one peripheral (its connects rewrite the same accept list), or a failed `set_filter_accept_list`.

The dongle slot (`Adv::Directed` / `Adv::DongleSeeking`) and split peripheral advertising are unchanged.

## Notes for review

- `support_le_privacy()` is a behavior change for all nRF keyboards even without a filtered advertisement: the controller now enables address resolution once a bond with an IRK is loaded. There's no `SDC_MEM_*` term for privacy in `sdc.h`, so the memory pool sizes are unchanged, and the examples below build with the current constants.
- Rust API users need to add `.support_le_privacy()` to their SDC builder (noted in the changelog); without it the fallback path keeps them on today's behavior.

## Testing

- `cargo check`/`clippy` on `split,vial,storage,async_matrix,_ble`, `rynk,_ble,split,async_matrix,storage`, `vial,storage,_ble`, `dongle,vial,_ble,storage`, `split,vial,storage,_ble,subrating`.
- `cargo nextest run` (577 tests) on `rynk,_ble,split,async_matrix,storage`; one new unit test in `adv.rs`.
- Built for the target: `use_config/nrf52840_ble`, `use_config/nrf52840_ble_split`, `use_config/nrf52840_ble_split_dongle`, `use_rust/nrf52840_ble`, `use_rust/nrf52840_ble_split`, `use_rust/nrf52840_ble_split_dongle`, `use_rust/nrf52832_ble`, `use_rust/nrf54l15_ble`, `use_rust/nrf54lm20_ble`, `use_rust/nrf_dongle/central`.
- Hardware: not yet. I'm about to test this on a Cornix (nRF52840 split, one peripheral) against two Windows 11 hosts switching BT0/BT1, and will move the PR out of draft once that's done.

### DIFF
--- a/docs/docs/main/docs/features/wireless.md
+++ b/docs/docs/main/docs/features/wireless.md
@@ -63,7 +63,7 @@ Vial user keycodes can be configured to operate wireless profiles. Suppose that 
 
 Vial also provides a way to customize the displayed keycode, see `customKeycodes` in [this example](https://github.com/rmk-rs/rmk/blob/main/examples/use_rust/nrf52840_ble/vial.json). If `customKeycodes` are configured, the `User0` ~ `User(N+3)` will be displayed as `BT0`, ..., `Switch Output`.
 
-If you've connected a host to a profile, other devices will not be able to connect to this profile: the keyboard immediately disconnects any device that doesn't match the profile's stored bond. To pair a different host to that profile, clear it first, for example by holding the profile's key for 5 seconds.
+If you've connected a host to a profile, other devices will not be able to connect to this profile: the keyboard advertises to the bonded host alone, so another host that also knows the keyboard can't take the profile while switching between them, and it immediately disconnects any device that doesn't match the profile's stored bond. To pair a different host to that profile, clear it first, for example by holding the profile's key for 5 seconds.
 
 ## BLE Passkey Entry
 
--- a/examples/use_rust/nrf52832_ble/src/main.rs
+++ b/examples/use_rust/nrf52832_ble/src/main.rs
@@ -62,6 +62,7 @@ fn build_sdc<'d, const N: usize>(
     sdc::Builder::new()?
         .support_adv()
         .support_peripheral()
+        .support_le_privacy()
         .support_dle_peripheral()
         .support_phy_update_peripheral()
         .support_le_2m_phy()
--- a/examples/use_rust/nrf52840_ble/src/main.rs
+++ b/examples/use_rust/nrf52840_ble/src/main.rs
@@ -76,6 +76,7 @@ fn build_sdc<'d, const N: usize>(
     sdc::Builder::new()?
         .support_adv()
         .support_peripheral()
+        .support_le_privacy()
         .support_dle_peripheral()
         .support_phy_update_peripheral()
         .support_le_2m_phy()
--- a/examples/use_rust/nrf52840_ble_split/src/central.rs
+++ b/examples/use_rust/nrf52840_ble_split/src/central.rs
@@ -78,6 +78,7 @@ fn build_sdc<'d, const N: usize>(
         .support_central()
         .support_adv()
         .support_peripheral()
+        .support_le_privacy()
         .support_dle_peripheral()
         .support_dle_central()
         .support_phy_update_central()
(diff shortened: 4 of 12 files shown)


HaoboGu:
This is a behavior change. When the profile is bonded, but the bonded peer is not near the keyboard, switching to the bonded profile makes the keyboard completely "disappear".

PR #1137, opened 2026-09-10
refactor(keyboard): replace next_deadline()/fire_expired() with a DeadlineSet

jpds:
This branch takes the keyboard timers that were spread across separate systems (including the macro executor blocking the whole task at certain points) and moves everything into a `DeadlineSet` registry.

The old code tracked deadlines in two parallel, hand-maintained lists that had to stay in the same order; missing a clear left `run()` spinning on a deadline that never expires. `DeadlineSet` gives each timeout source a named slot, so the dispatcher clears every due slot before it fires.


HaoboGu:
tbh I don't like this approach, the code and the binary bloat a lot, but I don't see any gain from this PR

PR #1144, opened 2026-09-14
chore: sort `Cargo.toml` dependency tables with `cargo-sort`

jpds:
Large change, but it's just the output of running `cargo-sort@2.1.4` across the workspace with its defaults.

Also adds a CI job to enforce this sorting going forward.


jpds:
@HaoboGu Done - it now uses the `--grouped` flag to keep the sections separate.

HaoboGu:
what's the grouping rule? and the comments are verbose

jpds:
Simply the ordered blocks that the dependencies are in. `cargo-sort` itself doesn't have a way of enforcing a rule - which is the tradeoff with just having the sorted just sort everything.

I've trimmed down the comments.

HaoboGu:
so cargo-sort is not used anymore?

jpds:
It's still used - it's just that a 'group' from its perspective within the dependencies is a space-separated block.

I think it'll be better to revert back to having everything sorted - but with a comment of what the important dependencies are.

HaoboGu:
All local dependencies also need to be put on the top.
I feel that adding extra script & ci is not needed because sorting the dependencies is not a critical task. Just do it in one-shot

PR #1147, opened 2026-09-15
fix(storage): keep stored data when one item fails to decode

### LINKED ISSUE #1105: An undecodable storage item erases the entire storage region and reboots the keyboard
### Summary

`KeyMap::new_from_storage` treats an error from **any** startup storage reader
(`read_keymap`, `read_behavior_config`, `read_macro_cache`, `read_combos`,
`read_forks`, `read_morses`) as a reason to erase the complete storage map and
call `reboot_keyboard()`.

`read_keymap` scans every item in the sequential-storage map and decodes every
value as `StorageData`, so a single item-level serialization failure aborts the
scan. The caller then erases the entire configured RMK storage region —
including stored keymap edits, settings and BLE bonds — and reboots. The reboot
also clears volatile diagnostic state before the failure can be inspected
on-device, which makes the wipe easy to misattribute to the power cycle or the
flash itself.

### Environment / versions

- RMK 0.9.0 — tag `rmk-v0.9.0`, commit `986017d901c7609a9a79167686524f014effa68a`
- Also present on `main` at `bbda6c439790a6e27aca0335e2fb1b8252d367a6`
  (checked 2026-08-30)
- `sequential-storage` 8.0.1
- Features: `storage`, `vial`
- RP2040, USB + Vial build, RMK storage in the board's external QSPI/XIP flash

### Steps to reproduce (upstream features only)

1. Build and flash a `keyboard.toml` firmware with `storage` and `vial`
   enabled, with `KEYBOARD_TOML_PATH` set in `.cargo/config.toml` as the
   `use_config` examples do.
2. Save at least one macro through Vial. Initial storage setup does not write a
   `MacroData` item; a host macro update is required.
3. Change only `[rmk] macro_space_size` in `keyboard.toml` (for example from
   256 to 128). Keep everything else identical: same RMK version, features,
   profile and target directory, and do not run `cargo clean`.
4. Rebuild, flash and boot.

What happens:

- `rmk-types/build.rs` sees the `keyboard.toml` change and regenerates
  `MACRO_SPACE_SIZE`, but `rmk/build.rs` does not re-run — it only declares
  `cargo:rerun-if-changed=build.rs` — so `BUILD_HASH` is reused and
  `check_enable` accepts the existing storage map. (I verified the Cargo
  behaviour with a minimal two-crate workspace rather than assuming it:
  changing a file tracked by a dependency's build script re-runs that script
  and changes its generated constant, while the dependent crate's build script
  is not re-run.)
- The previously stored `MacroData` value holds an array of the old length,
  which `macro_bytes_serde` rejects with `invalid_length`.
- On unmodified 0.9.0 the failure first occurs while `read_keymap` scans every
  map value as `StorageData`. `read_keymap` returns `Err`, and
  `KeyMap::new_from_storage` erases the complete map and reboots.
- If only the generic scan were changed to skip the item, the later keyed
  `read_macro_cache` call would still fail on the same value — which is why
  scan skipping alone is not sufficient (see suggested fix).

The stored keymap edits, settings and BLE bonds are gone afterwards. This
contradicts `docs/docs/main/docs/features/storage.md`, which says that
rebuilding after a `keymap`/`keyboard.toml` edit reuses the existing hash and
preserves those stored values.

### Host-side reproduction (no hardware needed)

The decode failure reproduces with the in-tree RAM-backed `TestFlash`:

1. Initialize a `Storage` over a `TestFlash` normally.
2. Store a raw value under `StorageKey::MacroData` that carries the current
   `MacroData` enum tag but a postcard byte-string payload shorter than the
   current `MACRO_SPACE_SIZE`.
3. Call `read_keymap` — on 0.9.0 it returns `Err`.
4. Call `read_macro_cache` — it also returns `Err`.

(Note: RMK's native async tests normally require `cargo-nextest`, because the
embassy-time mock driver is process-global.)

This host test demonstrates the decode/read failure directly; the erase and
reboot follow from the `new_from_storage` error branch. I have also seen the
end-to-end erase on RP2040 hardware, where I first hit this path through a
different undecodable item (see context).

### Expected behavior

An item-level serialization failure should not trigger an irreversible full
erase. Flash I/O and map-corruption errors may still abort restoration, but the
stored bytes should be preserved unless the firmware deliberately resets
storage (Vial EEPROM reset, `ClearEeprom`, `clear_storage`) or has positively
identified an incompatible layout (the `BUILD_HASH` check).

### Actual behavior / impact

The complete configured RMK storage region is erased and the keyboard reboots.
Stored keymap edits, RMK settings and BLE bonds are removed, and volatile
diagnostics are cleared before they can be inspected.

The erase is also best-effort while the reboot is unconditional: `erase_all()`
failures are discarded with `.ok()` and `reboot_keyboard()` runs anyway. If the
erase does not actually succeed, the next boot still finds the stored
`StorageConfig` with a matching `BUILD_HASH`, so `check_enable()` returns true,
`Storage::new` does not re-initialize, the same read fails again, the same erase
fails again, and the keyboard reboots again. On flash that cannot be erased this
is an unrecoverable boot loop — so the current policy does not reliably deliver
the known-good state that it destroys the user's data to reach.

### Suggested fix

Two related changes (I am running both on a fork with regression tests):

1. In the generic `read_keymap` scan, skip individual items whose key **or**
   value fails to deserialize (`SerializationError`), log a warning, and
   continue the scan. Keep propagating flash I/O, corruption and other
   map-level errors. This also makes the scan tolerant of items written by a
   newer firmware or by a downstream extension.
2. In `KeyMap::new_from_storage`, do not erase or reboot merely because a
   startup reader returned an error. Preserve the flash contents and continue
   booting with the current in-memory state: values that were not restored keep
   their initialized defaults, while values restored before a later failure may
   remain applied (the restore sequence is not transactional). This second
   change is still needed for keyed readers such as `read_macro_cache`, which
   do not use the generic scan and may still return `Err` for an undecodable
   item.

Known limitation of (1): the iterator's serialization error does not identify
the key, so if an older value for the same key was already applied before the
newest value failed to decode, the older value remains in memory rather than
the compiled default. This is fixable if you would rather not carry that
limitation: iterating with a value type whose `deserialize_from` always succeeds
and decoding `StorageData` in RMK's own code keeps the key available for the
failing item, so the scan can log which key is undecodable and reset just that
key to its compiled default. I left it out to keep this change minimal, and I am
happy to include it if you prefer the stronger behaviour.

Deliberate reset stays available: Vial EEPROM reset / `ClearEeprom` and
`clear_storage` for a full reset, and `clear_layout` for a startup layout-only
reset.

I would be happy to prepare a small `main`-based PR with the failing test if
this failure policy is acceptable.

### Context

Following the maintainer guidance in #932 to keep genuinely board-specific
persistence in our own fork, my firmware stores one board-specific item in the
same sequential-storage map, and an undecodable version of that item is how I
first ran into this path. That downstream item is not required to reproduce the
bug: a saved Vial macro followed by a `macro_space_size` change reaches the
same erase using only upstream features.

techmech-keeb:
Fixes #1105.

Two changes on the boot-time restore path:

1. `read_keymap` skips a value that does not decode as `StorageData` and keeps scanning, so items written after it are still applied. Flash I/O errors still abort `read_keymap`.
2. `KeyMap::new_from_storage` no longer erases the storage region and reboots when a reader fails. It logs the failure, keeps the region intact, and continues with what the readers restored; anything that could not be read stays at its compiled-in default.

Why: one undecodable item took every stored keymap, setting and BLE bond with it, and the ~100 ms reboot left nothing to inspect. It is reachable with upstream features alone: save a macro from Vial, change `[rmk] macro_space_size`, rebuild incrementally and flash. `MACRO_SPACE_SIZE` changes but `BUILD_HASH` (git HEAD + build time; `rerun-if-changed=build.rs` only) does not, so the stored `MacroData` fails `invalid_length` -- first in the scan, then in the keyed read. (Traced through the sources and a minimal cargo experiment; the end-to-end erase I saw on hardware came through a different undecodable item.)

Three consequences, stated rather than hidden:

- A skipped item stays in flash until its key is written again or the page is reclaimed, so each boot logs the warning until then.
- The map iterator yields every version of a key in write order, so when the newest value of a key does not decode, the previous decodable one is restored. `startup_scan_keeps_the_previous_value_when_the_newest_does_not_decode` pins that.
- On a flash I/O error mid-scan the runtime keeps what was restored up to that point. Erasing a flash that fails to read was the least reliable answer anyway: the erase can fail too, and the unconditional reboot then loops.

Not changed by this patch: Vial `EepromReset` (`host/via/mod.rs`), and `clear_storage` and the `build_hash` mismatch re-initialisation (both in `Storage::new`, which runs before `new_from_storage`). Making `BUILD_HASH` track configuration would be a complementary change; on its own it would still leave a full erase as the response to a single bad item.

Tests: three regression tests in `storage::tests` fail on `main` and pass here (`cargo nextest run --config-file .config/nextest.toml --manifest-path rmk/Cargo.toml --no-default-features --features vial,host_lock,_no_usb,steno,passkey_entry -E 'test(startup_)'`). `scripts/format_all.sh` and the clippy matrix are clean.


techmech-keeb:
Thanks for clarifying. Closing this in favor of the erase-on-mismatch direction in #1148.

PR #1148, opened 2026-09-15
Support read & request in storage

HaoboGu:


techmech-keeb:
One set of fingerprint inputs looks missing: the `keyboard.toml` constants that shape stored values — `MACRO_SPACE_SIZE`, and `COMBO_SIZE`/`MORSE_SIZE` for the heapless containers.

I checked `MACRO_SPACE_SIZE` on this branch by building the storage tests twice with different values and feeding one build's flash image to the other. The fingerprint is identical for 256 and 128 (`0xc4f9db10`), so the change is not seen as a new firmware. Growing the value (128→256) then fails in the startup scan with `SerializationError` on the old `MacroData` and reaches the read-failure erase in `keymap.rs`; shrinking it (256→128) makes `fetch(StorageConfig)` in `Storage::new` fail with `BufferTooSmall(260)` while walking past the old item, so `kept` is false and the region is reinitialized. Either way every stored item is lost, just by different paths.

Mixing `MACRO_SPACE_SIZE` into the fingerprint would route both of its cases through the firmware-mismatch erase, which is what #1105 is about — I'll tighten the issue text, since its "256 to 128" example takes the second path. `COMBO_SIZE`/`MORSE_SIZE` only break when shrunk, so leaving them out may well be deliberate; I'm just noting they share the property. (Also seen on an RP2040 board: flashing a build that differs only in `macro_space_size` wipes the Vial macros, key edits and layout options.)

